PR #34, opened 2026-02-08
Progress bar: show buffer ahead, update while paused. Remove retainBackBufferFromKeyframe option.

cowera:
**remove retainBackBufferFromKeyframe option**
  - backBufferDurationMs == 0 → back buffer is disabled, so retainBackBufferFromKeyframe is irrelevant
  - backBufferDurationMs > 0 → back buffer is enabled, and you'd always want retainBackBufferFromKeyframe = true for memory efficiency

  There's really no practical reason a user would enable back buffer but set retainBackBufferFromKeyframe = false. Without it, the back buffer retains from
  arbitrary positions which wastes memory with no meaningful benefit (the "slightly less precise seeking" tradeoff is negligible in practice since keyframes
  are typically every 2-5 seconds).

  So deriving it as retainBackBufferFromKeyframe = backBufferDurationMs > 0 removes a confusing toggle from the UI with no real loss of control.

**maxBufferSizeMb formula aligned**
  1. Formula aligned: maxBufferSizeMb now uses 60% of free heap (maxMemory - usedMemory) instead of 40% of max heap — matching the PlayerViewModel's
  auto-calculation exactly.
  2. Cap raised: from 500MB to 1000MB to match PlayerViewModel's coerceIn upper bound.
  3. Removed unused Context constructor parameter, @ApplicationContext annotation, and ActivityManager import since the new formula only uses Runtime.

**show ahead buffer in player (white)**

**update progress bar position while paused, too. same for buffer position.**
- this includes a FIX: until now, the progress bar was not updated when playback was paused, even though it is possible to skip ahead or back while paused, using a remote for example.

**Changed some buffer settings texts for clarification**

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackSettingsScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackSettingsScreen.kt
@@ -717,6 +717,9 @@ fun PlaybackSettingsContent(
                         onValueChange = { newValue ->
                             coroutineScope.launch {
                                 viewModel.setBufferMinBufferMs(newValue * 1000)
+                                if (newValue * 1000 > playerSettings.bufferSettings.maxBufferMs) {
+                                    viewModel.setBufferMaxBufferMs(newValue * 1000)
+                                }
                             }
                         }
                     )


cowera:
added a fix for buffer settings: minBufferMs cannot be higher than maxBufferMs, but it was possible when increasing minBufferMs beyond maxBufferMs value.

tapframe:
https://github.com/tapframe/NuvioTV/pull/60#issuecomment-3878800435

PR #60, opened 2026-02-10
Immersive view

cowera:
See in action: https://jumpshare.com/s/mS1N9T2H0rWFxfWebKVJ

Fullscreen infinite poster wall.

Keeping focus on a tile for 2.5s shows a info box.
Keeping focus on a tile for 5s recenters tiles in all other rows.

Bildschirmfoto 2026-02-10 um 11 39 18

Added to Layout Selection Screen and Layout Settings Screen as well. 

## how it works

```
// ── Infinite-scroll grid with per-row recenter ──────────────────────────────
//
// The grid is an infinite carousel in both axes.  All rows share a single
// animated column value (animatedCol) so that left/right input scrolls every
// row by one tile in lockstep.  Each row may have a different number of unique
// items; items repeat via mod-wrapping to fill the infinite strip.
//
// Recenter system
// ───────────────
// After RECENTER_DELAY_MS of idle, every *non-focused* row smoothly slides to
// its nearest "col-0" (the position where item 0 is centred on screen).
// The focused row keeps its position — the user's scroll offset is preserved.
//
// Because rows have different item counts, their col-0 positions differ.
// A per-row offset map (rowOffsets) stores the delta between animatedCol and
// each non-focused row's actual column.  When the user scrolls, animatedCol
// changes and every offset stays constant, so all rows move together.
//
// Vertical navigation after recenter
// ───────────────────────────────────
// When the user presses Up/Down after a recenter has happened:
//   1. The old focused row's offset is recorded as 0 (its true position).
//   2. animatedCol is shifted to match the target row's visual column.
//   3. Every stored offset is compensated by the same shift so that no row
//      moves on screen.
// This guarantees zero visual jumps for any row regardless of item counts.
//
// The next recenter (after another idle timeout) will smoothly bring all
// non-focused rows back to their col-0.
```

### testing

- [ ] scroll around a lot and see if any sudden "jumps" of rows happen after re-centering
- [ ] select different views and ensure everything still works as expected
- [ ] continue watching movies/shows should appear in the first row
- [ ] wait for 60s to enter idle animation

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/immersive/ImmersiveScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/immersive/ImmersiveScreen.kt
@@ -31,7 +31,7 @@ fun ImmersiveScreen(
             .background(Color.Black)
     ) {
         when {
-            uiState.isLoading -> {
+            uiState.isLoading && uiState.catalogRows.isEmpty() -> {
                 Box(
                     modifier = Modifier.fillMaxSize(),
                     contentAlignment = Alignment.Center
--- a/app/src/main/java/com/nuvio/tv/ui/screens/immersive/ImmersiveViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/immersive/ImmersiveViewModel.kt
@@ -207,8 +207,8 @@ class ImmersiveViewModel @Inject constructor(
                 }
         }
 
-        // Wait for all catalogs to finish before showing the grid
-        jobs.forEach { it.join() }
+        // Wait for at least the first catalog to arrive, then mark not loading
+        jobs.firstOrNull()?.join()
         _uiState.update { it.copy(isLoading = false) }
     }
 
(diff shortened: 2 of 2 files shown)


AdityasahuX07:
This looks cool but it should be at side-bar tab for "Immersive View" not the default home-page.
made changes like that put it at sidebar tabs.

cowera:
> This looks cool but it should be at side-bar tab for "Immersive View" not the default home-page. made changes like that put it at sidebar tabs.

Hm, interesting thought. I kind of like it as my go-to homepage now that it includes "Continue watching" section as well. But I also understand that it might be nice as a tab for exploring in a different kind of way. Maybe, unless selected as homepage, it could sit in a tab?

tapframe:
Thanks for your contribution, it looks really cool. but, something like this isn’t currently planned for Nuvio. Also, we’re not accepting any UI or functionality related PRs at the moment unless they address a critical bug. we will announce when PRs are open to public, thanks

PR #138, opened 2026-02-16
Modern view

Rasimsson:
Added some additional fixes:

1. Better scrolling/movement/focus management
2. Hero now always utilizes backdrop instead of fallback and posters enlarged
3. Further cleanup/modernization of the posters


saimuelbr:
can u show a preview?

PR #145, opened 2026-02-16
Add auto frame rate option Improvements

kernexshadow:
Reworked Auto Frame Rate matching using a Kodi‑style flow (pause → switch display mode → resume), with DisplayManager listener + timeout for robustness.
Added AFR modes: Off, Start, and Start/Stop, with settings wiring.
Added on‑screen info showing the active refresh rate/resolution when playback starts or after a mode switch.

Improved AFR logic.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -29,7 +29,7 @@ android {
     compileSdk = 36
 
     defaultConfig {
-        applicationId = "com.nuvio.tv2"
+        applicationId = "com.nuvio.tv"
         minSdk = 26
         targetSdk = 36
         versionCode = 14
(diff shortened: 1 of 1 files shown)


kernexshadow:
image

image

tapframe:
@CrissZollo do you mind testing this on your tv?

CrissZollo:
It works as mentioned. Do we need to have a start only option? I think most will want to have the start/stop since it will return the refresh rate to the original after exiting the player.

@tapframe what do you think?

kernexshadow:
> It works as mentioned. Do we need to have a start only option? I think most will want to have the start/stop since it will return the refresh rate to the original after exiting the player.
> 
> @tapframe what do you think?

It might be helpful if one is playing Next Episode automatically. So the screen don't blink. Between epsiodes.

Altough it might be an option in the future to don't stop between episodes.

tapframe:
> It works as mentioned. Do we need to have a start only option? I think most will want to have the start/stop since it will return the refresh rate to the original after exiting the player.
> 
> @tapframe what do you think?

If it works as intended, let's merge it. we'll do the modification as needed in teh future.

PR #170, opened 2026-02-19
Replace expanding card with focused overlay for episode details

skoruppa:
I noticed that during scrolling through episodes there is a rebound effect because previously focused item shrinks, and newly focus one takes the new free space, adding this unnecessary small animation to the left. 

https://github.com/user-attachments/assets/696a10c9-6fbf-4f75-9124-20e014642ff6

With my change, episode cards now have fixed width (no rebound effect when navigating). On focus, episode title, overview and date fade in as an animated overlay on the thumbnail instead of expanding the card sideways

https://github.com/user-attachments/assets/3484f180-aa59-4547-a166-6b7393ffd11d

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/EpisodesSection.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/EpisodesSection.kt
@@ -27,7 +27,7 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.zIndex
 import androidx.compose.ui.focus.FocusRequester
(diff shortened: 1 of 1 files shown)


skoruppa:
A similar effect is visible on the catalog view on the home, but because expanding happens after some time and not immediately it is not bothering as much, because you can jump between items without expending each of them

kernexshadow:
I kinda like the first view. There is more space, the text isn't cluttered. But the animation should be smoother.

skoruppa:
> I kinda like the first view. There is more space, the text isn't cluttered. But the animation should be smoother.

To be honest I also like it more >< the problem is I didn't manage to get rid of the debounce effect even though I really tried. And it really bothers me... an alternative would be to have all episodes expanded

kernexshadow:
> > I kinda like the first view. There is more space, the text isn't cluttered. But the animation should be smoother.
> 
> To be honest I also like it more >< the problem is I didn't manage to get rid of the debounce effect even though I really tried. And it really bothers me... an alternative would be to have all episodes expanded

If the scrolling is smoother and slower, maybe it could work.

PR #174, opened 2026-02-19
Buffer and network settings

cowera:
## Summary

  Add buffer tuning settings and parallel fetch (multi-connection) capabilities for progressive streams. Both features are bundled in one PR because they both affect RAM usage and share a unified memory budget that must be enforced to prevent
  out-of-memory errors.

  **Buffer settings:**
  - `setTargetBufferBytes` is now always called on `LoadControl`. By default all buffer values match the `androidx.media3.exoplayer` defaults (e.g. 125 MB target buffer for non-local video playback via `getDefaultBufferSize`)
  - Exposed min/max buffer duration, initial buffer, rebuffer buffer, back-buffer duration, and target buffer size as user-configurable sliders

  **Parallel connections (`ParallelRangeDataSource`):**
  - Downloads progressive files (MKV, MP4) using multiple parallel HTTP range requests, multiplying effective throughput when per-connection CDN limits apply
  - Improved performance: buffer pool to avoid GC churn, dynamic chunk timeout (`max(60, chunkSizeMB * 8)` seconds), automatic single-connection fallback when the server doesn't support range requests
  - `ParallelRangeDataSource.Factory` now uses the generic `DataSource.Factory` instead of `OkHttpDataSource.Factory`
  - Added `parallelConnectionCount` and `parallelChunkSizeMb` user-configurable options

  **Settings UI:**
  - New "Buffer" and "Network" sections under Playback Settings
  - RAM usage warning hint shown when tweaking settings that affect memory (targetBufferSizeMb, useParallelConnections, parallelConnectionCount, parallelChunkSizeMb)
  - Memory budget enforcement via `MemoryBudget` — caps total allocation to `Runtime.getRuntime().maxMemory() * 0.70` (70% budget ratio, may be tuned based on user feedback)
  - Reset-to-default buttons for both buffer and network settings
  - 1-line clamping kept, but long setting descriptions now scroll horizontally

  **Refactoring:**
  - Extracted hardcoded buffer/network defaults into `PlayerSettings` constants
  - Extracted `MemoryBudget` into its own file for shared use between ViewModel and UI

  buffer

## PR type

  - Approved larger change (link approval below)

## Why

  Parallel connections solve a common issue where users have sufficient bandwidth but still experience buffering due to per-connection CDN throughput limits (~100 Mbps per TCP connection). Buffer tuning gives advanced users control over
[...]

PR #175, opened 2026-02-19
feat(settings): add buffer settings UI to playback settings

halibiram:
## Summary
Adds a configurable buffer settings section to the Playback Settings screen, allowing users to customize video buffering behavior for their network conditions and device capabilities.

## Changes
- New `PlaybackBufferSettings.kt` with buffer configuration UI components
- Added Buffer section to `PlaybackSettingsSections.kt` 
- Wired buffer settings callbacks in `PlaybackSettingsScreen.kt`

## Buffer Settings Added
| Setting | Description | Range |
|---------|-------------|-------|
| Minimum Buffer | Minimum buffered duration before playback starts | 10-120s |
| Maximum Buffer | Maximum buffered duration during playback | 10-120s |
| Buffer for Playback | Buffer needed to start or resume playback | 1-30s |
| Buffer After Rebuffer | Buffer needed after a rebuffer event | 1-60s |
| Target Buffer Size | Maximum buffer size in memory (0 = auto) | 0-500MB |
| Back Buffer Duration | Video kept behind current position for seeking back | 0-60s |
| Retain Back Buffer from Keyframe | Keep back buffer aligned to keyframes | Toggle |

## Why
- Users with slow connections can increase buffer to reduce stuttering
- Users with fast connections or limited memory can decrease buffer
- Advanced users can fine-tune playback behavior for their specific setup
- Previously, buffer settings existed in DataStore but had no UI

## Validation
- [x] Code compiles successfully
- [x] UI follows existing settings design patterns (sliders, toggles)
- [x] Settings are properly connected to ViewModel and DataStore
- [x] Focus handling works correctly for TV navigation

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackBufferSettings.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackBufferSettings.kt
@@ -1,0 +1,180 @@
+@file:OptIn(ExperimentalTvMaterial3Api::class)
+
+package com.nuvio.tv.ui.screens.settings
+
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyListScope
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Memory
+import androidx.compose.material.icons.filled.Storage
+import androidx.compose.material.icons.filled.Timer
+import androidx.compose.material.icons.filled.Speed
+import androidx.compose.material.icons.filled.History
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import androidx.tv.material3.ExperimentalTvMaterial3Api
+import androidx.tv.material3.MaterialTheme
+import androidx.tv.material3.Text
+import com.nuvio.tv.data.local.BufferSettings
+import com.nuvio.tv.ui.theme.NuvioColors
+
+internal fun LazyListScope.bufferSettingsItems(
+    bufferSettings: BufferSettings,
+    onSetBufferMinBufferMs: (Int) -> Unit,
+    onSetBufferMaxBufferMs: (Int) -> Unit,
+    onSetBufferForPlaybackMs: (Int) -> Unit,
+    onSetBufferForPlaybackAfterRebufferMs: (Int) -> Unit,
+    onSetBufferTargetSizeMb: (Int) -> Unit,
+    onSetBufferBackBufferDurationMs: (Int) -> Unit,
+    onSetBufferRetainBackBufferFromKeyframe: (Boolean) -> Unit,
+    onItemFocused: () -> Unit = {},
+    enabled: Boolean = true
+) {
+    item {
+        Text(
+            text = "Buffer",
+            style = MaterialTheme.typography.titleMedium,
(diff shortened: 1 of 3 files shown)


cowera:
Hey, the buffer changes clash with my PR (#174). However, looks like you pretty much took my old implementation. My PR contains some improvements on that.

My suggestion would be to merge all other changes, which do look quite useful (finally someone decided to gitignore idea folder 😂). And for the future, I'd advice split PRs where possible. Cheers

johnneerdael:
@cowera @halibiram There is also some overlap with my PR, both these PR's heavily rely on network-direct streaming while my PR implements a VOD cache system which based on my testing makes a huge difference and prevents Socket Timeouts during high file sizes that are streamed and effectively prevent buffering, which is one of my main use-cases and most likely for many others that have fast connections at home and use a debrid service.

When you stream a very large movie without a disk cache (This branch), ExoPlayer stores the downloaded video chunks in your device's RAM. To prevent the app from crashing due to out-of-memory (OOM) errors, ExoPlayer usually caps this RAM buffer at around 32MB to 50MB. For a 25+GB movie, a 50MB buffer only gives you a few seconds of "runway." If your network connection blips for even 5 seconds, the buffer drains, the player starves, and you get a spinning wheel or a SocketTimeoutException.

When you use the SimpleCache (My branch), the app writes the video data directly to the device's persistent storage (Flash/SSD) instead of RAM. My branch allocates up to 2GB of disk space for the cache (VOD_CACHE_MAX_BYTES = 2L * 1024L * 1024L * 1024L). That 2GB gives the player several minutes of runway. Your network can completely drop out or throttle, and the player will happily keep reading the movie from the local disk while the OkHttpClient quietly recovers the connection in the background.

Note: in reality I'm are not filling up 2GB, due to load control we only put 50s in the cache, which on my 28GB movie equaled less then 300MB at all times.

cowera:
> @cowera @halibiram There is also some overlap with my PR, both these PR's heavily rely on network-direct streaming while my PR implements a VOD cache system which based on my testing makes a huge difference and prevents Socket Timeouts during high file sizes that are streamed and effectively prevent buffering, which is one of my main use-cases and most likely for many others that have fast connections at home and use a debrid service.
> 
> 
> 
> When you stream a very large movie without a disk cache (This branch), ExoPlayer stores the downloaded video chunks in your device's RAM. To prevent the app from crashing due to out-of-memory (OOM) errors, ExoPlayer usually caps this RAM buffer at around 32MB to 50MB. For a 25+GB movie, a 50MB buffer only gives you a few seconds of "runway." If your network connection blips for even 5 seconds, the buffer drains, the player starves, and you get a spinning wheel or a SocketTimeoutException.
> 
> 
> 
> When you use the SimpleCache (My branch), the app writes the video data directly to the device's persistent storage (Flash/SSD) instead of RAM. My branch allocates up to 2GB of disk space for the cache (VOD_CACHE_MAX_BYTES = 2L * 1024L * 1024L * 1024L). That 2GB gives the player several minutes of runway. Your network can completely drop out or throttle, and the player will happily keep reading the movie from the local disk while the OkHttpClient quietly recovers the connection in the background.
> 
> 
> 
> Note: in reality I'm are not filling up 2GB, due to load control we only put 50s in the cache, which on my 28GB movie equaled less then 300MB at all times.

The default cache (target buffer size) is not 25 or 50 MB, but 125 MB. On my Shield Pro I can go up to 400 MB without problems, but in my case that doesn't prevent buffering. What actually does prevent buffering for me is the parallel connection feature which is part of my PR too.

Anyway, I thought about disk cache too. Hope we can merge both.

PR #193, opened 2026-02-20
Performance fixes - replace poster label Text with drawWithCache to reduce scroll jank

skoruppa:
I tried to find the reason for janky animations on my smart TV (TCL 85MQLED85) while scrolling through catalogs (In Classic Mode). I applied some basic optimizations - skip recomposition on focus change when backdrop/trailer features are disabled, fix unnecessary full-screen recomposition on focus restore - but the main culprit turned out to be text rendering under poster cards.

Each Text composable inside a LazyRow item triggers a full text layout pass during composition every time a new item enters the screen while scrolling. Replacing it with drawWithCache + TextMeasurer moves text layout out of the composition phase - it's measured once and cached, only redrawn on subsequent frames.

Applied to classic view (ContentCard) and grid view (GridContentCard). Modern view was left unchanged - there's a lot going on there and I didn't notice measurable gains.

Also fixed a crash: FocusRequester is not initialized when scrolling down quickly - caused by downFocusRequester pointing to a LazyRow that wasn't composed yet.

Results (measured with adb shell dumpsys gfxinfo, warm cache):
- Horizontal scroll (LazyRow): ~55% → 30% janky frames
- Vertical scroll (LazyColumn): ~30% → 15% janky frames

It is still not perfect, but with my slow smart tv CPU there is probably not much more I can do

**Before**
https://github.com/user-attachments/assets/d0f182c8-160d-4039-ba95-e4b17917e27e

**After**
https://github.com/user-attachments/assets/282a1f76-6e36-4358-809c-0f68d77be2f3

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/EpisodesSection.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/EpisodesSection.kt
@@ -6,11 +6,10 @@ import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
+import androidx.compose.foundation.Canvas
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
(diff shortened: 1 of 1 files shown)


tapframe:
Great find. I can see in the video that it has improved noticeably. Can you continue the PR by optimizing other parts of the app? The lack of a physical device with me is causing issues with gathering real time performance benchmark.

skoruppa:
> Great find. I can see in the video that it has improved noticeably. Can you continue the PR by optimizing other parts of the app? The lack of a physical device with me is causing issues with gathering real time performance benchmark.

To be honest, there is not much more I can or know how to do. I now applied my fix to the EpisodeSection. The gains there are not noticeable as much, as only episode badge text is present all the time during scroll, but "_it is something_"
image

PR #195, opened 2026-02-20
Close when pressing back two times

skoruppa:
Quick change to close the app instead of minimizing it when pressing back at the navigation root.

Now, the first back press shows a toast: "_Press back one more time to close the app_", and a second press within 2 seconds calls `finish()` to fully close it.

I was thinking about hiding this behind a setting, but if user want to minimize the app pressing the "**home**" button does it, so functionality is not lost


ram130:
Can you double check if this is still needed.

PR #235, opened 2026-02-21
Fix trailer autoplay in background without resetting focus state

### LINKED ISSUE #234: When trailer autoplay is enabled, trailers can start playing after the app is sent to the Android home screen.
This happens in:
- Classic Home with expanded poster preview
- Details screen where trailer autoplay starts after the Play-button idle delay

### Steps to reproduce
1. Enable trailer autoplay in settings.
2. Go to Classic Home and focus an item (or open a title details page and leave focus on Play).
3. Press Home to send the app to background before the autoplay delay ends.
4. Wait for the configured delay.

### Actual result
Trailer playback (often audio) starts while the app is in background.

### Expected result
Trailer autoplay must not start while the app is not in foreground (`RESUMED`).
Any pending autoplay timer should be cancelled on `ON_PAUSE/ON_STOP` and only re-armed when the app returns to foreground.

kernexshadow:
## Summary
This PR fixes trailer autoplay starting while the app is in the background (after pressing Home), and preserves the current focused/expanded context when returning to the app.

## What changed
- Added lifecycle `RESUMED` gating to trailer/autoplay flows in:
  - `ContentCard` (classic focused/expanded poster flow)
  - `ModernHomeContent` (expanded card and hero trailer autoplay flow)
- Added app foreground/background event handling in detail screen flow:
  - New event: `OnAppForegroundChanged`
  - `MetaDetailsScreen` now reports lifecycle foreground changes
  - `MetaDetailsViewModel` cancels idle autoplay timer when app goes background
- Kept focus/expanded state intact on app resume (no forced reset to first poster/profile screen behavior from this fix).

## Result
- Trailer autoplay no longer starts while the app is outside foreground.
- Returning to the app keeps the previous navigation/focus context instead of resetting it.

Closes #234

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/components/ContentCard.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/components/ContentCard.kt
@@ -15,6 +15,7 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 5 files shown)


kernexshadow:
https://github.com/NuvioMedia/NuvioTV/issues/592 Already does this.

kernexshadow:
https://github.com/NuvioMedia/NuvioTV/issues/592 Already does this.

PR #251, opened 2026-02-22
Add "Sync by Line" subtitle flow from the Delay overlay

### LINKED ISSUE #242: Subtitle text-based delay options
Thank you for fixing the subtitle add-on issue. It would be great if you could add a text-based delay option for the subtitles, similar to what Cloudstream does.

kernexshadow:
## Summary

This adds a subtitle sync flow that starts from the existing Delay screen.

It lets the user open Sync by Line, capture the exact playback moment, choose the matching subtitle line, and immediately return to Delay to fine tune if needed.

In practice, this makes subtitle sync faster when subtitles are badly offset, while keeping the normal delay controls for small adjustments.

The sync dialog is focused on the current playback context:

it works around the current time window (±3 minutes),
Shows lines after pressing Sync Now,
The user press the line and it applies the result directly back into the player.

## Why

For subtitles that are way off, its really difficult to sync with standard Audio Delay. 

This is really helpfull, for subtitles when the timings are really off.

## Screenshots

[Screenshots:](https://github.com/user-attachments/assets/05cd7f87-7423-4d87-97ac-622c7b395d03)

## Testing

Tested in video above.

## Breaking changes

None.

## Linked issues

Closes: #242


klemenko1:
Excellent feature

If possible, it would also be great to consider adding an option to rescale subtitles to a specific FPS. The main challenge is that SRT files do not contain any frame rate information, so any FPS based correction would need to rely on user input or inferred values rather than metadata from the file itself.

In the Android TV (Kotlin) app, this could be implemented in several practical ways:

- Add a simple dialog where the user selects the assumed source FPS (e.g., 23.976, 24, 25, 29.97) and the target FPS, then rescale all timestamps using a ratio (sourceFps / targetFps).
- Automatically read the video frame rate from ExoPlayer (videoFormat?.frameRate) and offer a “Match subtitles to video FPS” action.
- Integrate it into the existing Delay/Sync flow as an advanced option alongside global delay and line-based sync.
- Apply a non-destructive global timeScale factor during rendering instead of permanently modifying parsed subtitle data, making it easy to toggle or undo.
- In a more advanced scenario, calculate the scale factor automatically from two manual sync points to compensate for drift over time.

This would complement the current sync improvements very well, especially in cases where timing issues are caused by FPS mismatches rather than a simple offset.

skoruppa:
@tapframe any thoughts? I always liked the idea, but I'm not sure if the "Sync By Line Interface" matches rest of the UI

kernexshadow:
> @tapframe any thoughts? I always liked the idea, but I'm not sure if the "Sync By Line Interface" matches rest of the UI

We can always try to change the UI. It's usefull for subtitles that are really off. Using normal delay to correct those subtitles it's very hard.

kernexshadow:
> @tapframe any thoughts? I always liked the idea, but I'm not sure if the "Sync By Line Interface" matches rest of the UI

UI changed:


https://github.com/user-attachments/assets/05cd7f87-7423-4d87-97ac-622c7b395d03

RSAK2105:
@tapframe Nuvio Guru, "Sync by line" is excellent feature to synch subtitles to overcome current capability which requires lot of time to match the timing correctly. Pls consider to merge this 🙏

tapframe:
> > @tapframe any thoughts? I always liked the idea, but I'm not sure if the "Sync By Line Interface" matches rest of the UI
> 
> 
> 
> UI changed:
> 
> 
> 
> 
> 
> https://github.com/user-attachments/assets/05cd7f87-7423-4d87-97ac-622c7b395d03
> 
> 
> 
> 

Reduce the focus scaling so it doesn’t clip on the sides. Also, make the focused subtitle use a filled accent color instead of an outline, and hide the controls while adjusting. Once these changes are done, we’re good to merge.

kernexshadow:
> > > @tapframe any thoughts? I always liked the idea, but I'm not sure if the "Sync By Line Interface" matches rest of the UI
> > 
> > 
> > UI changed:
> > 
> >   
> >     
> >     
> > 
> >     Syncbyline.mp4
> 
> Reduce the focus scaling so it doesn’t clip on the sides. Also, make the focused subtitle use a filled accent color instead of an outline, and hide the controls while adjusting. Once these changes are done, we’re good to merge.

Changed. Any more adjusted needed?

https://github.com/user-attachments/assets/5e67d205-330e-4a0a-a061-6ab48db1cf16

PR #272, opened 2026-02-23
feat[#265]: add addon subtitle startup modes to surpass reloads on subtitle switch

### LINKED ISSUE #265: [BUG] Selecting subtitles from addon triggers player reload on NuvioTV
**Description:**
When selecting a subtitle from an addon during playback, the player reloads the stream. While it resumes from the last played position, the reload is unnecessary and disrupts the viewing experience.

**Steps to Reproduce:**
1. Open NuvioTV and start playing any content.
2. Open the subtitle selection menu.
3. Select a subtitle (from an addon).
4. Observe that the player reloads the entire stream.

**Expected Behavior:**
The subtitle should be applied to the current playback session without reloading the player, similar to how it works on Nuvio Mobile.

**Actual Behavior:**
The player reloads and resumes from the last position.

**Comparison:**
This issue does not occur in Nuvio Mobile. On the mobile version, selecting subtitles from an addon applies them immediately without triggering a player reload. 

**Device Information:**
NuvioTV Version: v0.4.3

Thanks @tapframe

rubkn:
This fixes addon subtitle switching causing stream reloads during playback on TV.

> [!NOTE]
> **Startup speed tradeoff:**  
> To reduce subtitle-switch reloads on TV, this feature can prefetch and pre-attach addon subtitles before playback starts.  
> This may increase initial player startup time, especially in:
> - **Preferred languages** mode (moderate impact), and
> - **All addon subtitles** mode (highest impact).
>  
> **Fast startup** keeps the quickest start, but selecting a not-yet-attached addon subtitle can still trigger one media re-prepare/reload (Media3 limitation).

## Why this approach (Media3 / ExoPlayer constraint)

In Media3, sideloaded external subtitles are attached through `MediaItem.SubtitleConfiguration` as part of the media item/media source setup (before `prepare`).  
When a subtitle is not already attached, switching to it requires rebuilding the media source + `prepare()`, which is the visible reload.

So the fix is:
- avoid rebuild when the target subtitle track is already attached, and
- attach more subtitle tracks upfront (optional modes) so later switches are in-place.

References:
- https://github.com/google/ExoPlayer/issues/1649
- https://developer.android.com/media/media3/exoplayer/media-items

## What changed

- Added stable IDs for addon subtitle tracks.
- Reused already attached addon subtitle tracks for in-place switching (no reload path).
- On first attach, included fetched addon subtitles so subsequent switches are usually in-place.
- Added **Addon Subtitle Loading** modes:
  - **Fast startup**: quickest start, may reload when selecting a not-yet-attached addon subtitle.
  - **Preferred languages**: prefetch + attach only preferred/secondary subtitle languages.
  - **All addon subtitles**: prefetch + attach all addon subtitles before playback.
- Improved subtitle language normalization/order so preferred + secondary language sorting is consistent (I had previously implemented this but somehow it wasn't ordered now, someone must have changed this incorrectly).

Closes #265

### With preferred languages only
https://github.com/user-attachments/assets/7a7a1018-5f45-4a08-a8a6-55ed4915da75

### With all languages
https://github.com/user-attachments/assets/69dee758-4188-4a9c-ad9a-094caa70de16

### Playback settings (new options)
Screenshot 2026-02-27 at 12 13 35 PM

### Addon subtitle loading dialog
Screenshot 2026-02-27 at 12 13 52 PM

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerTrackSelection.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerTrackSelection.kt
@@ -107,6 +107,40 @@ internal fun PlayerRuntimeController.applyAddonSubtitleOverride(addonTrackId: St
     return false
 }
 
+internal fun PlayerRuntimeController.applyAddonSubtitleOverrideByLanguage(
+    language: String
+): Boolean {
+    val player = _exoPlayer ?: return false
+    player.currentTracks.groups.forEach { trackGroup ->
+        if (trackGroup.type != C.TRACK_TYPE_TEXT) return@forEach
+        for (i in 0 until trackGroup.length) {
+            val format = trackGroup.getTrackFormat(i)
+            if (format.id?.contains(PlayerRuntimeController.ADDON_SUBTITLE_TRACK_ID_PREFIX) != true) {
+                continue
+            }
+            if (!PlayerSubtitleUtils.matchesLanguageCode(format.language, language)) {
+                continue
+            }
+            val override = TrackSelectionOverride(trackGroup.mediaTrackGroup, i)
+            player.trackSelectionParameters = player.trackSelectionParameters
+                .buildUpon()
+                .setOverrideForType(override)
+                .setTrackTypeDisabled(C.TRACK_TYPE_TEXT, false)
+                .build()
+            Log.d(
+                PlayerRuntimeController.TAG,
+                "applyAddonSubtitleOverrideByLanguage: found id=${format.id} lang=${format.language} at group/track $i"
+            )
+            return true
+        }
+    }
+    Log.d(
+        PlayerRuntimeController.TAG,
+        "applyAddonSubtitleOverrideByLanguage: track not found yet for language=$language"
+    )
+    return false
+}
+
 internal fun PlayerRuntimeController.selectSubtitleTrack(trackIndex: Int) {
     _exoPlayer?.let { player ->
         Log.d(PlayerRuntimeController.TAG, "Selecting INTERNAL subtitle trackIndex=$trackIndex")
@@ -177,34 +211,19 @@ internal fun PlayerRuntimeController.selectAddonSubtitle(subtitle: Subtitle) {
 
         val normalizedLang = PlayerSubtitleUtils.normalizeLanguageCode(subtitle.lang)
         val addonTrackId = buildAddonSubtitleTrackId(subtitle)
-        val alreadyAttachedTrackIndex = _uiState.value.subtitleTracks.indexOfFirst {
-            it.trackId == addonTrackId
-        }
         val preAttachedByStartup = attachedAddonSubtitleKeys.contains(addonSubtitleKey(subtitle))
-        val fallbackAttachedTrackIndex = if (alreadyAttachedTrackIndex < 0 && preAttachedByStartup) {
-            _uiState.value.subtitleTracks.indexOfFirst { track ->
-                PlayerSubtitleUtils.matchesLanguageCode(track.language, normalizedLang)
-            }
-        } else {
-            -1
-        }
+        val appliedWithoutReload = applyAddonSubtitleOverride(addonTrackId) ||
+            (preAttachedByStartup && applyAddonSubtitleOverrideByLanguage(normalizedLang))
 
-        val trackIndexWithoutReload = when {
-            alreadyAttachedTrackIndex >= 0 -> alreadyAttachedTrackIndex
-            fallbackAttachedTrackIndex >= 0 -> fallbackAttachedTrackIndex
-            else -> -1
-        }
-
-        if (trackIndexWithoutReload >= 0) {
+        if (appliedWithoutReload) {
             Log.d(
                 PlayerRuntimeController.TAG,
-                "Switching ADDON subtitle without media reload id=${subtitle.id} index=$trackIndexWithoutReload"
+                "Switching ADDON subtitle without media reload id=${subtitle.id}"
             )
             pendingAddonSubtitleLanguage = null
             pendingAddonSubtitleTrackId = null
             pendingAudioSelectionAfterSubtitleRefresh = null
 
-            selectSubtitleTrack(trackIndexWithoutReload)
             _uiState.update {
                 it.copy(
                     selectedAddonSubtitle = subtitle,
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
@@ -383,7 +383,8 @@ internal fun resolvePreferredAudioLanguages(
             normalize(secondaryPreferredAudioLanguage)
         ).distinct()
     }
-    
+}
+
 internal suspend fun PlayerRuntimeController.prepareStartupSubtitles(
     mode: AddonSubtitleStartupMode,
     preferredLanguage: String,


rubkn:
I might have misunderstood the issue, I thought the problem was the video was restarting form beginning and not reloading while maintaining the last position. Will investigate this issue further.

rubkn:
@tapframe I investigated this further, and found out the issue: Media3 only switches between subtitle tracks that are already attached at prepare time. How do you feel about having a capped startup prefetch (1 or 2s) to fetch addon subtitles before first prepare, so later subtitle switching is instant and doesn’t reload the player?

We can scope it even simpler to only run with only addon subtitles matching preferred + secondary languages and if  the user later picks a subtitle outside those languages, we'll re-prepare. Let me know your thoughts on this.

skoruppa:
> @tapframe I investigated this further, and found out the issue: Media3 only switches between subtitle tracks that are already attached at prepare time. How do you feel about having a capped startup prefetch (1 or 2s) to fetch addon subtitles before first prepare, so later subtitle switching is instant and doesn’t reload the player?
> 
> We can scope it even simpler to only run with only addon subtitles matching preferred + secondary languages and if the user later picks a subtitle outside those languages, we'll re-prepare. Let me know your thoughts on this.

Out of curiosity, how does Stremio handle this? Not once I saw that subtitles from addons were returned after stream started and even for such cases, streams were not reloaded when selecting subtitles from those addons.

rubkn:
@skoruppa I believe Stremio fetches the addons subtitles after video is prepared, hence also the 1 or 2 second loading, I've encontered it everytime, now it makes sense I guess

skoruppa:
> @skoruppa I believe Stremio fetches the addons subtitles after video is prepared, hence also the 1 or 2 second loading, I've encontered it everytime, now it makes sense I guess

This confirms your findings https://github.com/google/ExoPlayer/issues/1649

So maybe an option? Let user decided what they prefer. Longer loading but changing subtitles does not restart your stream, or quicker loading but subtitles from addons restart stream

rubkn:
> > @skoruppa I believe Stremio fetches the addons subtitles after video is prepared, hence also the 1 or 2 second loading, I've encontered it everytime, now it makes sense I guess
> 
> This confirms your findings [google/ExoPlayer#1649](https://github.com/google/ExoPlayer/issues/1649)
> 
> So maybe an option? Let user decided what they prefer. Longer loading but changing subtitles does not restart your stream, or quicker loading but subtitles from addons restart stream

Sounds good! Will do!

rubkn:
This is now ready for review. Let me know if this is something you want to see implemented or not. @tapframe

tfouto:
I suggest prefered subtitles should only be fetched if no embedded subtitle is availbe for the language. When there is embedded subtitles it's always good enough.

tapframe:
great work. I was reviewing it, but it seems like the fetch all addons doesn’t work as expected. It does reloads the first few languages. In my case, English works without reload, but scrolling further down and choosing another language reloads the stream. tried app restart. Can you confirm this

rubkn:
For sure. I'm away today and tomorrow but will look into it as soon as I can

app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerTrackSelection.kt
```diff
@@ -151,21 +176,55 @@ internal fun PlayerRuntimeController.selectAddonSubtitle(subtitle: com.nuvio.tv.
...
+        if (trackIndexWithoutReload >= 0) {
+            Log.d(
+                PlayerRuntimeController.TAG,
+                "Switching ADDON subtitle without media reload id=${subtitle.id} index=$trackIndexWithoutReload"
+            )
+            pendingAddonSubtitleLanguage = null
+            pendingAddonSubtitleTrackId = null
+            pendingAudioSelectionAfterSubtitleRefresh = null
+
+            selectSubtitleTrack(trackIndexWithoutReload)
```
halibiram: Hey, nice work on this! I noticed one thing that might explain the issue @tapframe mentioned (English works but other languages still reload):

This calls `selectSubtitleTrack(trackIndexWithoutReload)` — but that function has a `continue` that skips any track with the `nuvio-addon-sub:` prefix. So the addon tracks you attach at startup can never actually be selected through this path, and the index gets misaligned too since skipped tracks aren't counted.

I think either a new function like `selectSubtitleTrackByExoId(addonTrackId)` that doesn't skip addon tracks, or adding a parameter to `selectSubtitleTrack` to include them, would fix this.

halibiram:
@skoruppa  I've fixed the issue, could you please take a look? #506

rubkn:
> @skoruppa I've fixed the issue, could you please take a look? #506

This is not ethical at all.

halibiram:
> This is not ethical at all.

Could you please elaborate on why you think this is not ethical? I'd really like to understand your perspective.

rubkn:
> > This is not ethical at all.
> 
> Could you please elaborate on why you think this is not ethical? I'd really like to understand your perspective.

This is an open PR. I'm working on it. So I'm expected a chance to develop on it. Opening a separate "fix" (which is actually a whole new implementation) PR without aligning with me first feels like you're just hijacking or credit grabbing. You are and were welcome to comment and make suggestions but what you did just duplicated efforts and your implementation might be the best way to go so my work was just for nothing.

halibiram:
> This is an open PR. I'm working on it. So I'm expected a chance to develop on it. Opening a separate "fix" (which is actually a whole new implementation) PR without aligning with me first feels like you're just hijacking or credit grabbing. You are and were welcome to comment and make suggestions but what you did just duplicated efforts and your implementation might be the best way to go so my work was just for nothing.

The reason I opened a new PR is that my solution takes a completely different technical approach. My only goal was to completely eliminate the issue. However, you are right about one thing: it would have been better if I had consulted you before opening the PR.

rubkn:
> > This is an open PR. I'm working on it. So I'm expected a chance to develop on it. Opening a separate "fix" (which is actually a whole new implementation) PR without aligning with me first feels like you're just hijacking or credit grabbing. You are and were welcome to comment and make suggestions but what you did just duplicated efforts and your implementation might be the best way to go so my work was just for nothing.
> 
> The reason I opened a new PR is that my solution takes a completely different technical approach. My only goal was to completely eliminate the issue. However, you are right about one thing: it would have been better if I had consulted you before opening the PR.

I get that. We all want the same goal. But the way you did it is unethical.

skoruppa:
> @skoruppa I've fixed the issue, could you please take a look? #506

I'm not sure how is your implementation better. In the end you still "slightly delay playing the video until the subtitle links are scraped" and just offload subtitle downloading to a different service. I think that using a http server just unnecessary complicates everything

halibiram:
> > @skoruppa I've fixed the issue, could you please take a look? #506
> 
> I'm not sure how is your implementation better. In the end you still "slightly delay playing the video until the subtitle links are scraped" and just offload subtitle downloading to a different service. I think that using a http server just unnecessary complicates everything 

I understand your concern about the HTTP server adding complexity. However, the main advantage of this implementation is the parallel processing.
​While the video is doing its initial buffering to load, all the subtitle files are being downloaded simultaneously in the background. Because of this, we aren't really adding a new delay; we are utilizing the existing initial load time.
​The biggest benefit is that once the movie starts, if the user decides to switch to a different subtitle variant or another language, there is absolutely no need to wait for the video to re-buffer or restart. It switches instantly without ruining the experience.

tapframe:
Let me know when this is ready to merge.

rubkn:
@tapframe everything working properly now, I tested it and it's working as expected but please do test it again to make sure

tapframe:
Works as expected. Merging.

JFeverish:
One question, if one is in Fast Startup and choose a subtitle from a language, let say English, or if one choose a language outside of the prefered languages, let says English is outside of the prefered languages.

Does it fetch in that moment all english subtitles? So if he decides to choose another english subtitle it wont reload again? 

Either way it's already an amazing implementation right now.

rubkn:
> One question, if one is in Fast Startup and choose a subtitle from a language, let say English, or if one choose a language outside of the prefered languages, let says English is outside of the prefered languages.
> 
> Does it fetch in that moment all english subtitles? So if he decides to choose another english subtitle it wont reload again?
> 
> Either way it's already an amazing implementation right now.

Right now on `Fast Startup` the expected behaviour is one reload when the first chosen subtitle gets selected. But that reload fetches all addon subtitles so subsequent changes don't trigger a reload.

JFeverish:
> > One question, if one is in Fast Startup and choose a subtitle from a language, let say English, or if one choose a language outside of the prefered languages, let says English is outside of the prefered languages.
> > Does it fetch in that moment all english subtitles? So if he decides to choose another english subtitle it wont reload again?
> > Either way it's already an amazing implementation right now.
> 
> Right now on `Fast Startup` the expected behaviour is one reload when the first chosen subtitle gets selected. But that reload fetches all addon subtitles so subsequent changes don't trigger a reload.

That's cool. And if one is in prefered language, let say pt-pt, and decides do select an English one? Does it also fetch all addon subtitles?

rubkn:
> > > One question, if one is in Fast Startup and choose a subtitle from a language, let say English, or if one choose a language outside of the prefered languages, let says English is outside of the prefered languages.
> > > Does it fetch in that moment all english subtitles? So if he decides to choose another english subtitle it wont reload again?
> > > Either way it's already an amazing implementation right now.
> > 
> > 
> > Right now on `Fast Startup` the expected behaviour is one reload when the first chosen subtitle gets selected. But that reload fetches all addon subtitles so subsequent changes don't trigger a reload.
> 
> That's cool. And if one is in prefered language, let say pt-pt, and decides do select an English one? Does it also fetch all addon subtitles?

Yep, same logic. Preferred subtitle won't reload, but any subtitle outside of the preferred language will reload the first time, and subsequent ones won't reload.

JFeverish:
> > > > One question, if one is in Fast Startup and choose a subtitle from a language, let say English, or if one choose a language outside of the prefered languages, let says English is outside of the prefered languages.
> > > > Does it fetch in that moment all english subtitles? So if he decides to choose another english subtitle it wont reload again?
> > > > Either way it's already an amazing implementation right now.
> > > 
> > > 
> > > Right now on `Fast Startup` the expected behaviour is one reload when the first chosen subtitle gets selected. But that reload fetches all addon subtitles so subsequent changes don't trigger a reload.
> > 
> > 
> > That's cool. And if one is in prefered language, let say pt-pt, and decides do select an English one? Does it also fetch all addon subtitles?
> 
> Yep, same logic. Preferred subtitle won't reload, but any subtitle outside of the preferred language will reload the first time, and subsequent ones won't reload.

Awesome...

PR #275, opened 2026-02-23
Create strings

affcosta7:
added portuguese (Portugal)


rubkn:
merge #273 and #274 with this

PR #280, opened 2026-02-23
feat: add loading indicators for addon filter chips

tsoltan:
Added a loading state to the addon filter chips in both the stream screen and player side panels to provide visual feedback while addons are being fetched. This improves user experience by indicating that more addons may be loading.
￼￼￼
image
image
image
image

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/StreamComponents.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/StreamComponents.kt
@@ -178,7 +178,7 @@ internal fun AddonFilterChips(
         ) {
             item {
                 AddonChip(
-                    name = "All",
+                    name = stringResource(R.string.stream_chip_all),
                     isSelected = selectedAddon == null,
                     onClick = { onAddonSelected(null) }
                 )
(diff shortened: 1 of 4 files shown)


skoruppa:
I was about to start implementing this! As sometimes I want to see if anything more is loading :)

But what if I have more addons and it takes whole space on the right next to "All"? Maybe we should have this indicator bellow?

Also, I was going to implement what addons are still loading (when only 1-2 are left), like in Stremio. It would be nice to have this here, but would require a bigger rewrite

tsoltan:
> I was about to start implementing this! As sometimes I want to see if anything more is loading :)
> 
> But what if I have more addons and it takes whole space on the right next to "All"? Maybe we should have this indicator bellow?
> 
Moved text to next line - in both places
image

> Also, I was going to implement what addons are still loading (when only 1-2 are left), like in Stremio. It would be nice to have this here, but would require a bigger rewrite

A simple indicator that more add-ons are loading should be enough for now. The full Stremio-like approach would likely need a bigger rewrite, which I’m not planning to work on at this time.

tsoltan:
already implemented by dev

PR #308, opened 2026-02-24
Add Turkish (tr) translation

halibiram:
I translated the entire app into Turkish (`values-tr/strings.xml`). 

I also added "Turkish" to the in-app Language picker (in ThemeSettingsScreen) and registered `tr` to `locale_config.xml`. 

I've tested it on my end and adjusted a few TV-specific terms to keep the UI clean. Let me know if everything looks good!


halibiram:
Updated the `values-tr/strings.xml` to include all the new keys introduced recently (like the ones in PR #291). The Turkish translation is now 100% synced with the main branch. Thanks!

PR #317, opened 2026-02-24
Update ThemeSettingsScreen.kt

traidor:
Added ro in Language Picker


skoruppa:
You don't do one pull request per file. Please delete all your pull requests and create one with all changes and correct description and tittle

PR #264, opened 2026-02-23
i18n support with Polish translation

skoruppa:
Adds proper internationalization support to the app as well a language selector in Appearance settings. Using Polish as the first additional language.

Because application needs to be "**closed**" to apply selected language (if user want different one from the system ui), it would be good to merge #195 with it 

# To add new language
- copy `app/src/main/res/values/strings.xml` into `app/src/main/res/values-`**code**`/strings.xml`
  - For language only: use ISO 639-1 code, e.g. `values-pt/`
  - For language + region: use `language-rREGION` format, e.g. `values-pt-rBR/`
- add new language to `app/src/main/res/xml/locale_config.xml`
- add new language to `supportedLocales` in `ThemeSettingsScreen.kt`:
  - language only: `"pt" to "Português"`
  - with region: `"pt-BR" to "Português (Brasil)"`
```
    val supportedLocales = remember {
        listOf(
            null to "System default",
            "en" to "English",
            "pl" to "Polski"
        )
    }
```
- translate your new `strings.xml`

# i18n Progress

### UI Components
- [x] `ui/components/ContinueWatchingSection.kt`
- [x] `ui/components/ErrorState.kt`
- [x] `ui/components/CatalogRowSection.kt`
- [x] `ui/components/HeroCarousel.kt`
- [x] `ui/components/NuvioDialog.kt`
- [x] `ui/components/SidebarNavigation.kt`
- [x] `ui/components/EmptyScreenState.kt`

### Home
- [x] `ui/screens/home/HomeScreen.kt`
- [x] `ui/screens/home/GridHomeContent.kt`
- [x] `ui/screens/home/ModernHomeContent.kt`
- [x] `ui/screens/home/ClassicHomeContent.kt`

### Detail
- [x] `ui/screens/detail/HeroSection.kt`
- [x] `ui/screens/detail/EpisodesSection.kt`
- [x] `ui/screens/detail/CastSection.kt`
- [x] `ui/screens/detail/MoreLikeThisSection.kt`
- [x] `ui/screens/detail/MetaDetailsScreen.kt`

### Settings
- [x] `ui/screens/settings/SettingsScreen.kt`
- [x] `ui/screens/settings/PlaybackSettingsSections.kt`
- [x] `ui/screens/settings/PlaybackSettingsScreen.kt`
- [x] `ui/screens/settings/PlaybackAudioSettings.kt`
- [x] `ui/screens/settings/PlaybackSubtitleSettings.kt`
[...]

PR #320, opened 2026-02-24
Add Romanian translation

traidor:
-Created values-ro/strings.xml
-Edited ThemeSettingsScreen.kt
-Edited locale_config.xml


halibiram:
> -Created values-ro/strings.xml -Edited ThemeSettingsScreen.kt -Edited locale_config.xml

Please avoid modifying the main strings.xml file directly. You should add the new values to the base file and then provide a separate localization file for your translations. Please close this PR and take a look at PR #264 as a reference for the correct structure.

PR #330, opened 2026-02-24
fix(home): make modern home screen feel snappier

halibiram:
I've made a few adjustments to the modern home screen to improve the overall feel and performance:

- **FPS Boost**: Switched `activeRowIndex` and similar lookups to use `derivedStateOf`. This avoids re-scanning the row list on every recomposition, which helps keep the CPU load down when the list gets long.
- **Better Responsiveness**: Reduced the hero focus debounce to 16ms. The UI reacts much faster to remote inputs now.
- **Faster Transitions**: Shortened the background and trailer fade durations. Long animations can make the system feel a bit sluggish on TVs, so I tightened these up to make the navigation feel lighter and more fluid.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeContent.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeContent.kt
@@ -467,6 +467,15 @@ fun ModernHomeContent(
             }
         }
     }
+    // OPTIMIZATION: Cache row index lookup to avoid repeated map access
+    val activeRowIndex by remember(activeRow) {
+        derivedStateOf {
+            activeRow?.let { row ->
+                carouselRows.indexOfFirst { it.key == row.key }.takeIf { it >= 0 }
+            } ?: -1
+        }
+    }
+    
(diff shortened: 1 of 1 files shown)


tapframe:
can you share a before and after video on how it performs now, thank you.

halibiram:
before:
https://drive.google.com/file/d/1qNShjnREsfSj_NIXpiwnTtwJtucoVQNn/view?usp=drive_link

After:
https://drive.google.com/file/d/1YJat50gsiXPSADiu6542qldS96bsHa3J/view?usp=drive_link

tfouto:
It would be usefull for someone to try on a slower device.

skoruppa:
> It would be usefull for someone to try on a slower device.

I'm this someone :D using the same technique I used for #193

**Before**: 100% janky frames - modern view works terrible on my tv
**After**: 100% janky frames - I don't see any improvements

The modern view is interesting, because when nothing "moves" on it, it still slows down the whole app - even menu navigation start to fell sluggish with it. Something is probably causing recomposition all the time, even if no focus is being changed. I tried to analyze that but with my limited skills I didn't find anything

PR #338, opened 2026-02-25
Fix subtitle addons - wrong content ID, addon tracks in built-in list, videoHash/videoSize/filename pipeline

skoruppa:
Fix subtitle addons
- Addon subtitles no longer appear in the built-in subtitles tab (ExoPlayer prepends group index to track IDs, fixed startsWith -> contains)
- Fixed wrong content ID being sent to subtitle addons
- Added videoHash/videoSize/filename pipeline from stream to subtitle requests; OpenSubtitlesHasher computes hash on-the-fly when not provided by stream

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ scripts
 supabase
 tv-samples
 HOME_UI_OPTIMIZATION_NOTES.local.md
+.kotlin/
--- a/app/src/main/java/com/nuvio/tv/core/player/OpenSubtitlesHasher.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/OpenSubtitlesHasher.kt
@@ -1,0 +1,91 @@
+package com.nuvio.tv.core.player
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.InputStream
+import java.net.HttpURLConnection
+import java.net.URL
+
+/**
+ * Calculates the OpenSubtitles file hash for a video stream.
+ * Algorithm: hash = fileSize + sum of first 64KB longs + sum of last 64KB longs (little-endian)
+ * https://trac.opensubtitles.org/projects/opensubtitles/wiki/HashSourceCodes
+ */
+object OpenSubtitlesHasher {
+
+    private const val CHUNK_SIZE = 65536L // 64 KB
+    private const val LONG_SIZE = 8
+
+    data class Result(val hash: String, val fileSize: Long)
+
+    suspend fun compute(url: String, headers: Map<String, String>): Result? =
+        withContext(Dispatchers.IO) {
+            try {
+                val fileSize = getContentLength(url, headers) ?: return@withContext null
+                if (fileSize < CHUNK_SIZE * 2) return@withContext null
+
+                var hash = fileSize
+                hash += readChunkSum(url, headers, offset = 0, length = CHUNK_SIZE)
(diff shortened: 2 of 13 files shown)


tapframe:
What does this PR fixes, and how do I test it? thanks

skoruppa:
I'm the developer of https://github.com/skoruppa/stremio-community-subtitles. I noticed incorrect behavior when watching anime (MAL ID) and using my addon: subtitles were returned for the wrong title and addon's dashboard entries pointed to incorrect content


### What it fixes
- Stops unnecessary recompilation of `contentId` (content+season+episode). The original full `contentId` is now forwarded to add-ons. Some content ids don't have season in them
- Properly forwards `videoHash`, `videoSize`, and `filename` from [streams](https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/responses/stream.md) to subtitle add-ons
- If those values are not provided by stream objects, generates them (OpenSubtitles-compatible hasher) and [forwards](https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/requests/defineSubtitlesHandler.md) (extra parameters) them
- Prevents add-on subtitles from appearing in the `"embed"` tab (ExoPlayer treated selected external subtitles as embedded)

### How to test
1. Install my add-on.
2. Play something on Nuvio **without** this fix:
   - `videoHash` and `filename` are missing in the add-on dashboard
   - Selecting add-on subtitles makes them appear in the `"embed"` tab
3. Apply this PR and test again:
   - `videoHash`, and `filename` are present.
   - Add-on subtitles no longer appear in the `"embed"` tab after selecting them

rubkn:
@skoruppa does this help in choosing the subtitles for the determined video? For example we might need to adjust some subtitles to a specific video.

skoruppa:
> @skoruppa does this help in choosing the subtitles for the determined video? For example we might need to adjust some subtitles to a specific video.

It just align subtitles addons behavior to how they work in Stremio. Nothing more

tfouto:
Would it be possible to show subtitles names to the user? With would help users tremendously in choosing the right one, instead of trying blindly.

skoruppa:
> Would it be possible to show subtitles names to the user? With would help users tremendously in choosing the right one, instead of trying blindly.

Well, we could extend subtitle addons with a new field, but addons would need to use it :P And I doubt anyone right now would care about this. Maybe in future, when Nuvio will be stable and reach feature parity, we can think about making changes in addons

tfouto:
> > Would it be possible to show subtitles names to the user? With would help users tremendously in choosing the right one, instead of trying blindly.
> 
> Well, we could extend subtitle addons with a new field, but addons would need to use it :P And I doubt anyone right now would care about this. Maybe in future, when Nuvio will be stable and reach feature parity, we can think about making changes in addons

Since you already develop https://github.com/skoruppa/stremio-community-subtitles, you addon could extend that field. That way your addon would be in advantage. Maybe others would follow along.

skoruppa:
> > > Would it be possible to show subtitles names to the user? With would help users tremendously in choosing the right one, instead of trying blindly.
> > 
> > 
> > Well, we could extend subtitle addons with a new field, but addons would need to use it :P And I doubt anyone right now would care about this. Maybe in future, when Nuvio will be stable and reach feature parity, we can think about making changes in addons
> 
> Since you already develop https://github.com/skoruppa/stremio-community-subtitles, you addon could extend that field. That way your addon would be in advantage. Maybe others would follow along.

Maybe. But this is a discussion for other time. Not this PR :p

PR #341, opened 2026-02-25
Refactoring Modern Home Layout to use LazyColumn instead of custom scroll animation (WIP)

tapframe:
optimzing modern layout to reduce jank and scroll perfomance by using LazyColumn (WIP)

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeRows.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeRows.kt
@@ -5,6 +5,7 @@ package com.nuvio.tv.ui.screens.home
 import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
+import androidx.compose.animation.core.AnimationSpec
 import androidx.compose.foundation.gestures.BringIntoViewSpec
 import androidx.compose.foundation.gestures.LocalBringIntoViewSpec
 import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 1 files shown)


tapframe:
@skoruppa would like to know how it performs compared to previous state as it stands. noticeable difference, if any.

skoruppa:
> @skoruppa would like to know how it performs compared to previous state as it stands. noticeable difference, if any.

According to gfxinfo it is _slightly_ better
```
Total frames rendered: 908
Janky frames: 1004 (97.86%)
Janky frames (legacy): 1004 (97.86%)
```
but nothing noticeable by user. This may even be measurement error. Each animation feels sluggish - even navigating in classic menu. I feel like the culprit here is hero backdrop. It somehow triggers recomposition of whole screen with each animation frame - I will investigate this a bit, maybe I will find something. But I agree we should use lazyColumn instead of custom scroll animation

halibiram:
Unfortunately, it didn't provide the expected performance boost. There is continuous stuttering (jank) in almost every frame that is straining the system. The scrolling experience doesn't feel smooth.


* The latest version
photo_5834575624532594578_w

*After the LazyColumn change
photo_5834575624532594582_w

*Other layouts
photo_5834575624532594579_w

skoruppa:
> Unfortunately, it didn't provide the expected performance boost. There is continuous stuttering (jank) in almost every frame that is straining the system. The scrolling experience doesn't feel smooth.
> 
>     * The latest version
>       photo_5834575624532594578_w
> 
> 
> *After the LazyColumn change photo_5834575624532594582_w
> 
> *Other layouts photo_5834575624532594579_w

I noticed improvements after I removed the black gradient on the left, but don't know why...it should not change anything in performance.

How did you turn on those graphs? Would be helpful when investing

halibiram:
> How did you turn on those graphs? Would be helpful when investing

Developer options>profile hwui rendering>on screen as bars

PR #369, opened 2026-02-25
enhanced regex auto-stream select

### LINKED ISSUE #368: [Bug] Only first link generated by regex is passed
If the the first link generated by regex is not  resolving (404) then auto-play fails and the user is redirected to the individual links for manual selection.

We need for the app to collect all regex passed links and test the first link for resolution before passing it to the player and move on to the next regex passed link if the 1st fails to resolve.

craitose:
fixes #368 

1. Added regex preset to exclude remux/hdr.
2. Added functionality for the links that pass the regex test to be collected and if the first link returns 404 then test the next working link before sending to the player

the 2nd point solves the issue where if the first link that passed the test was not available from the host (403),  then the autoplay will fail


Nazim1981:
Does the app now sequentially validate links until it finds the first playable one, or does it only attempt the second link if the first returns 404?

Please add better link transparency.

When autoplay fails, and the manual link screen appears, the app should clearly show:

* Which links were already tried and failed
* Which link is currently being attempted
* Which link is successfully playing

Failed links should be clearly marked, and the currently playing link should be highlighted (ideally showing the full URL).

Right now, there’s no way to know which link was tried, which failed, or which is playing, which causes repeated attempts on the same links.

craitose (commented):
Added the "skip regex search if regex pattern blank" to avoid merge conflicts

PR #388, opened 2026-02-26
skip regex search if pattern is null to facilitate binge group speed 

### LINKED ISSUE #351: When using the Auto Stream Selection with auto play regex match.  The  finding stream source takes too long to load.
The current option of Auto Play regex match, with auto play next episode option, prefer binge group options are on. 

Takes 30sec to 1 minute. 

image

https://github.com/user-attachments/assets/4235cc11-158e-446d-953b-ea6e3be60813

Regularly selecting a source with the feature off loads up right away.

craitose:
Fixes #351 

In a scenario where users select regex to auto select a stream but don't supply a pattern - the regex function will conduct a search which can take up to a minute (depending on the source scope).

This can become an issue when the user has episode auto-play enabled coupled with binge group functionality.
The next binge group will only be selected after the regex search completes.

By skipping regex search if the pattern is empty , the next binge group episode should load as fast as the source addon/plugin allows.


ram130:
Not working with new update with the merge. I replied on the main issue thread.  

https://github.com/user-attachments/assets/95d9e239-78fe-4ddb-8d9c-f7bfcf3c4aa2

PR #394, opened 2026-02-26
Add OSD clock format setting (12-hour/24-hour)

### LINKED ISSUE #344: [Feature] Option to Select Hour Format for OSD Clock in Video Player
Adding an option in settings to select either 12-hour or 24-hour time format in the internal player.

Currently it's only 24-hour format.

Image

halibiram:
Add OSD clock format setting (12-hour/24-hour)

- Add OsdClockFormat enum with HOUR_12 and HOUR_24 options
- Add clock format preference to PlayerSettingsDataStore
- Update PlayerScreen and PauseOverlay to respect format setting
- Add collapsible clock format selector in playback settings (similar to Auto Frame Rate)
- Add localized strings for all supported languages (en, tr, fr, it, pl, pt-PT, ro, sk, sl)
- Remove unused OsdClockFormatOptions function and fillMaxSize import
- Improve Turkish string consistency (standardize verbs to present tense)

Fixes #344
Screenshot_20260226_193054

Screenshot_20260226_192955

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackSettingsSections.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackSettingsSections.kt
@@ -10,7 +10,6 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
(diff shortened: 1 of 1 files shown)


skoruppa:
There is also a clock on the "pause overlay" that is displayed after 5s from pausing

halibiram:
I've added that part too. Thank you for the feedback!

Screenshot_20260226_230802

skoruppa:
> I've added that part too. Thank you for the feedback!

I'm still not sure if we need a whole new setting for this.... 
In #412 I changed clocks to follow system format as this will be something that most user will want

halibiram:
> > I've added that part too. Thank you for the feedback!
> 
> I'm still not sure if we need a whole new setting for this.... In #412 I changed clocks to follow system format as this will be something that most user will want

Your approach makes more sense, you solved the issue with just 4 lines. It's more appropriate to follow the local time format; I also agree that users wouldn't feel the need to choose a format other than what they're used to. Therefore, I am closing this PR.

JFeverish:
> > I've added that part too. Thank you for the feedback!
> 
> I'm still not sure if we need a whole new setting for this.... In #412 I changed clocks to follow system format as this will be something that most user will want

I prefer the english language and the 22:15 format. But I understand the simplicity of this approach.

PR #409, opened 2026-02-26
feat: fix localization issues and improve translations

### LINKED ISSUE #402: Please update the Portuguese (Portugal) translation of Nuvio
Hi there
Great implementation of the translation of the app!
I don't mean to be picky but the Portuguese (Portugal) translation of the app contains a few mistakes and mistranslations and "contaminations" of Brazilian Portuguese. For example, director is translated to "Director" when it should be "Realizador". The "Play" button was not translated. I really don't mean to be rude and to disrespect the translation of the original translator, but I've translated the file myself. I don't know how to update and pull request so I'll just attach the file here, hoping someone will update it 😊

[strings.xml](https://github.com/user-attachments/files/25582821/strings.xml)

halibiram:
Fixed several localization bugs:
- Play button text now updates correctly when changing language
- Home screen category names are now properly translated
- Added missing translations for catalog names across all languages

Improvements:
- Added translations for 50+ common catalog names (Popular, Trending, New, Top Rated, etc.)
- Catalog names now translate in Modern, Classic, and Grid layouts
- TMDB language automatically syncs with app language selection
- Applied Portuguese (Portugal) translation improvements (fixes #402)

The play button issue was caused by @ApplicationContext being created at app startup with the initial locale. Fixed by adding a helper function that reads the current locale from SharedPreferences and creates a fresh context for each string fetch.

For catalog translations, added a translateCatalogName() function that normalizes catalog names and maps them to localized strings. This works for streaming addons (Movies, Series, 4K), debrid services, and anime catalogs (Top Airing, Most Popular, etc.).

Portuguese improvements include changing from formal to informal language, replacing technical terms (complemento→addon, extensão→plugin), and fixing 766+ strings for more natural European Portuguese.

Fixes #402

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
@@ -81,6 +81,37 @@ class MetaDetailsViewModel @Inject constructor(
 
     private var isPlayButtonFocused = false
 
+    /**
+     * Gets a localized string with the current locale configuration.
+     * This ensures that strings are always fetched with the up-to-date locale,
+     * even if the app language was changed after the ViewModel was created.
+     */
+    private fun getLocalizedString(resId: Int, vararg formatArgs: Any?): String {
+        // Read the current locale from SharedPreferences
+        val prefs = context.getSharedPreferences("app_locale", Context.MODE_PRIVATE)
+        val localeTag = prefs.getString("locale_tag", null)
+        
+        return if (!localeTag.isNullOrEmpty()) {
+            // Create a new configuration with the selected locale
+            val locale = java.util.Locale.forLanguageTag(localeTag)
+            val config = android.content.res.Configuration(context.resources.configuration)
+            config.setLocale(locale)
+            val localizedContext = context.createConfigurationContext(config)
+            if (formatArgs.isEmpty()) {
+                localizedContext.getString(resId)
+            } else {
+                localizedContext.getString(resId, *formatArgs)
+            }
+        } else {
+            // Use system default
+            if (formatArgs.isEmpty()) {
+                context.getString(resId)
+            } else {
+                context.getString(resId, *formatArgs)
+            }
+        }
+    }
+
     init {
         observeMetaViewSettings()
         observeTrailerAutoplaySettings()
@@ -747,7 +778,7 @@ class MetaDetailsViewModel @Inject constructor(
                         nextVideoId = meta.id,
                         nextSeason = null,
                         nextEpisode = null,
-                        displayText = context.getString(R.string.detail_btn_resume)
+                        displayText = getLocalizedString(R.string.detail_btn_resume)
                     )
                 } else {
                     NextToWatch(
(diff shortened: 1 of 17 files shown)


rubkn:
I disagree with this. If if's localization why are we allowing foreign words? Are we expecting portuguese users to know english? If so why have localization at all. There is available translation for those words that you said were "techical terms" so why use english?

More info in my comment on the related issue: https://github.com/tapframe/NuvioTV/issues/402#issuecomment-3969767966

Also, I saw that you did the PR for the turkish translation, so are you bilingual in turkish and portuguese from portugal?

app/src/main/res/values-fr/strings.xml
rubkn: You should leave the translations to someone who is native on the language.

app/src/main/res/values-pl/strings.xml
rubkn: Same here, you should leave the translations to someone who is native on the language. I believe @skoruppa is the man for the job here.

app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
```diff
@@ -81,6 +81,37 @@ class MetaDetailsViewModel @Inject constructor(
...
+            }
+        } else {
+            // Use system default
+            if (formatArgs.isEmpty()) {
+                context.getString(resId)
+            } else {
+                context.getString(resId, *formatArgs)
+            }
+        }
+    }
```
rubkn: What is this for? I believe this is just duplicating the way the app handles locales? Plus, even if this is strictly needed since this is not a composable, we should move this into a single shared provider abstraction instead of per-file locale helpers.

app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeModels.kt
```diff
@@ -288,6 +299,64 @@ internal fun catalogRowTitle(
...
+        "trending series", "trend diziler" -> context.getString(R.string.catalog_trending_series)
+        "trending shows", "trend programlar" -> context.getString(R.string.catalog_trending_shows)
+        "best", "en iyi" -> context.getString(R.string.catalog_best)
+        "top", "en üst" -> context.getString(R.string.catalog_top)
+        "all", "tümü" -> context.getString(R.string.catalog_all)
+        
+        else -> catalogName // Return original if no translation found
+    }
+}
+
```
rubkn: Again, this whole block is just so weird. Besides introducing hard-coded translation aliases, it creates again a new localization path in this file only.

halibiram:
Hey @rubkn , thanks for being so direct. You're completely right on all points. I definitely shouldn't have touched the Portuguese translations or any language I don't speak. I relied on AI/translators to fill in the gaps and it obviously didn't work out. I'll leave those entirely to native speakers like @skoruppa from now on.
As for the code, yeah, it's a mess. I was too focused on just making the UI display the text and ended up hacking the logic directly into ThemeModel and duplicating the locale stuff. It was a bad approach and it clearly breaks the architecture.
I'm going to close this PR now. I need to spend more time actually reading and understanding the codebase before making big changes.
Sorry for wasting your time reviewing this one. I'll do better.

PR #448, opened 2026-02-28
Auto-load discover results near the end

SwoleRonson:
## Summary
  - Auto-load additional Discover results when the grid gets near the
  end
  - Keep the existing Load More / Show More action card as a fallback


SwoleRonson:
https://github.com/user-attachments/assets/89103015-747f-44e7-ab4b-2803de8ce0a6

Preview of the autoloading, including throttling network speeds.

skoruppa:
This is already implemented :)

PR #506, opened 2026-03-01
Fix: Eliminate video rebuffering to make subtitle switching instant

halibiram:
**The Problem:**
Changing addon subtitles was super annoying. Every time you picked a new language, the app would restart the whole video stream (`setMediaSource`) just to load the new subtitle URL. This completely wiped the ExoPlayer buffer, causing the video to freeze, re-buffer, and sometimes even jump back to an old saved position. It felt really clunky and completely interrupted the watch experience.

**The Fix / Solution:**
I totally reworked how we handle addon subtitles so switching them is instantly smooth, just like embedded tracks. 

Here's how it works now:
- Spun up a tiny local HTTP server (`SubtitleServer` via `NanoHTTPD`) inside the app.
- Built a background downloader (`SubtitleCache`) that quietly grabs *all* available addon subtitles the moment the video player opens.
- When we first init the player, we toss all those subtitles into ExoPlayer using our local `127.0.0.1:8155` proxy links. We slightly delay playing the video until the subtitle links are scraped.
- Now, when you select a subtitle from the UI, we don't rebuild the media source at all. We just tell ExoPlayer to switch text tracks via `TrackSelectionParameters`. ExoPlayer hits our local proxy, which instantly serves the pre-downloaded file from the disk cache.

Zero buffering. Zero video jumping. Instant subtitle switching.

fixes: #265

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/player/SubtitleCache.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/SubtitleCache.kt
@@ -30,7 +30,7 @@ class SubtitleCache(context: Context) {
         private const val TAG = "SubtitleCache"
         private const val CACHE_DIR_NAME = "subtitles"
         private const val MAX_CACHE_SIZE_BYTES = 100L * 1024 * 1024 // 100 MB
-        private const val DOWNLOAD_TIMEOUT_SECONDS = 10L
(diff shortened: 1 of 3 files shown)


halibiram:
Since PR #272 has been successfully merged, this pr can now be closed.

PR #507, opened 2026-03-01
Move addons into Settings and enforce profile-based access UI

Chanceium:
- Remove Addons from sidebar and root routes
- Add Addons section inside Settings with embedded AddonManager content
- Hide Addons/Plugins sections when active profile inherits primary resources
- Guard direct Addons/Plugins routes and redirect to Settings when blocked
- Align embedded Addons layout with Plugins spacing and chrome
- Apply global green active / gray inactive styling for settings chips and toggles
- Enable debuggable debug build and default debug signing behavior


Chanceium:
Addons removed from main
image
Addons inside settings
image
Ticks are green when enabled
image
Changed profile to one inheriting addons
image

Chanceium:
image
fixed chips not being consisent

RSAK2105:
@Chanceium 

It looks good and tidy. Hope, it will be effective in Modern Sidebar too.

I have reported some layout misalignment here. Pls check it if have time to fix.

https://github.com/tapframe/NuvioTV/issues/535

PR #515, opened 2026-03-02
change focus to nextToWatch season, and scroll to nextToWatch episode

skoruppa:
Simple change so that when scrolling down on meta view, when nextToWatch is filled, season is focused on a correct one, and episode is scrolled to the next one


tapframe:
was literally working on this as I wrote this. Good qol implementation.

PR #550, opened 2026-03-03
Dutch translations

LastpostNL:
Hi, not sure if this is the way to add it. But here are the Dutch translations (Netherlands) for the NuvioTV app.


halibiram:
> Hi, not sure if this is the way to add it. But here are the Dutch translations (Netherlands) for the NuvioTV app.

Thanks for contributing this new translation! 

To make sure the new language is fully recognized by the system and appears in the app's language selection menu, it needs to be registered in two additional places.

Could you please add the new locale code to the following files as well?

- `app/src/main/res/xml/locale_config.xml` (This is required for Android's per-app language preferences).
- The `supportedLocales` list inside `ThemeSettingsScreen.kt`  (So it shows up in our Compose UI settings).

I highly recommend reviewing PR #440  for reference.

PR #558, opened 2026-03-03
Fix: preserve local scraper headers across HLS redirects

### LINKED ISSUE #549: Local scraper headers not sent correctly to HLS CDN (403 Forbidden on playback)
## Description

Headers returned by local scrapers in the `headers` field of the stream object are not being correctly forwarded to the CDN when playing HLS streams, resulting in **403 Forbidden** errors on playback.

## Steps to Reproduce

1. Install a local scraper that returns an HLS stream with custom `headers` (Referer, Origin, User-Agent)
2. Select a stream from that scraper
3. Player attempts to load the m3u8 → **403 Forbidden**

## Evidence

The scraper returns a valid stream object with correct headers:

```json
{
  "name": "La.Movie",
  "url": "https://hls1.goodstream.one/hls2/...master.m3u8?t=...&e=43200",
  "quality": "Auto",
  "headers": {
    "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
    "Referer": "https://goodstream.one/embed-sgf4johjlyvz.html",
    "Origin": "https://goodstream.one"
  }
}
```

The m3u8 URL is accessible with those exact headers — verified via Node.js test:

```
✓ [200] Referer: embed URL (exact)   → Valid m3u8, 3 variants
✓ [200] Referer: goodstream.one/     → Valid m3u8, 3 variants
✓ [200] No Referer (bare UA)         → Valid m3u8, 3 variants
✗ [403] No headers at all            → Forbidden
```

**Conclusion:** The CDN requires at minimum a valid `User-Agent`. Without it → 403. The scraper provides it correctly, but Nuvio's player does not forward it.

## Root Cause (from source code)

In `PlayerMediaSourceFactory.kt`:

```kotlin
val okHttpFactory = OkHttpDataSource.Factory(getOrCreateOkHttpClient()).apply {
    setDefaultRequestProperties(sanitizedHeaders) // ← headers set here
    setUserAgent(                                 // ← UA overwritten here
        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) " +
        "AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
    )
}
```

Two issues:

1. **`setUserAgent()` is called after `setDefaultRequestProperties()`**, overwriting the `User-Agent` provided by the scraper with a hardcoded one. While in this case the UA happens to be the same, this is fragile and will break scrapers that require a different UA.

2. **It is unclear whether `Referer` and `Origin` from `setDefaultRequestProperties()` are forwarded for every HLS segment request**, or only for the initial manifest request. CDNs like Goodstream validate headers on every segment fetch, not just the manifest.

## Expected Behavior

All headers defined in the stream object should be forwarded for **every** request — manifest, variants, and segments.

## Suggested Fix

```kotlin
val okHttpFactory = OkHttpDataSource.Factory(getOrCreateOkHttpClient()).apply {
    setDefaultRequestProperties(sanitizedHeaders)
    // Only set UA if the scraper didn't provide one
    if (!sanitizedHeaders.containsKey("User-Agent")) {
        setUserAgent(
            "Mozilla/5.0 (Windows NT 10.0; Win64; x64) " +
            "AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
        )
    }
}
```

## Environment

- Nuvio TV (Android)
- Local scraper: La.Movie (ported from CloudStream)
- CDN: Goodstream (`hls1/hls2.goodstream.one`)
- Node.js v20.20.0 (used for testing)

tsureda1:
Fixes #549


trokpail:
hope this gets merged in main, it fixes some streams and providers for me :(, if not i would have to keep using the apk i built

tsureda1 (commented):
Nice

PR #564, opened 2026-03-03
feature/i18n web config

halibiram:
## Summary

- Made NanoHTTP server pages (Addon and Repository Web Pages) i18n compatible.
- Added Turkish translations for web configuration pages.
- Escaped single quotes in JavaScript web UI translated strings to prevent syntax errors.
- Ensure that the app-selected locale is used for NanoHTTPD web config pages instead of the device's default language.

## Why

The web configuration pages for addons and repositories were previously using hardcoded text and the device's default language. This PR ensures that the web pages respect the app's selected language instead, providing a consistent localized experience for users accessing the configuration pages via their browser.

## Testing

- Manually verified that accessing the NanoHTTPD web configuration pages uses the correct translated strings based on the app's selected language.
- Checked that changing the app language dynamically updates the web page language without needing a server restart.
- Verified that single quotes in translations are properly escaped in the generated JavaScript.

No screen changes were made.

## Breaking changes

No breaking changes were made.

## Linked issues

No linked issues.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/server/AddonConfigServer.kt
+++ b/app/src/main/java/com/nuvio/tv/core/server/AddonConfigServer.kt
@@ -1,5 +1,6 @@
 package com.nuvio.tv.core.server
 
+import android.content.Context
 import com.google.gson.Gson
 import com.google.gson.reflect.TypeToken
 import fi.iki.elonen.NanoHTTPD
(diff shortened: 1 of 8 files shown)


skoruppa:
I think we can finally merge this. Almost everything can be translated now, so the Web Helpers should also receive some love. 

@halibiram It's been some time since you proposed this, do you think there are any amendments required here?

halibiram:
> I think we can finally merge this. Almost everything can be translated now, so the Web Helpers should also receive some love. 
> 
> @halibiram It's been some time since you proposed this, do you think there are any amendments required here? 

Finally! I'll do one final review on my end and let you know right away

halibiram:
image

image

@skoruppa  It appears to be in sync with the app language.

PR #586, opened 2026-03-04
fix(ui): adjust subtitle tabs to be elastic and center aligned

saimuelbr:
## Summary

Elastic Tab: We implemented IntrinsicSize.Min and wrapContentWidth in the SubtitleTab component to allow tabs to automatically adjust to the string size (an essential change for longer strings like "Sincronização" in PT-BR).

Layout Alignment: I adjusted the main line in the legend dialog box to ensure correct center alignment (Arrangement.Center) of the tabs, eliminating the asymmetry of the edges that could be evident if the alignment were not implemented.

## Why

In languages ​​like Portuguese (pt-BR), words tend to be longer than in English (for example, "Sincronização" vs "Sync"). The previous layout used fixed sizes or forced distribution, which caused line breaks in tabs or excessive white space for shorter names, such as "Atraso". This change makes the subtitle panel interface more concise and adaptable to any language.

## Testing

Manual: Tested correctly by switching between PT-BR and EN-US to verify that there are no padding issues.

Focus navigation: Verified that D-Pad navigation remains fully functional after enabling `wrapContentWidth`.

EN:

image

PT-BR:

image

## Breaking changes

**None**

## Linked issues

Myself noticed this issue.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/SubtitleDialog.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/SubtitleDialog.kt
@@ -238,7 +238,7 @@ private fun SubtitleTab(
         Row(
             modifier = Modifier
                 .padding(horizontal = 12.dp, vertical = 10.dp)
-                .width(IntrinsicSize.Min), 
+                .width(IntrinsicSize.Max), 
             verticalAlignment = Alignment.CenterVertically,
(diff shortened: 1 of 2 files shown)


danilopagotto82:
In the modern visual menu, newly added items often get cut off even when resized. It would be more effective if the menu were elastic and could adapt automatically to the word length.

Issue #353, opened 2026-02-25
When exiting a tv show the app should return to the current season and episode the user was watching.


ram130:
This is a show called Sliders. I'm currently on S4/E15. I exited to pause where I'm at. App returns to the episode selection screen at Season 1 Ep 1. Should return to current season and episode in the list. 

https://github.com/user-attachments/assets/dba72dae-61ba-4cc3-bec1-3d85457505aa

timmyp53:
That'd be cool.

ram130:
Think you guys can look at this? @CrissZollo @craitose @rubkn

rubkn:
@ram130 does this still happen? I might take a look.

ram130:
> @ram130 does this still happen? I might take a look.

Yep still happening.

tapframe:
this shouldn't happen from 0.4.9-beta. are you still facing this?

ram130:
> this shouldn't happen from 0.4.9-beta. are you still facing this?

I can confirm it was fixed. Apologies forgot to report back.

PR #604, opened 2026-03-05
fix[#353]: preserve return focus for watched episode

rubkn:


rubkn:
Closing since issue was resolved.

Issue #587, opened 2026-03-04
[Bug]: Can't rewind with player ui hidden


Trulecenttt:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

Latest

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Onn 4k plus 

### Android version

Unknown

### Player mode

Internal player

### Area (tag)

Playback (start/stop/buffering)

### Steps to reproduce

Press left on dpad when In player and Player UI is hidden no longer rewinds the show.
Pressing Right correctly fastforwards

### Expected behavior

Left on DPAD should rewind

### Actual behavior

Doesn't Rewind as expected

### Frequency

Always

### Did this work before?

Yes, it used to work

### Logs (optional but helpful)

```shell

```

mistersandman-64:
Experiencing this issue as well on 0.4.10.

Also, clicking select on seek bar while player UI is shown no longer plays the video from that point. Have to actually click Play button now.

tapframe:
Please install the APK from the release again. The first build was broken, and I had to push a new build as mentioned in the release notes.

https://github.com/tapframe/Nuviotv/releases/#:~:text=Note%20for%20users,auto%20stream%20selection.

PR #608, opened 2026-03-05
feat(player): improve and add seek speed in holding

saimuelbr:
## Summary

**Progressive Seek Acceleration**: Implemented dynamic seek logic using `nativeKeyEvent.repeatCount`. The jump interval is now scaled based on the press duration, starting at 10 seconds and gradually increasing to 2-minute jumps (10s -> 15s -> 30s -> 1min -> 2min).

**Key Interception**: Updated the global `onKeyEvent` event to handle search events differently depending on the user interface state.

## Why

**Efficient Navigation**: Navigating through long movies or series was tedious with fixed 20/30-seconds timestamps. This acceleration allows users to quickly scroll through long timelines while maintaining precise control.

## Testing

**Stress Test**: Verified that the speed increase was achieved by holding down the right directional pad for more than 10 seconds; the search correctly reached the 2-minute jump limit.

**Visual Synchronization**: Verified that the search overlay correctly reflects the accelerated time jumps in real time.

https://github.com/user-attachments/assets/f9bf4ac7-85c4-41d9-a085-52dc42484ffb

## Breaking changes

**None**.

## Linked issues

image

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerScreen.kt
@@ -406,14 +406,31 @@ fun PlayerScreen(
                             if (!uiState.showControls) {
                                 val repeatCount = keyEvent.nativeKeyEvent.repeatCount
                                 val deltaMs = when {
-                                    repeatCount >= 8 -> 30_000L
-                                    repeatCount >= 3 -> 20_000L
+                                    repeatCount >= 30 -> 120_000L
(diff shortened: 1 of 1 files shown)


skoruppa:
If you are already touching it, maybe fix #587 ;)?

tapframe:
> If you are already touching it, maybe fix #587 ;)?

No need. The fix was already applied and pushed with the updated APK.

https://github.com/tapframe/NuvioTV/issues/587#issuecomment-4009198422

Issue #602, opened 2026-03-04
[Feature]: IMDb Reviews


thenbhdIum:
### Area (tag)

Other

### Problem statement

No reviews to determine if interested in watching.

### Proposed solution

IMDb reviews. IMDb has been around the longest and has more reviews from users than the other services.

EzzatBoukhary:
Yup agreed. Imdb reviews would probably be the best but if not, I don't mind other things like Trakt Comments (or any other platform reviews) which are available on the mobile app.

PR #627, opened 2026-03-05
feat(detail): add TMDB reviews tab with auto-scroll and pause/resume …

JFeverish:
## Summary

Added a new Reviews tab in the detail screen, next to More like this.
Implemented TMDB review fetching (movie/{id}/reviews, tv/{id}/reviews) and mapped results into a new domain model (MetaReview).
Wired reviews into detail state/viewmodel with loading and error handling.
Added a dedicated ReviewsSection UI for TV:
Horizontal review cards
Auto-scroll for long review text
OK button toggles pause/resume of scrolling
Added new strings for reviews tab and states.

## Why

Issue #602 asked for IMDb-style reviews in the detail page.
This implementation uses TMDB reviews instead of IMDb reviews because:

IMDb does not provide a straightforward public/free reviews API for this use case.
TMDB is already integrated in this project (same metadata pipeline), so it is more reliable and maintainable.
TMDB provides official review endpoints with localization support, avoiding scraping-based or brittle solutions.

## Testing

Manually validated:
Reviews tab appears in detail page.
Reviews load and render correctly.
Long text auto-scrolls after delay.
Pressing OK pauses scroll; pressing OK again resumes.
No regressions observed in existing detail tabs.

https://github.com/user-attachments/assets/28097624-7cb0-4d88-9bd5-f3c097430daa

## Breaking changes

None.

## Linked issues

#602 - Address IMDb reviews request via TMDB reviews integration (chosen due to IMDb API availability/reliability constraints).

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/tmdb/TmdbMetadataService.kt
+++ b/app/src/main/java/com/nuvio/tv/core/tmdb/TmdbMetadataService.kt
@@ -7,10 +7,12 @@ import com.nuvio.tv.data.remote.api.TmdbImage
 import com.nuvio.tv.data.remote.api.TmdbPersonCreditCast
 import com.nuvio.tv.data.remote.api.TmdbPersonCreditCrew
 import com.nuvio.tv.data.remote.api.TmdbRecommendationResult
+import com.nuvio.tv.data.remote.api.TmdbReviewResult
 import com.nuvio.tv.domain.model.ContentType
 import com.nuvio.tv.domain.model.MetaCastMember
 import com.nuvio.tv.domain.model.MetaCompany
 import com.nuvio.tv.domain.model.MetaPreview
+import com.nuvio.tv.domain.model.MetaReview
 import com.nuvio.tv.domain.model.PersonDetail
 import com.nuvio.tv.domain.model.PosterShape
 import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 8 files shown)


tapframe:
This is a great addition. but is it modular under tmdb enrichment? or just available regardless of the toggle

JFeverish:
> This is a great addition. but is it modular under tmdb enrichment? or just available regardless of the toggle

It is modular under TMDB enrichment.

In the current implementation, reviews are loaded only when TMDB enrichment is enabled (shouldLoadReviews() checks settings.enabled). If enrichment is off, we return early with empty reviews state and no TMDB reviews request is made.

image

Would it make sense to have TMDB enrichment enabled by default?

EzzatBoukhary:
Awesome feature that is rarely made in TV apps but definitely needed!

In my opinion, it would be better if pressing OK opened the full review in a dialog or something instead of having to wait for a slow scroll. 

As for the settings, does it have a separate toggle for Reviews only so that TMDB enrichment can be on but reviews specifically off? 

Lastly, as a UI tweak, I'd suggest putting the rating at the right end of the card and making it pop out a bit more (maybe through color-coding or an icon?) It currently blends with the name and other texts somewhat.

mistersandman-64:
> As for the settings, does it have a separate toggle for Reviews only so that TMDB enrichment can be on but reviews specifically off?
> 

Yes, please include a switch to turn off reviews entirely if unwanted, similarly to the More like this option.

PR #632, opened 2026-03-06
Perrformance Improvement - Media Details and Streams Screen

### LINKED ISSUE #573: [Bug]: Performance when scrolling or browsing horizontal
### Pre-flight checks

### App version / commit

0.4.9

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Fire TV 4K Max 2nd Gen

### Android version

Android 11

### Player mode

Internal player

### Area (tag)

UI / Layout

### Steps to reproduce

Performance when browsing the app is choppy or dropping frames while scrolling.  It’s worse when you’re viewing a show with multiple seasons and try to browse horizontally through the episode or going up or down. 

### Expected behavior

Smoother scrolling.  

### Actual behavior

Stuttering.  Drop frames. Seems cpu heavy. 

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

### Logs (optional but helpful)

```shell

```

### Anything else? (optional)

Video:
Stremio first as an example then Nuvio

 https://github.com/user-attachments/assets/0619dede-a2aa-442d-bf52-ba4a1df1b331

skoruppa:
## Summary
Performance optimization pass across the Media Details and Streams Screen (only gradient in the second one), focused on reducing per-frame GPU and UI-thread work.

Key changes:
- **Gradients replaced with pre-rendered bitmaps** - the most impactful change. Backdrop gradients in both the detail and stream screens are now computed once and reused as static textures instead of being recalculated every frame. Dithering applied to reduce visible banding.
- **Episode card rendering simplified** - reduced unnecessary draw passes and object allocations per card per frame; fixed jagged rounded corners
- **Scroll behavior improved** - horizontal focus movement inside the episode list no longer triggers unnecessary vertical re-layout of the whole screen

## Why
The detail screen was causing visible janky animation when Episode Section was visible on my 85MQLED85. The main bottlenecks were per-frame gradient computation and unnecessary work triggered by focus movement in the episode row.

## Testing
Tested manually on device - before/after comparison attached

**Before**:
IMG20260306001259

**After**:
IMG20260306020824

Episode card is slightly simplified: no 1px border, no scale-up on focus. The bitmap gradient is visually less smooth than the previous version, but the performance difference justifies the trade-off.
IMG20260306024530

## Breaking changes
None

## Linked issues
Fixes #573

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/EpisodesSection.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/EpisodesSection.kt
@@ -104,7 +104,7 @@ fun SeasonTabs(
             shape = RoundedCornerShape(20.dp)
         )
     )
-    val tabScale = CardDefaults.scale(focusedScale = 1.02f)
+    val tabScale = CardDefaults.scale(focusedScale = 1.0f)
     val typography = MaterialTheme.typography
     val tabTextStyle = remember(typography) { typography.titleMedium }
     val textSecondary = NuvioTheme.extendedColors.textSecondary
@@ -484,7 +484,7 @@ private fun EpisodeCard(
             shape = shape
         )
     )
-    val cardScale = CardDefaults.scale(focusedScale = 1.02f)
+    val cardScale = CardDefaults.scale(focusedScale = 1.0f)
     val cardGlow = CardDefaults.glow()
 
     Card(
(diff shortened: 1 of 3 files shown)


halibiram:
Could you check the cast details screen? I'm getting frame drops when the poster changes.

ram130:
> Could you check the cast details screen? I'm getting frame drops when the poster changes.

This will improve or fix it.

PR #633, opened 2026-03-06
feat: refactor movie watched status tracking to use a global set of w…

### LINKED ISSUE #610: [Bug]: Movie watched icon (Check Mark)
### Pre-flight checks

### App version / commit

0.4.10-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Nvidia Shield Pro 2019

### Android version

Android 11 or 9.2.4

### Player mode

Internal player

### Area (tag)

Watch Progress (resume, watched state, history)

### Steps to reproduce

Open home page. 

Navigate to any movie lists.

Previous movies watched are no showing the check mark on poster. 

Click on movie details and it’s showing the “eye” icon with no slash through it. Meaning it’s been watched. 

### Expected behavior

I would expect even with Cinemeta removed the watch icon to show up working both in details and poster. 

### Actual behavior

Movie watched check mark is not showing up after completion of movie. 

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

### Logs (optional but helpful)

```shell

```

### Anything else? (optional)

Currently using AlOMetadata, no need for Cinemeta.
Once I remove Cinemeta, the marked as watched icon no longer show up for any movies on my home catalog page.
Even if I keep the Cinemeta addon and disable its catalog list or move them below my AlOmetadata the marked watched icon disappears.

This is why it would be great to have Cinebye support for Nuvio or vice versa. 

If Nuvio developers could implement a way that users could remove Cinemeta and still have the check mark present on home catalog page of any movies watched within their lists.

Maybe this is not an actual bug but more of a request. 

Thank you..

tapframe:
## summary

this changes the home watched badge logic to use a shared set of watched movie ids instead of creating individual watched observers per poster.

that means home can show watched badges for all loaded movie posters again, while still keeping the performance protections that were added earlier to avoid startup lag.

it also fixes watched badges in see all, which was not reading watched state before.

## why

this is needed to fix the bug where watched movies were showing correctly in details, but not on posters in home for a lot of users, especially when using addons like aiometadata without relying on cinemeta ordering.

the old home logic only observed watched state for a limited number of posters to avoid lag. that helped performance, but it also meant a lot of posters never got watched updates at all.

this keeps the fast path, but moves watched state to a single shared source instead of lots of peritem observers.

## testing

manual:
- checked watched badges on classic, grid, and modern home layouts
- checked watched badges in see all
- checked that detail watched state still lines up with poster watched state

## screenshots / video (ui changes only)

none

## breaking changes

none

## linked issues

fixes #610

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/WatchedItemsPreferences.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/WatchedItemsPreferences.kt
@@ -6,6 +6,7 @@ import androidx.datastore.preferences.core.stringSetPreferencesKey
 import com.nuvio.tv.core.profile.ProfileManager
 import com.google.gson.Gson
 import com.nuvio.tv.domain.model.WatchedItem
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flatMapLatest
@@ -14,6 +15,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
+@OptIn(ExperimentalCoroutinesApi::class)
 class WatchedItemsPreferences @Inject constructor(
     private val factory: ProfileDataStoreFactory,
     private val profileManager: ProfileManager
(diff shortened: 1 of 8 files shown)


CruelIn10tions:
I see that you closed my issue and referred it to something similar. I read over the info and understand some of it. 

My only question is with my issue be update in future versions of the app? Can I look forward to my issue no longer being an issue. Or is it a setting I need to change now. 

Thank you..

tapframe:
> I see that you closed my issue and referred it to something similar. I read over the info and understand some of it. 
> 
> My only question is with my issue be update in future versions of the app? Can I look forward to my issue no longer being an issue. Or is it a setting I need to change now. 
> 
> Thank you..

This will only be updated in the upcoming build. We can reopen the issue if you still face the problem after the new release.

PR #635, opened 2026-03-06
Remove redundant backdrop image flash in trailer expand; fade poster to black before expand

mistertomlinson:
## Summary

Replaces the brief backdrop image flash in the expanded trailer card with a clean black overlay for both the Modern UI and Classic UI. Adds pre-expansion fade-to-black animation and gates card expansion on trailer URL availability.

## Why

When using the Modern or Classic UI layouts with auto-trailers enabled on expanded cards, the expanded card displays a backdrop image as a placeholder while the trailer loads. This backdrop is the same cinematic image already displayed prominently in the hero section at the top of the home screen. The same image ends up appearing twice on the screen simultaneously — once large in the hero section, and again in the expanding card directly below it, only for a second. This seems redundant and unnecessary.

## Testing

Tested manually on a physical Android TV device with the following scenarios:
- Portrait and landscape poster modes with trailers enabled
- Trailer delay settings of 1, 3, and 7 seconds
- Movies with and without available trailers
- Moving focus away before, during, and after the fade sequence
- Letting the trailer play to completion
- Disabling auto trailers entirely (original behavior preserved)
- Classic UI and Modern UI both verified

Manual testing performed on device — before/after behavior verified

## Breaking changes

None

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeContent.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeContent.kt
@@ -103,6 +103,8 @@ import android.view.KeyEvent as AndroidKeyEvent
 import kotlinx.coroutines.flow.distinctUntilChanged
 
 private const val KEY_REPEAT_THROTTLE_MS = 80L
+private const val TRAILER_PRE_FADE_DURATION_MS = 400L
+private const val TRAILER_URL_POLL_INTERVAL_MS = 200L
 
 @Composable
 fun ModernHomeContent(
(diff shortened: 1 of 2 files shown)


mistertomlinson:
I'm realizing now the background image was necessary for erroneous trailers where the card expands but no trailer plays. That way it doesn't load a blank box. I can't think of an elegant way to fix it. I implemented a 1 second timeout so if a trailer exists but doesn't play, the box collapses, but most people probably wouldn't want this behavior. A second image that's different from the hero section image would probably be a good solution, or maybe someone can figure out what to do about trailers failing.

PR #637, opened 2026-03-06
apply dither to gradient bitmap - feels a bit smoother

skoruppa:
## Summary
Smoother a bit the prerendered gradient introduced in #632 using dither

## Why
The new gradient has very visible "banding"

## Testing
Tested manually on device

The gradient has less visible "banding" but is still pretty visible when there is no backdrop image

## Breaking changes
No breaking changes introduced.

## Linked issues
Fixes #632


JFeverish:
Where is the dithering being applied?

There are some people that are sensitive to dithering. Myself included. But usually more on computer, less on TV.

skoruppa:
> Where is the dithering being applied?
> 
> There are some people that are sensitive to dithering. Myself included. But usually more on computer, less on TV.

On the "dark to transparent" gradients that are visible in episode section and streams section. I think it is not even visible
IMG20260306024530
But makes the bending there way more noticable

PR #658, opened 2026-03-07
hotfix: unify the corner radius of the poster across all poster cards…

saimuelbr:
## Summary

Introduced dynamic corner radius for poster cards across the application. The shape now reacts to the user's preference set in the Home settings (e.g., Rounded vs. Squared), ensuring UI consistency between the home screen, and episode lists.

## PR type

## Why

Previously, poster formats were coded into multiple components, which caused a visual discrepancy when the user changed the overall configuration. This change links these components to the central component.

## Policy check

## Testing

**Manual Test**: Verified and tested with a compiled version of the application.

**Scenarios**: 
**1**. Changed the poster format to "Rounded".

**2**. Changed the poster format to "Square" -> Verified that all cards were updated instantly.

**3**. Navigated through the continue watching, seasons, episodes, library, discovery section and search results to ensure there were no layout regressions.

**Ex with Square**:

image
image
image
image
image

## Breaking changes

**None**.

## Linked issues

Issue noticed with personal use of the app.


tapframe (commented):
I'm having mixed feeling about this, especially having same corner radius in episode cards and cw section cards while corner radius is set to 0

saimuelbr:
> I'm having mixed feeling about this, especially having same corner radius in episode cards and cw section cards while corner radius is set to 0

The intention is that all poster cards in the application will follow the user's preference regarding the poster card shape, whether it's episodes, catalogs, or CW. This preference will be followed by all poster cards. The user will have total control over the desired card shape, whether square, round, pill, classic, or subtle. Regardless of the choice, all cards will follow the preference. My images above are just demonstrations using the square style.

skoruppa:
@tapframe I know it's been some time since this was proposed, but I think I agree. If user WANTS to have specific posters shape, we should apply it on all posters - collections, cw, discovery, search - not only home catalogs

But I also agree that episodes should not be taken into account here. Do you agree?

skoruppa:
Implemented in #3190

Issue #538, opened 2026-03-03
[Enhancement] 0.5 min Threshold Minutes option


mistersandman-64:
A lower 0.5 min Threshold Minutes option for Next Episode Threshold Mode

burggraf:
Please add a 0.0 min Threshold Minutes option as well -- I would like to never see this popup.

PR #675, opened 2026-03-07
feat: add 0 and 0.5 minute options for Next Episode Threshold Mode

burggraf:
## Summary

This PR adds new options for "Next Episode Threshold Mode", allowing users more granular control over when the next episode popup appears.

## Changes

- **Extended minimum threshold values**: Previously the lowest setting was 1 minute, now users can select:
  - **0 minutes** - When selected, the Next Episode popup is never shown
  - **0.5 minutes (30 seconds)** - For finer control near the end of episodes

- **UI Updates**: 
  - Slider range extended to support 0-99.5% (percentage mode) and 0-3.5 minutes (time mode)
  - Displays "Never" instead of "0%" or "0 min" when threshold is set to 0

- **Logic Updates**:
  - Added early return `false` in `shouldShowNextEpisodeCard()` when threshold is <= 0
  - Updated `coerceIn` ranges to support 0f minimum values

## Files Modified

- `PlayerSettingsDataStore.kt` - Updated min/max ranges and normalization
- `PlaybackAutoPlaySettings.kt` - Updated slider UI with conditional display
- `PlayerNextEpisodeRules.kt` - Added 0 threshold handling logic
- `strings.xml` - Added `action_never` string resource

## Testing

- [ ] Verify slider shows 0, 0.5, 1, 1.5, etc. options in minutes mode
- [ ] Verify slider shows 0%, 0.5%, 1%, etc. options in percentage mode  
- [ ] Verify "Never" text appears when 0 is selected
- [ ] Verify Next Episode popup does not appear when threshold is set to 0
- [ ] Verify existing threshold values (1-3.5 min, 97-99.5%) still work correctly


skoruppa:
Should fix #538

@burggraf but if the button is never displayed, does the auto play still work?

burggraf:
No, a new episode should never autoplay (if that's turned off).  If autoplay is turned on, then I guess yes, it would play.  My use case is autoplay off, no next episode button.  (We generally do not binge watch.)

nsglcck:
I guess it depends on the series, but in my experience the current highest value (3.5 minutes) isn't even enough. Can this be increased up to 5 minutes?

PR #676, opened 2026-03-07
fix(player): restore libass ASS/SSA subtitle pipeline

kernexshadow:
## Summary

This PR restores the experimental libass subtitle path for ASS/SSA tracks.

Changes:
- Added libass parser compatibility for Media3 subtitle tracks exposed as `application/x-media3-cues` with `codecs=text/x-ssa`.
- Wired subtitle parser override only when `libass` is enabled, and reset to default when `libass` is disabled.
- Kept subtitle settings/toggle and related strings consistent with the restored behavior.

## PR type

- Bug fix

## Why

ASS/SSA subtitles were being parsed through the default Media3 cues pipeline instead of the libass path in some cases, which caused embedded ASS styling (for example colors/effects) to be lost.  
This change ensures ASS/SSA tracks are routed to libass when enabled, while keeping the non-libass path unchanged.

## Policy check

- [] If this is a larger or directional change, I linked the issue where it was approved. (N/A: not a larger/directional change)

## Testing

- Manual playback test with internal ASS/SSA tracks:
  - libass enabled: ASS tracks are created/selected through libass-compatible path and styling renders correctly.
  - libass disabled: player falls back to normal subtitle behavior.

[None.](https://github.com/user-attachments/assets/d167d6e4-a069-4136-acc3-95dbc2c6ef69)

## Breaking changes

None.

## Linked issues

N/A.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
@@ -226,6 +226,11 @@ class PlayerRuntimeController(
     internal var hasRequestedScrobbleStartForCurrentItem: Boolean = false
     internal var scrobbleStartRequestGeneration: Long = 0L
     internal var hasSentCompletionScrobbleForCurrentItem: Boolean = false
+    internal var requestedUseLibassByUser: Boolean = false
+    internal var libassPipelineOverrideForCurrentStream: Boolean? = null
+    internal var activePlayerUsesLibass: Boolean = false
+    internal var libassPipelineSwitchInFlight: Boolean = false
+    internal var libassPipelineDecisionStreamUrl: String? = null
     internal var episodeStreamsJob: Job? = null
     internal var episodeStreamsCacheRequestKey: String? = null
     internal val streamCacheKey: String? by lazy {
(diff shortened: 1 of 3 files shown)


tapframe:
Thanks, would be great if you could provide a video of testing. Doesn’t have an apt hardware to test it myself.

kernexshadow:
> Thanks, would be great if you could provide a video of testing. Doesn’t have an apt hardware to test it myself.


https://github.com/user-attachments/assets/d167d6e4-a069-4136-acc3-95dbc2c6ef69

kernexshadow:
@tapframe It's not good enough to be merged yet.
for some reason when there is not ass subtitles, watching a movie the colors of the movie will be washed out. Not sure why.

kernexshadow:
@tapframe  The washed-out colors were caused by libass being applied to the entire playback pipeline whenever it was enabled, even on videos without ASS/SSA subtitles; in practice this could force a GL-based rendering path that changes color/tone behavior on some devices. I fixed it by making libass activation conditional per stream: the player now keeps libass only when ASS/SSA subtitle tracks are actually present, and automatically falls back to the normal pipeline otherwise, while preserving playback position.

It's good to merge now.

kernexshadow:
If anyone wants to try:

https://we.tl/t-UMMwrezJST

PR #699, opened 2026-03-08
feat: DNS over HTTPS support with in-memory caching

halibiram:
## Summary

Adds DNS over HTTPS (DoH) support to the playback settings, giving users a way to use secure providers like Cloudflare, Google, or AdGuard. To make sure metadata and posters still load fast, I've also added a simple in-memory DNS cache so we're not constantly hitting the DoH servers for every single image/request.

## PR type

## Why

Sometimes local DNS causes ISP blocks or just slows down metadata for TMDB/Stremio add-ons. Adding DoH helps fix that by increasing privacy and bypassing simple DNS restrictions, so metadata/posters load more reliably.

## Policy check

## Testing

- Manual test: Switched DNS to AdGuard in Settings -> Playback -> Network.
- Verified via logcat (`NuvioDNS` tag) that it's hitting `dns.adguard.com` over HTTPS instead of standard port 53.
- Confirmed that images and metadata load quickly with the new `CachedDns` wrapper; it avoids the DoH handshake overhead for repeated requests.
- Verified build/install with `assembleDebug` and `installDebug`.

N/A

## Breaking changes

None. The default option is `0` (System Default), so the app's behavior stays exactly the same unless a user manually changes it.

## Linked issues

Improves metadata reliability. No specific issue was linked for this.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -206,6 +206,7 @@ dependencies {
     implementation(libs.retrofit.moshi)
     implementation(libs.okhttp)
     implementation(libs.okhttp.logging)
+    implementation(libs.okhttp.dnsoverhttps)
     implementation(libs.moshi)
     ksp(libs.moshi.codegen)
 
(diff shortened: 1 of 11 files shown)


skoruppa:
@tapframe I don't know if you had a chance to look at this. From what I saw, there are user who would benefit from this change. But I'm not sure if you want it. 

@halibiram I think this setting can be moved into "Advanced" tab where we already have speed test option (and nothing else ><) as it changes whole network stack, not only for Player. Does this cache

PR #712, opened 2026-03-09
back loop on episode end, auto stream selection priority, anime skip intro season mapping

skoruppa:
## Summary

Fixed remaining back navigation loop paths: when an episode ends naturally and auto source selection navigates to the next episode's stream screen, pressing back could loop back through stale stream/player entries.

Also improved auto stream selection reliability by waiting for all addons to respond before picking a stream, with a configurable timeout (default 3s, range 0–10s + Unlimited to wait for results of all addons).

Additionally:
- Better season mapping for MAL/Kitsu IDs in skip intro sources
- Added translations for "continuing", "ended" and "current" series status badges
- Added Polish translations

## PR type

- Bug fix

## Why

- When an episode ended naturally with auto source selection enabled, `onPlaybackEnded` was using `popUpTo(Player)` which only removed the player from the back stack, leaving the previous stream screen behind. Navigating to the next episode's stream screen then stacked on top of the old one, causing a back loop. Fixed by using `popUpTo(Detail)` to clear both player and stream before navigating.
- Auto stream selection was triggering on the first addon that responded rather than waiting for all of them. This meant that if your primary addon was slightly slower than a backup one, the backup would get selected instead. It was also easily triggered by addons that fail fast (e.g. rate limit or not found errors are near-instant), which would cause the selector to run on an empty or incomplete list and fall through to a lower-priority source. Now waits up to a configurable timeout (default 3s) for all addons to respond before selecting — if nothing arrives within the timeout, it waits for the first successful result before selecting, and falls back to the stream selection screen only if all addons fail. The timeout is user-configurable (0–10s) in playback settings; 0 means select on first result (original behavior).
- Skip intro for anime-skip was returning wrong results or no results for anime season 2+ when using MAL or Kitsu IDs, because the season number was not being correctly derived from the ID mapping.
- Series status badge strings ("ended", "continuing", "current") were not localized — they are now mapped to translatable string resources.

## Policy check

## Testing

- Reproduced and verified the reported back navigation loop is resolved
- Manually verified ARM season index mapping for MAL IDs across multiple multi-season anime
- Confirmed AnimeSkip fallback correctly filters by season number when using season-1 AniList ID
- Verified auto stream selection now respects addon priority order
- Verified timeout setting correctly controls selection timing

IMG20260309201110
IMG20260309201100

## Breaking changes

Should not break anything

## Linked issues

Issues reported directly on Discord

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
@@ -439,7 +439,7 @@ class PlayerSettingsDataStore @Inject constructor(
                 streamAutoPlayNextEpisodeEnabled = prefs[streamAutoPlayNextEpisodeEnabledKey] ?: false,
                 streamAutoPlayPreferBingeGroupForNextEpisode =
                     prefs[streamAutoPlayPreferBingeGroupForNextEpisodeKey] ?: true,
-                streamAutoPlayTimeoutSeconds = (prefs[streamAutoPlayTimeoutSecondsKey] ?: 3).coerceIn(0, 10),
+                streamAutoPlayTimeoutSeconds = (prefs[streamAutoPlayTimeoutSecondsKey] ?: 3).coerceIn(0, 11),
                 nextEpisodeThresholdMode = prefs[nextEpisodeThresholdModeKey]?.let {
                     runCatching { NextEpisodeThresholdMode.valueOf(it) }.getOrDefault(NextEpisodeThresholdMode.PERCENTAGE)
                 } ?: NextEpisodeThresholdMode.PERCENTAGE,
@@ -646,7 +646,7 @@ class PlayerSettingsDataStore @Inject constructor(
 
     suspend fun setStreamAutoPlayTimeoutSeconds(seconds: Int) {
         store().edit { prefs ->
-            prefs[streamAutoPlayTimeoutSecondsKey] = seconds.coerceIn(0, 10)
+            prefs[streamAutoPlayTimeoutSecondsKey] = seconds.coerceIn(0, 11)
         }
     }
 
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerStreams.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerStreams.kt
@@ -823,7 +823,7 @@ internal fun PlayerRuntimeController.playNextEpisode() {
             }
 
             val timeoutMs = playerSettings.streamAutoPlayTimeoutSeconds * 1_000L
-            if (timeoutMs > 0L) {
+            if (timeoutMs > 0L && playerSettings.streamAutoPlayTimeoutSeconds < 11) {
                 delay(timeoutMs)
             }
             timeoutElapsed = true
--- a/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackAutoPlaySettings.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/settings/PlaybackAutoPlaySettings.kt
@@ -126,10 +126,10 @@ internal fun LazyListScope.autoPlaySettingsItems(
 
     item(key = "autoplay_stream_timeout") {
         val timeoutSec = playerSettings.streamAutoPlayTimeoutSeconds
-        val valueText = if (timeoutSec == 0) {
-            stringResource(R.string.autoplay_timeout_instant)
-        } else {
-            "${timeoutSec}s"
+        val valueText = when (timeoutSec) {
+            0 -> stringResource(R.string.autoplay_timeout_instant)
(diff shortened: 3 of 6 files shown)


Trulecenttt:
I personally wouldn't like the 5sec wait implementation. I purposely have my addon set to timeout at 1.5 seconds so my streams can autoplay ASAP, then I have another addon with a more generous timeout so if for some reason my first addon shows no results, or weak results, I just change source In player. But I never have to do that tbh

skoruppa:
> I personally wouldn't like the 5sec wait implementation. I purposely have my addon set to timeout at 1.5 seconds so my streams can autoplay ASAP, then I have another addon with a more generous timeout so if for some reason my first addon shows no results, or weak results, I just change source In player. But I never have to do that tbh

I think the lowest we can go is 3s :d a balance between going fast, and let addons for regular users to load a bit

@tapframe thoughts?

skoruppa:
Or in a spirit of Nuvio configurability, I will add a setting for the timeout. It would be easy, and let user to set this timeout from 0 to s, with a default of 3s. Probably under "auto source select" dropdown if other option than manually is selected. 

@tapframe, do you agree?

Trulecenttt:
Sounds like a great middle ground for everyone!

tapframe:
> Or in a spirit of Nuvio configurability, I will add a setting for the timeout. It would be easy, and let user to set this timeout from 0 to s, with a default of 3s. Probably under "auto source select" dropdown if other option than manually is selected. 
> 
> 
> 
> @tapframe, do you agree? 

This will do it. Just ensure you properly test for any edge cases before pushing. Thanks again.

skoruppa:
> > Or in a spirit of Nuvio configurability, I will add a setting for the timeout. It would be easy, and let user to set this timeout from 0 to s, with a default of 3s. Probably under "auto source select" dropdown if other option than manually is selected.
> > @tapframe, do you agree?
> 
> This will do it. Just ensure you properly test for any edge cases before pushing. Thanks again.

Done :)
There should be no edge cases. The logic is just like it was before, the only difference is that it now wait n seconds before doing a regex or picking the first stream from the list. If users set the timeout to 0, it will behave exactly as it was - after first addon return success, stream will be selected.

mistersandman-64:
> > > Or in a spirit of Nuvio configurability, I will add a setting for the timeout. It would be easy, and let user to set this timeout from 0 to s, with a default of 3s. Probably under "auto source select" dropdown if other option than manually is selected.
> > > @tapframe, do you agree?
> > 
> > 
> > This will do it. Just ensure you properly test for any edge cases before pushing. Thanks again.
> 
> Done :) There should be no edge cases. The logic is just like it was before, the only difference is that it now wait n seconds before doing a regex or picking the first stream from the list. If users set the timeout to 0, it will behave exactly as it was - after first addon return success, stream will be selected.

Will this also be implemented for manual selection with auto-play next episode/prefer binge group enabled?

skoruppa:
> > > > Or in a spirit of Nuvio configurability, I will add a setting for the timeout. It would be easy, and let user to set this timeout from 0 to s, with a default of 3s. Probably under "auto source select" dropdown if other option than manually is selected.
> > > > @tapframe, do you agree?
> > > 
> > > 
> > > This will do it. Just ensure you properly test for any edge cases before pushing. Thanks again.
> > 
> > 
> > Done :) There should be no edge cases. The logic is just like it was before, the only difference is that it now wait n seconds before doing a regex or picking the first stream from the list. If users set the timeout to 0, it will behave exactly as it was - after first addon return success, stream will be selected.
> 
> Will this also be implemented for manual selection with auto-play next episode/prefer binge group enabled?

Yes it will, as all paths uses the same code, so by default it will now wait 3s for addons. But I think that maybe the setting should be always visible. I unnecessary hidden it for the "manual" option

mistersandman-64:
> > > > > Or in a spirit of Nuvio configurability, I will add a setting for the timeout. It would be easy, and let user to set this timeout from 0 to s, with a default of 3s. Probably under "auto source select" dropdown if other option than manually is selected.
> > > > > @tapframe, do you agree?
> > > > 
> > > > 
> > > > This will do it. Just ensure you properly test for any edge cases before pushing. Thanks again.
> > > 
> > > 
> > > Done :) There should be no edge cases. The logic is just like it was before, the only difference is that it now wait n seconds before doing a regex or picking the first stream from the list. If users set the timeout to 0, it will behave exactly as it was - after first addon return success, stream will be selected.
> > 
> > 
> > Will this also be implemented for manual selection with auto-play next episode/prefer binge group enabled?
> 
> Yes it will, as all paths uses the same code, so by default it will now wait 3s for addons. But I think that maybe the setting should be always visible. I unnecessary hidden it for the "manual" option

Awesome, thanks! If you guys think it's a good option, would it be possible to also display the Auto-play Source Scope/Allowed Addons options for Manual if auto-play is enabled?

skoruppa:
This is for sure out of scope of this small and easy bugfix ><

mistersandman-64:
> This is for sure out of scope of this small and easy bugfix ><

Understood. Thanks for working on those other bug fixes. Greatly appreciated!

PR #770, opened 2026-03-11
feat: Caching the stream links

swapByte:
## Summary
Adds persistent caching for scraper results to avoid re-running scrapers when the same content is opened again. Results are stored in DataStore with configurable TTL and automatically reused if valid. Also introduces a `forceRefresh` option to bypass cache when needed.

## PR type

## Why
Scrapers were executed every time a user opened the same movie/episode, even if results had already been fetched moments earlier.
This caused unnecessary network requests, slower UI loading, and redundant scraping.

This change introduces a lightweight caching layer that:

- Stores scraper results per content key
- Reuses cached results when still valid
- Allows manual refresh to bypass cache
- Automatically expires entries based on user-defined duration

## Policy check

- [X] If this is a larger or directional change, I linked the issue where it was approved.

## Testing
I have manually tested it.
- Opened movie - scrapers executed and results cached
- Reopened same movie - cached results returned instantly
- Pressed back during loading - results still saved in cache
- Disabled cache - cache cleared correctly
- Force refresh - scrapers executed again ignoring cache

Added settings UI for:
- Enabling/disabling stream cache
- Selecting cache duration

## Breaking changes

No Breaking changes

## Linked issues

No Linked issues


skoruppa:
Hello. Can you resolve conflicts? 

Is the cache also used by the "Change Source" view available in the player? 
Have you tested if cache contains all results with auto stream picking, when timeout picks a stream before all addons return results?

I think there also should be an option to manually enforce obtaining results without cache from both Stream Screen and Change Source Player view when option is enabled

swapByte:
@skoruppa 

Cache is used for "Change Source" view as well.
in "Change Source" View we already have reload button on press of that I am invalidating cache and refetching streams.
Auto stream is also working fine with timeouts.
By toggling "Stream Cache" it will clear the cache.

swapByte:
debug build:
https://drive.google.com/file/d/1AiTeYQsAKi4oCw-jqsD-KXnoXcwDabU5/view?usp=sharing

swapByte:
@skoruppa 
Fixed merge conflicts.
its working with cloudStream extensions as well.

skoruppa:
@swapByte looks good! But same "**Refresh**" button as on "Change Source" view is required now for the "Streams Screen". User should be able to manually revalidate the responses by pressing a button, without the need to go and disable a setting

swapByte:
@skoruppa added Reload button. please review.

skoruppa:
StreamCacheDataStore caches results only from JS plugins/scrapers but not from Stremio addons. If we are going to implement a cache it should cache everything, not only plugins. 

When you enter the stream screen, addon streams are fetched via StreamRepository, which has no caching layer at all

I'm also not sure if I like the placement of the "Refresh" button. Maybe we can have additional tab, before the "All" with refresh icon, that would be shown only when Stream Cache is enabled. Default focus should still be on the "All: tab (after entering details screen)

sidsharths1245:
@swapByte I am still facing this issue. Is there some setting we need to update to fix the issue?

skoruppa:
Implemented in #3031

PR #780, opened 2026-03-11
feat(player): PCM amplification in audio 

kernexshadow:
## Summary

- Adds a new PCM amplification feature in the player audio overlay.
- Shows `Amplification (PCM)` controls next to audio track selection.
- Allows real-time gain control with `- / +` buttons (range: `0 dB` to `10 dB`).
- Adds `Persist between sessions` toggle to save/load the selected gain.
- Stores settings in `PlayerSettingsDataStore`.
- Applies gain through `GainAudioProcessor` only for PCM audio (`ENCODING_PCM_16BIT` and `ENCODING_PCM_FLOAT`).
- Passthrough/bitstream audio is not modified.

## PR type

## Why

Some streams play with low PCM volume.  
Also downmixing for people with stereo speakers, the sound is lowered. And stays too low. No need to switch to Kodi to watch some streams.
This adds direct in-player PCM gain control while keeping passthrough audio unchanged.

## Policy check

- [] If this is a larger or directional change, I linked the issue where it was approved. 

## Testing

- Manual:
  - Open player audio overlay.
  - Select audio track and confirm normal behavior.
  - Change amplification with `- / +` and confirm value updates.
  - Toggle `Persist between sessions`, reopen playback, confirm saved behavior.

image

image

## Breaking changes

None,

## Linked issues

None.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
@@ -165,6 +165,8 @@ data class PlayerSettings(
     val decoderPriority: Int = 1, // EXTENSION_RENDERER_MODE_ON (0=off, 1=on, 2=prefer)
     val tunnelingEnabled: Boolean = false,
     val skipSilence: Boolean = false,
+    val audioAmplificationDb: Int = 0,
+    val persistAudioAmplification: Boolean = false,
     val preferredAudioLanguage: String = AudioLanguageOption.DEVICE,
     val secondaryPreferredAudioLanguage: String? = null,
     val loadingOverlayEnabled: Boolean = true,
@@ -255,6 +257,8 @@ class PlayerSettingsDataStore @Inject constructor(
 ) {
     companion object {
         private const val FEATURE = "player_settings"
+        private const val AUDIO_AMPLIFICATION_DB_MIN = 0
+        private const val AUDIO_AMPLIFICATION_DB_MAX = 10
     }
 
     private fun store(profileId: Int = profileManager.activeProfileId.value) =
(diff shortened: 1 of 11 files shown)


tapframe:
Can you top-align the audio card so it is placed just below the header? currently there is too much empty space in my opinion

kernexshadow:
> Can you top-align the audio card so it is placed just below the header? currently there is too much empty space in my opinion

Like this? 
image
Is the header too high, or good enough?

tapframe:
> > Can you top-align the audio card so it is placed just below the header? currently there is too much empty space in my opinion
> 
> 
> 
> Like this? 
> 
> image
> 
> Is the header too high, or good enough?
> 
> 

The header placement should remain the same as before. Only the card needs to be placed below the header to fill the empty space. The left rail will automatically position itself at the top when more audio cards are shown.

kernexshadow:
> > > Can you top-align the audio card so it is placed just below the header? currently there is too much empty space in my opinion
> > 
> > 
> > Like this?
> > image
> > Is the header too high, or good enough?
> 
> The header placement should remain the same as before. Only the card needs to be placed below the header to fill the empty space. The left rail will automatically position itself at the top when more audio cards are shown.

Done. 
image

johnle92:
Hey @tapframe, I noticed you implemented the PCM amplification feature in #780 to help with low volume on stereo setups. I'm facing a related but slightly different issue, when playing AAC 5.1 audio on a 2-speaker setup, the dialogue (center channel) is very quiet while effects and surrounds are loud. The imbalance remains even with PCM amplification since it boosts all channels equally.
A potential fix would be adding an optional toggle for stereo downmix via setMaxAudioChannelCount(2) on the DefaultTrackSelector. This would tell ExoPlayer to properly fold all channels including center channel dialogue into front L/R before output. Kodi has a similar feature called "Downmix: Center Mix Level" which solves this exact problem.
Would you consider adding this as an optional toggle in the audio settings? Happy to submit a PR if you can point me to the right files.

halibiram:
> A potential fix would be adding an optional toggle for stereo downmix via setMaxAudioChannelCount(2) on the DefaultTrackSelector

@johnle92 
It's not quite that simple. Source audio comes in as 5.1. Even if you force it to 2 channels via the track selector, ExoPlayer will just ignore that constraint because an actual 2-channel track doesn't exist in stream

LeAiQuoc:
> > A potential fix would be adding an optional toggle for stereo downmix via setMaxAudioChannelCount(2) on the DefaultTrackSelector
> 
> @johnle92 It's not quite that simple. Source audio comes in as 5.1. Even if you force it to 2 channels via the track selector, ExoPlayer will just ignore that constraint because an actual 2-channel track doesn't exist in stream

> @halibiram Found that Kodi implements this via FFmpeg's libswresample with a centerMix parameter passed into the resampler. Since Nuvio already bundles FFmpeg decoders, the capability is already there. Would a custom AudioProcessor similar to GainAudioProcessor that uses swresample for downmixing be the right approach for this?

PR #811, opened 2026-03-13
Feature/player episode selector season progress

halibiram:
## Summary

This started as a cleanup of the episode selector in the player side panel, but it ended up making the whole thing much easier to use.

The old horizontal season tabs are replaced with a proper season rail, each season now shows quick progress info, and the episode list gives better context around what is currently playing, what has already been watched, and what is unavailable. I also added a small summary block at the top so you do not lose track of where you are while browsing.

## PR type

## Why

On shows with a lot of seasons, the previous layout felt a little cramped and made it too easy to lose context while moving around. You could still get where you needed to go, but it took more effort than it should have.

This keeps the same overall flow, but makes the panel easier to scan, easier to navigate with a remote, and better at showing progress without forcing extra clicks.

## Policy check

- [x] This PR is not cosmetic-only.
- [x] This PR does not add a new major feature.
- [x] This PR is small in scope and focused on one existing part of the app.
- [x] This is not a larger product-direction change that required prior approval.

## Testing

Tested manually on Android TV.

- Opened shows with multiple seasons and verified the season rail selection updates correctly.
- Checked focus movement between the season rail and the episode list with remote navigation.
- Verified watched counts, progress indicators, and the "Now Playing" state against the active episode.
- Tried episodes across different seasons and confirmed the normal source loading flow still works.
- Checked specials and the new strings to make sure labels render correctly.

image

## Breaking changes

There are no breaking changes here. This stays within the existing player episode side panel and does not change playback APIs, data shape, or anything outside of this UI flow.

## Linked issues

There is no linked issue for this one. I kept it intentionally small and limited to improving the existing player episode selector.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/EpisodesSidePanel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/EpisodesSidePanel.kt
@@ -295,6 +295,24 @@ private fun EpisodesListView(
             .groupBy(keySelector = { it.first }, valueTransform = { it.second })
             .mapValues { (_, episodes) -> episodes.size }
     }
+    val seasonWatchedCounts = remember(
+        uiState.episodesAll,
+        uiState.episodeWatchProgressMap,
+        uiState.watchedEpisodeKeys
+    ) {
+        uiState.episodesAll
+            .groupBy { it.season }
+            .mapNotNull { (season, episodes) ->
+                season?.let { seasonNumber ->
+                    seasonNumber to episodes.count { episode ->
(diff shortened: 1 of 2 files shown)


timmyp53:
Great work.

EzzatBoukhary:
It would be nice if you can include the episode ratings in there too. 

Also the "Season 1" "15 episodes" "2 watched" summary chips might be redundant because the same info is shown in the seasons section. Removing it would give more space for the episode list and make it look cleaner imo. There isn't enough height being given to the episode list I feel like.

halibiram:
closing this as it's unnecessary now.

PR #894, opened 2026-03-17
Remember audio & subtitle track selection persistently

skoruppa:
## Summary

Audio and subtitle track selections are now persisted per content item and restored automatically when playback starts. Previously the selection by `sameSeriesTrackSelectionPreference` was only kept in memory for the current session.

## PR type

- Bug fix
- New feature

## Why

When returning to a movie or episode via Continue Watching, the previously selected audio track and subtitles were reset to defaults. If you went from one episode to another, but in between had Stream Selection Screem, it would also reset to defaults. Users had to manually reselect their preferred language/track every time

## Policy check

- [x] This PR is not cosmetic-only, unless it is a translation PR.
- [x] This PR does not add a new major feature without prior approval.
- [x] This PR is small in scope and focused on one problem.
- [ ] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

- Opened a movie, changed audio track and subtitles, exited player, returned via Continue Watching -> correct tracks restored
- Opened a series, changed tracks mid-episode, switched to next episode -> correct tracks restored
- Changed tracks on episode 3 (auto-restore failed), manually reselected on episode 3, moved to episode 4 -> episode 4 picked up the corrected preference
- Fresh install / no saved preference -> default track selection behavior unchanged

## Screenshots / Video (UI changes only)

Nothing changed UI wise

## Breaking changes

Nothing should break 

## Linked issues

Fixes #870
Fixes #801 
Fixes #660

Issue #870, opened 2026-03-16
[Bug]: Remembered subtitle selection broken in latest update


y-a13k:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

0.4.19-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Amazon Fire TV Stick 4K Max (2025)

### Android version

Android 11

### Player mode

Internal player

### Area (tag)

Subtitles (download, styling, sync)

### Steps to reproduce

1. Open a series / show stream with more than one embedded subtitles e.g. English (no CC), English (forced) and English CC
2. Click the next episode button and choose the exact same stream.

### Expected behavior

The player should remember the selected embedded subtitles chosen in the previous episode (with the same stream of course).

### Actual behavior

In update 1.4.18, this issue was fixed and it remembered the subtitle selection in the previous episode, however now it seems to be broken again and now defaults back to the first embedded subtitle even if another was selected in the previous stream.

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

- Content type: series only

### Logs (required for crash reports)

```shell

```

ram130:
@skoruppa this been a good request for a lot of users. 

- To add to this. Subtitles setting should be per show/movie if set by the user during playback, and only apply their default setting in settings if there is no data for that content

- Should also sync the settings with Nuvio sync to other devices.

skoruppa:
@ram130 what is saved in Nuvio's sync depends completely on @tapframe ;) and I don't know if he would like to save such things

As for the rest, will look when I get a chance

ram130:
> @ram130 what is saved in Nuvio's sync depends completely on @tapframe ;) and I don't know if he would like to save such things
> 
> As for the rest, will look when I get a chance 

Copy.  Thanks for looking into it. I was binging a show that needed subtitles when I was testing for our infamous loop bug, that's when I noticed why remembering should totally be implemented lol.

y-a13k:
> @skoruppa this been a good request for a lot of users. 
> 
> - To add to this. Subtitles setting should be per show/movie if set by the user during playback, and only apply their default setting in settings if there is no data for that content
> 
> - Should also sync the settings with Nuvio sync to other devices. 

[...]

PR #915, opened 2026-03-18
Fix race condition & cleanup for persistent audio & subtitle track selection

skoruppa:
## Summary
Fixing race condition in persistent audio & subtitle track selection and small cleanup for #894 

**Current Logic**
When restoring a saved audio or subtitle preference, the player tries to find the best matching track using the following priority:

1. Track ID + language + name - all three must match. Used when resuming the exact same file/source.
2. Exact name + language - used when the source changed, but the track has the same name
3. Name contains + language - looser name match, still requires language for matching
4. Language only - last resort, used when moving to the next episode where track IDs and names differ, but the language stays the same (e.g. "ja").

For subtitles, the same matching logic applies for internal (embedded) tracks.
Addon subtitles are matched first by exact id + addon name, then by language.
If the user had subtitles disabled, that state is also persisted and restored - the player will keep subtitles off even if it would normally auto-select one based on the preferred language setting.

If no match is found at any step, the preference is not applied and the player falls back to its default track selection.

## PR type

- Bug fix

## Why

After more testing made by @ram130 in #870 after my changes from #894 we found it was not always working.

Previously the restore could silently fail due to a race condition when switching episodes, or get overridden by ExoPlayer re-selecting tracks after the initial restore. Both issues are now fixed - the preference is re-applied on each track update until ExoPlayer confirms the correct track is active, and cleared as soon as the user makes a manual change.

Additionally, the `maybeApplyRememberedAudioSelection` was now redundant, and it didn't notice that earlier, so I did a little clean-up  

## Policy check

## Testing

- Opened a movie, changed audio track and subtitles, exited player, returned via Continue Watching -> correct tracks restored
- Opened a series, changed tracks mid-episode, switched to next episode -> correct tracks restored
- Changed tracks on episode 3 (auto-restore failed), manually reselected on episode 3, moved to episode 4 -> episode 4 picked up the corrected preference
- Fresh install / no saved preference -> default track selection behavior unchanged

Nothing changed UI wise

## Breaking changes

Nothing should break 

## Linked issues

My changes were not released yet

### DIFF
--- a/app/src/main/res/values-pl/strings.xml
+++ b/app/src/main/res/values-pl/strings.xml
@@ -970,4 +970,10 @@
     <string name="watchlist_added">Dodano do listy</string>
     <string name="watchlist_removed">Usunięto z listy</string>
     <string name="account_sync_restart_note">Synchronizacja nie jest natychmiastowa na aktywnych urządzeniach. Uruchom ponownie to urządzenie po zalogowaniu lub aby pobrać zmiany wprowadzone gdzie indziej.</string>
+    <string name="audio_mix_label">Wzmocnienie (PCM)</string>
+    <string name="audio_mix_persist_on">Zachowaj między sesjami: WŁ</string>
+    <string name="audio_mix_persist_off">Zachowaj między sesjami: WYŁ</string>
+    <string name="audio_mix_range">Zakres: %1$d dB do %2$d dB</string>
+    <string name="audio_mix_range_saved">Zakres: %1$d dB do %2$d dB (zapisano)</string>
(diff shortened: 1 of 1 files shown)


ram130:
`If the user had subtitles disabled, that state is also persisted and restored - the player will keep subtitles off even if it would normally auto-select one based on the preferred language`


Now that's a big one. I waited so long for this. 😭😅

ram130:
After testing. Good to merge. @tapframe

skoruppa:
Yes, I also performed some heavy testing. It finally behaves as I wanted. Even when subtitles from addons are initialized after the stream is already playing

PR #920, opened 2026-03-18
Fix for #900. Use available pageSize with catalogs 

### LINKED ISSUE #900: [Bug]: Pagination Issue in loading Catalogs (skip value incorrect in Nuvio)
### Pre-flight checks

### App version / commit

0.4.20-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Chromecast 4K 

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

Other

### Steps to reproduce

Hello,

I’ve developed an addon where I implement lazy loading for catalog items (20 items per page) using defineCatalogHandler. I think this behavior was previously working correctly in Nuvio (not sure), but it no longer seems to function as expected.

In my catalog, when clicking “Load More”, the pagination skips 100 items instead of the expected 20 (the default limit). This differs from both the previous behavior in Nuvio and the current behavior in Stremio, and it results in missing items between pages.

It appears that Nuvio does not calculate the skip value based on the number of already displayed items (metas.length).

For example:

In Stremio: if the catalog currently displays 80 items, the next request uses skip=80.

In Nuvio: the skip value is always set to 100, regardless of how many items are already loaded.

Because of this:

With a limit of 20 items per page (default behavior), Stremio correctly loads the next 20 items.

In Nuvio, it always jumps ahead by 100 items, causing gaps in the catalog.

Could you please confirm if this is a bug or an intended behavior?

Thank you.

### Expected behavior

Logic: The skip value should be equal to the total number of items currently displayed (metas.length).

Pagination Flow:

Page 1: skip=0 (Displays items 1-20)

Page 2: User clicks "Load More" → Request sends skip=20 (Displays items 21-40)

Page 3: User clicks "Load More" → Request sends skip=40 (Displays items 41-60)

Result: A seamless, gap-free list of catalog items.

### Actual behavior

Logic: The skip value seems to jump by a fixed increment of 100, regardless of the limit set in the addon or the number of items already loaded.

Pagination Flow:

Page 1: skip=0 (Displays items 1-20)

Page 2: User clicks "Load More" → Request sends skip=100 (Displays items 101-120)

Result: Significant data gaps. Items 21 through 100 are completely skipped and never displayed in the catalog.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

skoruppa:
## Summary

Fix catalog pagination skip step to correctly respect addon-declared `pageSize` and prevent stale skip values across reloads.

## PR type

- Bug fix

## Why

Two related issues caused incorrect pagination offsets:

**1. Dead logic in `skipStep()`**

Addons can declare a `pageSize` field in their catalog manifest (e.g. AIOMetadata uses `pageSize: 50`). NuvioTV was attempting to derive the skip step from numeric `options` on the `skip` extra field, based on [Stremio documentation](https://github.com/Stremio/stremio-addon-sdk/blob/master/docs/api/requests/defineCatalogHandler.md) - however this pattern does not exist in any real addon manifest. In practice `skip` extra is always declared as `{"name": "skip"}` with no options, so the calculation always fell through to the hardcoded default of 100.

**2. Stale skip step surviving catalog reloads**

`CatalogRepositoryImpl` already had a dynamic correction (`effectiveSkipStep`) that adjusts the skip step based on actual items returned on page 0 - e.g. if an addon without `pageSize` returns 20 items, `skipStep` is corrected to 20. However, `skipStep` was included as part of the cache key. This caused a cache miss on subsequent reloads (same session), so the ViewModel's `catalogsMap` retained the old `CatalogRow` with `skipStep=100`. When the user scrolled to load more, `nextSkip = (0+1) * 100 = 100` was sent instead of the correct offset

## Fix

- Simplified `skipStep()` to use `pageSize` from the manifest when present, falling back to 100
- Removed `skipStep` from `buildCacheKey` - it is not part of the request identity and was causing cache misses that exposed the stale value

## Testing

- Verified with AIOMetadata (`pageSize: 50`) - pagination now follows `skip=0, 50, 100, ...` instead of `skip=0, 100, 200, ...`
- Verified addons without `pageSize` - dynamic correction from page 0 response applies correctly across reloads, fallback remains 100

## Policy check

## Breaking changes

Nothing should break

## Linked issues

Fixes #900

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/repository/CatalogRepositoryImpl.kt
+++ b/app/src/main/java/com/nuvio/tv/data/repository/CatalogRepositoryImpl.kt
@@ -158,6 +158,6 @@ class CatalogRepositoryImpl @Inject constructor(
             .sortedBy { it.key }
             .joinToString("&") { "${it.key}=${it.value}" }
         val normalizedBaseUrl = addonBaseUrl.trim().trimEnd('/').lowercase()
-        return "${normalizedBaseUrl}_${addonId}_${type}_${catalogId}_${skip}_${skipStep}_${normalizedArgs}"
+        return "${normalizedBaseUrl}_${addonId}_${type}_${catalogId}_${skip}_${normalizedArgs}"
     }
 }


fcarsenal12:
@skoruppa  I would also like to add that Stremio appears to ignore the pageSize field in the manifest, even when it is explicitly declared (as seen in other addons like AIOMetadata). In my own addon, I did not define a pageSize at all.

Instead, Stremio dynamically calculates the pagination by checking the actual item count returned in the first request (metas.length) and automatically using that value as the skip increment for subsequent pages.

PR #1011, opened 2026-03-23
fix: Inconsistent vertical offset between built-in and some external addon subtitles

### LINKED ISSUE #983: [Bug]: Inconsistent vertical offset between built-in and addon subtitles
### Pre-flight checks

- [ ] I can reproduce this on the latest release or latest main build.

### App version / commit

v0.4.24-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Google Chromecast with Google TV 4k

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

### Steps to reproduce

1. Play a video with built-in subtitles. Observe their default vertical position (usually higher).
2. ​Switch to subtitles provided by an add-on. Observe that they are positioned significantly lower by default.
3. ​Increase the global subtitle vertical offset (e.g., set to +15) to push the add-on subtitles to a standard, readable position.
4. ​Switch back to the built-in subtitles.

### Expected behavior

Both built-in and add-on subtitles should share the same default baseline position. Applying a global vertical offset should adjust both uniformly without making one of them unusable.

### Actual behavior

The global setting applies the offset to different baselines. Fixing the position for add-on subtitles pushes the built-in subtitles too high up the screen. The setting acts globally, but the sources are not aligned by default, making it impossible to find a value that works for both.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

kernexshadow:
## Summary
This PR fixes inconsistent vertical positioning for some external addon subtitles by normalizing externally loaded WebVTT cue positioning before rendering.

The player now clears embedded vertical cue positioning for external addon WebVTT tracks so they follow the same subtitle offset behavior as the rest of the player.

## PR type
Bug fix

## Why
This change is needed because users reported inconsistent subtitle placement between built-in subtitles and addon subtitles.

Some addons, such as OpenSubtitles Pro, could provide subtitles as external WebVTT tracks whose cue positioning caused them to render noticeably lower than built-in subtitles or other external subtitle formats.

Context:

[Bug]: Inconsistent vertical offset between built-in and addon subtitles #983

During investigation, the issue was narrowed down to WebVTT cue positioning itself:

built-in subtitles and external SubRip/SRT subtitles were aligned correctly
some external addon subtitles delivered as WebVTT carried cue line positioning that caused them to render lower
the previous approach compensated this in the UI, but that made manual subtitle offset behavior inconsistent
The latest PR implementation fixes the issue at the subtitle rendering path instead of adding a visual compensation layer.

Behavior change
For external addon WebVTT subtitles, embedded vertical cue positioning is normalized before cues reach SubtitleView.

This means:

external addon WebVTT subtitles now follow the same player offset behavior as other subtitle types
the manual subtitle vertical offset setting remains predictable
the previous WebVTT-specific UI compensation is no longer needed
Policy check
[X] This PR is not cosmetic-only, unless it is a translation PR.
[X] This PR does not add a new major feature without prior approval.
[X] This PR is small in scope and focused on one problem.
[ ] If this is a larger or directional change, I linked the issue where it was approved.

## Testing
Manual:

Compared built-in subtitles against external addon WebVTT subtitles in the player
Verified that external addon WebVTT subtitles now follow the normal subtitle offset path
Verified that built-in subtitles and external SubRip/SRT subtitles were unchanged
Verified that switching between subtitle sources during playback still worked correctly
Build:

:app:assembleDebug
Screenshots / Video (UI changes only)
None.

## Breaking changes
None expected, but external addon WebVTT subtitles that rely on embedded vertical cue positioning will now follow the player’s standard subtitle positioning behavior.

## Linked issues
Fixes #983

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
@@ -7,9 +7,12 @@ import android.util.Log
 import android.view.accessibility.CaptioningManager
 import androidx.media3.common.AudioAttributes
 import androidx.media3.common.C
+import androidx.media3.common.MimeTypes
 import androidx.media3.common.PlaybackException
 import androidx.media3.common.Player
 import androidx.media3.common.Tracks
+import androidx.media3.common.text.Cue
+import androidx.media3.common.text.CueGroup
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.exoplayer.DefaultLoadControl
 import androidx.media3.exoplayer.DefaultRenderersFactory
@@ -172,6 +175,11 @@ internal fun PlayerRuntimeController.initializePlayer(url: String, headers: Map<
             val renderersFactory = SubtitleOffsetRenderersFactory(
                 context = context,
                 subtitleDelayUsProvider = subtitleDelayUs::get,
+                shouldNormalizeCuePositionProvider = {
+                    val selectedAddonSubtitle = _uiState.value.selectedAddonSubtitle
+                    selectedAddonSubtitle != null &&
+                        PlayerSubtitleUtils.mimeTypeFromUrl(selectedAddonSubtitle.url) == MimeTypes.TEXT_VTT
+                },
                 gainAudioProcessor = gainAudioProcessor
             ).setExtensionRendererMode(playerSettings.decoderPriority)
                 .setMapDV7ToHevc(playerSettings.mapDV7ToHevc)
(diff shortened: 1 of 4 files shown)


gecXMAN:
I notices that problem is getting worse when  "Stremio Community Subtitles" addon is used !

kernexshadow:
I can't find any subtitles of  any movie/show with stremio community subtitles. Nevertless PR fix this Offset.

gecXMAN:
> I can't find any subtitles of any movie/show with stremio community subtitles. Nevertless PR fix this Offset.

You must select it on web page of addon or upload it by yourself and then it will be automaticly selected.

kernexshadow:
@skoruppa 
I changed the approach in this PR and would like a fresh review.

Originally this was implemented as a fixed UI-side compensation for external WebVTT subtitles. I replaced that with a rendering-path fix instead: external addon WebVTT cues now have their embedded vertical cue positioning normalized before reaching SubtitleView, so they follow the same subtitle offset behavior as the rest of the player.

This removes the WebVTT-specific compensation logic from the player UI and keeps the manual subtitle vertical offset predictable.

I have tested for different offsets, and it's always consistent now.

skoruppa:
@kernexshadow looks good. I think we can merge it @tapframe

PR #1004, opened 2026-03-23
Fix Continue Watching focus not landing on the first item

skoruppa:
## Summary

On app startup, the CW pipeline emits data in multiple intermediate batches before the final sorted list is ready (`inProgressOnly`, partial Next Up items). Each intermediate render was immediately pushed to the UI, on cold start causing focus to land on whichever item appeared first - not necessarily the most recently watched one.

The fix suppresses all intermediate CW renders during the initial pipeline cycle. Only `normalItems` (the final merged and sorted list from the `merge-lightweight` phase) triggers the first render, atomically clearing `cwIsLoading` at the same time. Subsequent pipeline cycles (e.g. after a data change) are unaffected.

Key changes:
- `cwIsLoading: Boolean` added to `HomeUiState` (true until the first full pipeline cycle completes)
- Intermediate renders (`inProgressOnly`, partial Next Up) are skipped when `cwIsLoading` is true
- `normalItems` atomically sets `continueWatchingItems` and clears `cwIsLoading` in a single `update` call

## PR type

- Bug fix

## Why

When using Nuvio Sync, CW data arrives in batches. The pipeline was publishing each intermediate batch immediately, so the CW row could appear multiple times with a different item order before settling. Focus would land on the first item that appeared and stay there, even if the final sorted list put a different item first.

https://github.com/user-attachments/assets/9ca9c6a8-0a49-413a-aac8-bdfcc479fb6a

## Policy check

- [x] This PR is not cosmetic-only, unless it is a translation PR.
- [x] This PR does not add a new major feature without prior approval.
- [x] This PR is small in scope and focused on one problem.
- [ ] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

Reproduced by force-stopping the app and relaunching with Nuvio Sync enabled. Tested across all three display modes (Modern, Classic, Grid) with the same Continue Watching list (on both regular 0.5.0-beta and with my fix)

- [x] Modern - focus lands on correct first CW item on cold start
- [x] Classic - same
- [x] Grid - same

## Screenshots / Video (UI changes only)

Nothing changed

## Breaking changes

Nothing should break 

## Linked issues

No one reported the issue yet

Issue #849, opened 2026-03-15
[Bug]: inconsistent behavior in "continue watching"


mindgam3s:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

0.4.19

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

FireTV Cube Gen 3

### Android version

FireTV

### Player mode

Internal player

### Area (tag)

UI / Layout

### Steps to reproduce

- have an item in continue watching 
- it might either display the show's main poster (e.g. `Paradise (2025)` OR an unblurred poster for the next episode (e.g. `South Park`)

![continue watching](https://github.com/user-attachments/assets/650d439b-4cb9-4c36-9bac-02c2401ea64d)

### Expected behavior

either use show's main poster or blur episode specific poster 

### Actual behavior

unblurred poster OR show's main poster

### Frequency

Always

### Did this work before?

No, it never worked

### Logs (required for crash reports)

```shell

```

PR #1039, opened 2026-03-24
Fixes for CW after #1027

skoruppa:
## Summary

Add a dedicated setting to control blur for the Continue Watching row and fix related regressions.

- Added a new toggle in Settings to enable/disable blur specifically for Continue Watching (independent of the global blur setting introduced in #1027)
- Fixed TMDB enrichment, so movies are correctly enriched in the Continue Watching section
- Reverted PR #1004 which fixed initial focus for the first CW item but reintroduced a bug where the CW row was incorrectly focused when Trakt was enabled

## PR type

- Bug fix

## Why

In #1027 I introduced blur for Continue Watching per request in #849, but many users disliked the change. Instead of removing it entirely, this PR gives users control over it via a separate setting in **Home Content** section. Additionally, TMDB enrichment was broken for movies in CW, and the focus fix from #1004 caused a regression with Trakt-enabled CW row focus.

## Policy check

## Testing

- Verified blur toggle works independently for Continue Watching row
- Confirmed TMDB enrichment correctly resolves movie metadata in CW
- Tested CW row focus behavior with Trakt enabled/disabled after #1004 revert
- Manual testing on Android TV emulator and physical device

Nothing changed 

## Breaking changes

Nothing should break 

## Linked issues

Relates to #849
Reverts #1004
Follows up on #1027

### DIFF
--- a/app/src/main/res/values-pl/strings.xml
+++ b/app/src/main/res/values-pl/strings.xml
@@ -363,6 +363,8 @@
     <string name="layout_expand_delay_sub">Jak długo czekać przed rozwinięciem zaznaczonych kart.</string>
     <string name="layout_autoplay_trailer">Auto-odtwarzanie zwiastuna</string>
     <string name="layout_autoplay_trailer_expanded">Auto-odtwarzanie zwiastuna w rozszerzonej karcie</string>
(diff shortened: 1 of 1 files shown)


ram130:
Looks good

PR #1048, opened 2026-03-25
Changes for fullscreen gradients

### LINKED ISSUE #1033: [Feature]: On the new expanded backdrop image, add a fade from bottom to top, so the texts and cards are not mixed with background
### Area (tag)
UI / Layout

### Problem statement

First, thanks for implementing what I requested! (https://github.com/NuvioMedia/NuvioTV/issues/897)

If possible, add a bit of a fade horizontally (a bit more, but softer, so the white texts from the descriptions don't mix with the background in case the image is too bright).

And also, most important, add a vertical fade too, from bottom to top, because right now the poster cards are a bit "mixed" with the background... there is a fade from the left side, so the same thing but softer.. 

https://drive.google.com/file/d/1zjEnLpDjL0GoB2y8QajIwhO1SiRiIA0w/view?usp=sharing

### Proposed solution

Just add a SOFT fade/shadow from bottom to top...
And also expand the current fade from left to right too, make it softer but longer.

My suggestion, as I originally posted (Also notice how it's not a pure black color, it has just a bit of blue (#19171f)

Image

No, this is only a proposal

_No response_

_No response_

skoruppa:
## Summary

Updated horizontal and vertical gradients in the hero section to better fit the fullscreen backdrop mode. Gradients are now stronger and cover more area when fullscreen backdrop is enabled, improving readability of overlaid UI elements. Standard mode is unchanged.

## PR type

## Why

The existing gradients didn't look right with fullscreen backdrop enabled - hero section description lacked sufficient contrast against hero images.  Especially with light backdrop images. Same situation with posters cards.

Addresses #1033

## Policy check

## Testing

Manual testing on Android TV - verified gradient appearance in both fullscreen and standard backdrop modes across different hero images.

Before
IMG20260325063726

After
IMG20260325063710

## Breaking changes

Nothing should break

## Linked issues

Fixes #1033

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeHero.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeHero.kt
@@ -138,25 +138,46 @@ internal fun ModernHeroGradientLayer(
     Box(
         modifier = modifier
             .drawWithCache {
-                val horizontalFadeEndX = size.width * if (isFullScreen) 0.55f else 0.45f
+                val horizontalFadeEndX = size.width * if (isFullScreen) 0.65f else 0.45f
                 val horizontalGradient = Brush.horizontalGradient(
-                    colorStops = arrayOf(
-                        0.0f to bgColor,
-                        0.22f to bgColor.copy(alpha = 0.86f),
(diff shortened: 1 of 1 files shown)


marcoshimselfxd:
🚀🚀

PR #1070, opened 2026-03-26
Perf/modernhome performance optimization

halibiram:
## Summary

- move Modern Home row mapping and presentation building out of the composable hot path and into a dedicated ViewModel pipeline
- introduce prebuilt Modern Home presentation state, row lookups, and row/item caches so focus, hero, and row access do not need repeated scans
- add `ModernHomePresentationInput`, `ModernHomePresentationState`, `ModernHeroSceneState`, and richer `CarouselRowLookups` models to centralize Modern Home rendering inputs and lookups
- add `rowKeyByGlobalRowIndex`, `firstHeroPreviewByRow`, and `fallbackBackdropByRow` lookups so focus restore and hero fallback selection can resolve directly from prepared state
- add `observeModernHomePresentationPipeline()` in `HomeViewModelPresentationPipeline.kt` and run presentation building on `Dispatchers.Default`
- warm a smaller initial Modern Home presentation first, then build the full presentation in the background to reduce startup work
- use a smaller warm-start row count depending on whether Continue Watching is present
- warm-start only the first `2` catalog rows when Continue Watching exists, otherwise the first `3` rows
- store the prebuilt Modern Home presentation in `HomeUiState` so the UI can consume a stable prepared structure
- inject `@ApplicationContext` into `HomeViewModel` and keep a reusable `ModernCarouselRowBuildCache` there for cross-rebuild row reuse
- make ModernHomeContent consume the prebuilt presentation instead of rebuilding rows locally
- switch focus row restoration from raw visible catalog row indexing to `rowKeyByGlobalRowIndex` from the prepared presentation state
- resolve active rows directly from `rowByKey` instead of rescanning the row list
- seed the initial hero item from the first prepared carousel item so the hero has content immediately
- resolve focus rows from the presentation state and improve pending row focus restore behavior
- limit row image prefetch work to the active row and reduce extra work during fast vertical and horizontal navigation
- gate row image prefetch work while vertical scrolling is active and only start that work for the active row
- add a shared `buildPrefetchRequest()` helper and prefetch hero backdrop images at full-screen size when needed
- optimize D-pad repeat navigation with safer directional repeat handling while keeping OK long-press behavior intact
- restrict synthetic repeat handling to D-pad directions only, use axis-specific throttle windows, and dispatch movement through `focusManager.moveFocus()`
- use `112ms` repeat gating for vertical D-pad movement and `80ms` repeat gating for horizontal D-pad movement
- stabilize the Modern hero scene during scroll and prefetch hero backdrops at full-screen size where needed
- introduce `ModernHeroScene()` as a wrapper for the media layer plus gradient layer so hero scene state can be frozen and reused more cleanly
- keep a stable hero scene snapshot while vertical scrolling is active to avoid visible hero/backdrop churn during fast movement
- tune rapid hero navigation handling with a `130ms` rapid-nav threshold and `170ms` settle delay for fast focus moves
- keep landscape card backdrop and logo art stable while metadata enrichment updates in the background
- carry forward frozen backdrop and logo URLs through cached carousel items so the first non-blank art survives cache rebuilds and navigation
- persist the current hero backdrop across Home -> Detail navigation and reuse it in detail loading to avoid a blank or flashing transition
- add `HeroBackdropState`, pass `heroBackdropUrl` through the detail route, and hydrate the detail screen with the current Home hero backdrop
- add a hero-backed detail loading transition and backdrop-aware details skeleton state
- make `MetaDetailsSkeleton` backdrop-aware, add `BackdropLoadingScrim()`, and soften episode card / shimmer treatment when loading over a reused backdrop
- reuse the seeded hero backdrop as an underlay in `MetaDetailsContent` and avoid unnecessary crossfade when the detail backdrop matches the carried Home hero backdrop
- use a `400ms` backdrop crossfade only when swapping from the carried Home hero backdrop to a different detail backdrop, and disable crossfade when the backdrop can be reused directly
- clean up initial home startup behavior, including the startup loading gate and initial content enter behavior
- skip the initial Home enter transition on first render, but keep the animated reveal for later visibility changes
- keep the existing startup Continue Watching gate timeout at `5000ms`, but avoid applying the content enter animation on the very first successful reveal
- pre-warm reverse TMDB cache during hero enrichment to make later metadata work return faster from cache

## PR type

## Why

- Modern Home still did too much work during initial composition, row building, and fast remote navigation
- focus, hero, and row selection logic were doing repeated lookups that could be precomputed once
- row image work and prefetching were too aggressive while moving quickly with the remote
- hero and backdrop updates could briefly flicker during enrichment, Home -> Detail transitions, and detail loading
- detail loading did not reuse the active Home hero backdrop, so the transition could feel visually disconnected
- this keeps the existing UI and interaction model intact while making startup, focus movement, scrolling, and transitions feel more stable

## Policy check

## Testing

- manually tested Modern Home cold start, warm start, and first content render timing
- manually tested continue watching startup gating and initial home content reveal
- manually tested row scrolling, focus movement, pending focus restore, and rapid D-pad repeat input
- manually tested active-row-only prefetch behavior during fast navigation
- manually tested Home -> Detail -> Back transitions and verified hero/backdrop continuity
- manually tested detail loading states and confirmed the seeded hero backdrop is shown during loading
- manually tested TMDB-enriched items to confirm row art stays visually stable after enrichment

- None

## Breaking changes

- None

## Linked issues

- None

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -44,8 +44,8 @@ android {
         applicationId = "com.nuvio.tv"
         minSdk = 24
         targetSdk = 36
-        versionCode = 44
-        versionName = "0.5.1-beta"
+        versionCode = 46
+        versionName = "0.5.3-beta"
 
         buildConfigField("String", "PARENTAL_GUIDE_API_URL", "\"${localProperties.getProperty("PARENTAL_GUIDE_API_URL", "")}\"")
         buildConfigField("String", "INTRODB_API_URL", "\"${localProperties.getProperty("INTRODB_API_URL", "")}\"")
--- a/app/src/main/java/com/nuvio/tv/data/remote/api/TraktApi.kt
+++ b/app/src/main/java/com/nuvio/tv/data/remote/api/TraktApi.kt
@@ -15,6 +15,7 @@ import com.nuvio.tv.data.remote.dto.trakt.TraktListItemDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktListItemsMutationRequestDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktListItemsMutationResponseDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktListSummaryDto
+import com.nuvio.tv.data.remote.dto.trakt.TraktMovieDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktPlaybackItemDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktReorderListsRequestDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktReorderListsResponseDto
@@ -24,6 +25,7 @@ import com.nuvio.tv.data.remote.dto.trakt.TraktScrobbleRequestDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktScrobbleResponseDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktSearchResultDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktSeasonDto
+import com.nuvio.tv.data.remote.dto.trakt.TraktShowDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktShowProgressResponseDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktTokenResponseDto
 import com.nuvio.tv.data.remote.dto.trakt.TraktUserEpisodeHistoryItemDto
@@ -168,6 +170,24 @@ interface TraktApi {
         @Query("page") page: Int = 1,
         @Query("limit") limit: Int = 10
     ): Response<List<TraktCommentDto>>
+
+    @GET("movies/{id}/related")
+    suspend fun getMovieRelated(
+        @Header("Authorization") authorization: String? = null,
+        @Path("id") id: String,
+        @Query("extended") extended: String = "full,images",
+        @Query("page") page: Int = 1,
+        @Query("limit") limit: Int = 20
+    ): Response<List<TraktMovieDto>>
+
+    @GET("shows/{id}/related")
+    suspend fun getShowRelated(
+        @Header("Authorization") authorization: String? = null,
+        @Path("id") id: String,
+        @Query("extended") extended: String = "full,images",
+        @Query("page") page: Int = 1,
+        @Query("limit") limit: Int = 20
+    ): Response<List<TraktShowDto>>
 
     @GET("search/{id_type}/{id}")
     suspend fun searchById(
--- a/app/src/main/java/com/nuvio/tv/data/remote/dto/trakt/TraktMediaDtos.kt
+++ b/app/src/main/java/com/nuvio/tv/data/remote/dto/trakt/TraktMediaDtos.kt
@@ -16,14 +16,49 @@ data class TraktIdsDto(
 data class TraktMovieDto(
     @Json(name = "title") val title: String? = null,
     @Json(name = "year") val year: Int? = null,
-    @Json(name = "ids") val ids: TraktIdsDto? = null
+    @Json(name = "ids") val ids: TraktIdsDto? = null,
+    @Json(name = "tagline") val tagline: String? = null,
+    @Json(name = "overview") val overview: String? = null,
+    @Json(name = "released") val released: String? = null,
+    @Json(name = "runtime") val runtime: Int? = null,
+    @Json(name = "country") val country: String? = null,
+    @Json(name = "status") val status: String? = null,
+    @Json(name = "rating") val rating: Double? = null,
+    @Json(name = "genres") val genres: List<String>? = null,
+    @Json(name = "certification") val certification: String? = null,
+    @Json(name = "languages") val languages: List<String>? = null,
+    @Json(name = "original_title") val originalTitle: String? = null,
+    @Json(name = "images") val images: TraktImagesDto? = null
 )
 
 @JsonClass(generateAdapter = true)
 data class TraktShowDto(
     @Json(name = "title") val title: String? = null,
     @Json(name = "year") val year: Int? = null,
-    @Json(name = "ids") val ids: TraktIdsDto? = null
+    @Json(name = "ids") val ids: TraktIdsDto? = null,
+    @Json(name = "tagline") val tagline: String? = null,
+    @Json(name = "overview") val overview: String? = null,
+    @Json(name = "first_aired") val firstAired: String? = null,
+    @Json(name = "runtime") val runtime: Int? = null,
+    @Json(name = "network") val network: String? = null,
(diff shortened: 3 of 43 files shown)


halibiram:
The HeroBackdrop doesn't re-render anymore when navigating between Home and MetaDetail


https://github.com/user-attachments/assets/8bd0e99b-06d9-4697-a94b-c26452e7da8e

halibiram:
I couldn't test CW due to missing an API key, so I can't guarantee it works as expected. This commit may need to be reverted if any issues arise.

[perf: decouple home startup from continue watching sync](https://github.com/NuvioMedia/NuvioTV/pull/1070/commits/6c2acafec1338810edcfc4976ef4eda10b9df107)

skoruppa:
> The HeroBackdrop doesn't re-render anymore when navigating between Home and MetaDetail
> document_5918189308656755363_compressed.mp4

What if the backdrop is different - tmdb enrichment enabled on home, but not on details?

> limits image prefetching to the active row only 

Shouldn't that be +1? We usually see active row and part of the second one

halibiram:
> What if the backdrop is different - tmdb enrichment enabled on home, but not on details?

Good point! Both Home and Detail read from the same global `TmdbSettings.enabled` toggle in `TmdbSettingsDataStore`, so they're always in sync — if `enabled=true`, both screens get the TMDB backdrop; if `enabled=false`, neither does.

The only Home-specific conditional is `modernHomeEnabled`, which can suppress enrichment on Modern Home layout. But in that case Home also falls back to the addon backdrop, so the only possible mismatch is the **reverse** of what you described: Detail shows a higher-quality TMDB backdrop while Home stays on the addon image. There's no combination where Home gets TMDB and Detail doesn't.

That said, since Detail always loads addon meta first and then applies TMDB enrichment asynchronously, a brief flash can occur when the backdrop URL switches — especially if the TMDB image isn't already in Coil's cache from Home. I'll add a crossfade transition to the backdrop image loading to smooth out that transition.

halibiram:
> Shouldn't that be +1? We usually see active row and part of the second one

Good catch — you're right, the peeking row at the bottom should be included in the warm start. I'll bump both values by +1.

skoruppa:
@halibiram resolve conflicts and I can prepare a build for you to test CW. If possible check for both Nuvio Sync and Trakt if the wrong row focus problem was not reintroduced

kernexshadow:
The Startup gate removed, I don't think it's a good idea. That way the landing row will not be the CW. Users said that they prefer to wait for cw than having focus on other rows.

halibiram:
> @halibiram resolve conflicts and I can prepare a build for you to test CW. If possible check for both Nuvio Sync and Trakt if the wrong row focus problem was not reintroduced

@skoruppa  I've resolved the merge conflicts. If you're available, could you run a build?

halibiram:
> The Startup gate removed, I don't think it's a good idea. That way the landing row will not be the CW. Users said that they prefer to wait for cw than having focus on other rows.

You're right, let's just give it a try. In some cases, this timeout delays the app launch by 5 seconds. If it doesn't work out like this, we can just revert that change.

halibiram:
> The Startup gate removed, I don't think it's a good idea. That way the landing row will not be the CW. Users said that they prefer to wait for cw than having focus on other rows.

I have a suggestion. Rather than blocking the UI for CW, we could display a shimmer placeholder. This would indicate to the user that the data is still loading.

skoruppa:
> > The Startup gate removed, I don't think it's a good idea. That way the landing row will not be the CW. Users said that they prefer to wait for cw than having focus on other rows.
> 
> I have a suggestion. Rather than blocking the UI for CW, we could display a shimmer placeholder. This would indicate to the user that the data is still loading.

If you can implement it, go for it. I tried and was not happy about the result :P remember we need to handle
- local CW
- Nuvio sync CW
- Trakt CW

Anyway, here is [the build](http://morigal.pl/upload/app-universal-benchmark-perf.apk). And as expected, with trakt enabled the initial focus is few rows bellow ;) We don't want to reintroduce this issue

halibiram:
> > > The Startup gate removed, I don't think it's a good idea. That way the landing row will not be the CW. Users said that they prefer to wait for cw than having focus on other rows.
> > 
> > 
> > I have a suggestion. Rather than blocking the UI for CW, we could display a shimmer placeholder. This would indicate to the user that the data is still loading.
> 
> If you can implement it, go for it. I tried and was not happy about the result :P remember we need to handle
> 
> * local CW
> * Nuvio sync CW
> * Trakt CW
> 
> Anyway, here is [the build](http://morigal.pl/upload/app-universal-benchmark-perf.apk). And as expected, with trakt enabled the initial focus is few rows bellow ;) We don't want to reintroduce this issue

Thanks for the test build! I see why the timeout is needed now—if we skip the profile screen too fast, the CW row loses focus (though everything else works fine). So, I'll go ahead and revert that commit.

halibiram:
Additionally, despite all optimizations, vertical scroll rendering still lags behind. To fix this, I introduced a 112ms (7-frame) delay for long presses, which I confirmed through profiling. When the cache is warm, it only takes about 3 frames to render, but a stable render speed is better for the UX than sudden lag spikes.

halibiram:
@tapframe I've completed my work and it's ready to be merged. Could you take a look?

halibiram:
Latest scroll implementation


https://github.com/user-attachments/assets/48afad09-3cc0-45ff-bdf5-ecf72674753f

skoruppa:
It is really smooth right now :D And I don't see any issues. Trakt's CW still has correct focus. Fullscreen and regular modern views both works.Horizontal and Vertical posters are fine, trailers player as epxected. 

 I'm only thinking if we could take over the backdrop to the shimmer details loading screen that is visible for a second when you are entering details screen for the first time

halibiram:
> I'm only thinking if we could take over the backdrop to the shimmer details loading screen that is visible for a second when you are entering details screen for the first time

That's a great idea! We could pass the herobackdrop as an argument during navigation. That way, it loads immediately behind the shimmer effect, making the transition completely seamless.Let me look into how we can implement it.

halibiram:
> It is really smooth right now :D And I don't see any issues. Trakt's CW still has correct focus. Fullscreen and regular modern views both works.Horizontal and Vertical posters are fine, trailers player as epxected.
> 
> I'm only thinking if we could take over the backdrop to the shimmer details loading screen that is visible for a second when you are entering details screen for the first time

Thanks for the feedback! I made the change and it turned out great. When the fullscreen hero backdrop is enabled, the image stays in the background seamlessly all the way to the source screen. By making the skeleton transparent, it looks exactly like what you suggested. I also fixed an issue where the image would disappear when the fullscreen hero backdrop was disabled. In both cases, that black background is gone now. It will only show black if there is genuinely no backdrop at all.

skoruppa:
Green light from me :D everything works and feels great. Now you need to do the same for other layouts as well ;)

PR #1079, opened 2026-03-26
fix(profile): localize hardcoded edit title and cleanup unused strings

SK-Gits:
## Summary
The profile modification screen was displaying a hardcoded title in the code. This PR fixes the localization by using an existing but unreferenced string resource and performs a thorough cleanup of XML files across all languages.

## PR type
- Bug fix
- Translation update

## Why
The title was previously locked in English because it was not linked to the Android resource system. This change allows the title to be displayed in the user's system language.

## Policy check

## Testing
Manual Test: Verified on-device using Android Studio Debug mode; the title now correctly follows the system language.

Reference Check: Manually searched the codebase and confirmed that neither profile_edit_title_id nor the section header were referenced before this PR.

Before
After

## Breaking changes
None.

## Linked issues
None.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/profile/ProfileSelectionScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/profile/ProfileSelectionScreen.kt
@@ -1374,7 +1374,7 @@ private fun EditProfileOverlay(
             ) {
                 Column(verticalArrangement = Arrangement.spacedBy(2.dp)) {
                     Text(
-                        text = "Edit Profile",
+                        text = stringResource(R.string.profile_edit_title),
                         color = NuvioColors.TextSecondary,
                         fontSize = 16.sp,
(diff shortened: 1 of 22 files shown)


SK-Gits:
Technical Details
Code Change: Replaced the hardcoded string "Edit Profile" with stringResource(R.string.profile_edit_title) in ProfileSelectionScreen.kt (line 1377).

Resource Recovery: Utilized the profile_edit_title key, which was already present in XML files but was not referenced anywhere in the codebase before this change.

Global Refactor (All Languages): In every strings.xml file of the project, removed the %1$s formatting code from profile_edit_title to make it a static string.

XML Cleanup (All Languages): Deleted the redundant section !-- ProfileSettingsContent --> which only contained the profile_edit_title_id key. This section and key were removed from all language files as they were completely unused across the entire project (verified with 0 occurrences found after a manual search).

PR #758, opened 2026-03-10
Better handling of Spanish embedded subtitles

kernexshadow:
## Summary

Fixes embedded subtitle auto-selection for Spanish regional variants so Spanish prefers Spain/European tracks instead of defaulting to Latin America in common CR-style naming patterns.

Added Spanish regional tag groups (Spain/European vs LatAm).
Added Spanish tie-break logic for embedded tracks (es vs es-419), mirroring the existing Portuguese regional behavior.
Added fallback for es-419 when track language is generic es but label/id indicates LatAm.

## PR type

- Bug fix

## Why

Users selecting Spanish were often auto-assigned Latin America subtitles even when Spain/European subtitles were available (for example CR Español vs Latin_America_CR Español). This corrects default embedded subtitle selection behavior.

## Policy check

 - [X] This PR is not cosmetic-only, unless it is a translation PR.
- [X] This PR does not add a new major feature without prior approval.
- [X] This PR is small in scope and focused on one problem.
- [ ] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

User tested in #692 

## Screenshots / Video (UI changes only)

N/A (logic-only change, no UI visual changes).

## Breaking changes

None.

## Linked issues

Fixes #692

PR #1083, opened 2026-03-26
Split regional subtitle variants into separate language tabs (pt, es)

### LINKED ISSUE #692: [Feature]: Better handling of Spanish subtitles
### Area (tag)

Subtitles

### Problem statement

Spanish subtitles are different depending if the subtitle is from Spain or from Latin America. If you select the preferred subtitle language to Spanish it almost always defaults to the Latin America one.
For example most of the animes with CR subs, Spain one is named CR Español, Latin one is Latin_America_CR Español. Ssometimes the Spanish subtitle is also named European or just Spanish.

Image
Image

### Proposed solution

Please make a way to set the Spanish subtitles from Spain as default if possible.

No, this is only a proposal

_No response_

_No response_

skoruppa:
## Summary

Split regional subtitle variants (Brazilian Portuguese, European Portuguese, Latin American Spanish, Castilian Spanish) into separate language tabs in the subtitle selection overlay. Previously with embed subtitles all variants were merged under one generic tab, causing wrong auto-selection and making it hard to find the right accent.

Based in part on PR #758 by @kernexshadow.

## PR type

- Bug fix

## Why

Embedded subtitle tracks typically report a generic language code (`por`, `spa`) regardless of regional variant. This caused:
- Brazilian and European Portuguese merged into one "Portuguese" tab
- Latin American and Castilian Spanish merged into one "Spanish" tab
- Auto-selection picking the wrong variant (e.g. Brazilian when user wants European, simply because it was first in the list)
- Persisted track restore falling back to the wrong variant on episode switch

The fix detects regional accents from track name/label tags (e.g. "Brazil", "Latino", "Portugal", "Castilian") and groups them into separate tabs. Addon subtitles already return proper language codes (`pob`, `es-419`) so they were already grouped correctly - now embedded tracks match.

## Policy check

## Testing

- Manual: verified PT-BR and PT tracks split into separate tabs when embedded tracks have accent tags in name/label
- Manual: verified ES-419 and ES tracks split into separate tabs
- Manual: verified auto-selection picks correct variant (PT-BR user gets Brazilian, PT user gets European)
- Manual: verified auto-selection prefers non-forced tracks over forced when multiple Brazilian tracks exist
- Manual: verified persisted track restore falls back to addon with matching variant when internal track is missing on episode switch
- Manual: verified PT user does not get Brazilian subtitles by accident (neither via auto-select nor via persisted restore)
- Provided test builds to user using PT and PT-BR

IMG20260326130348
IMG20260326130315

## Breaking changes

Nothing should break - changes are only for Spanish and Portuguese subtitles 

## Linked issues

Fixes #692

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
@@ -57,12 +57,6 @@ class PlayerRuntimeController(
         internal const val TAG = "PlayerViewModel"
         internal const val TRACK_FRAME_RATE_GRACE_MS = 1500L
         internal const val ADDON_SUBTITLE_TRACK_ID_PREFIX = "nuvio-addon-sub:"
-        internal val PORTUGUESE_BRAZILIAN_TAGS = listOf(
-            "pt-br", "pt_br", "pob", "brazilian", "brazil", "brasil"
-        )
-        internal val PORTUGUESE_EUROPEAN_TAGS = listOf(
-            "pt-pt", "pt_pt", "iberian", "european", "portugal", "europeu"
-        )
     }
 
     internal data class PendingAudioSelection(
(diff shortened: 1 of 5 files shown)


marcoshimselfxd:
Damn, this is incredible

PR #1180, opened 2026-03-31
Series Watched Badges - trakt improvements

skoruppa:
## Summary

Overhauled Trakt series watched badge resolution and improved TMDB-only ID fallback for Trakt API endpoints.

## PR type

- Bug fix

## Why

After user testing it became clear that Trakt watched status handling needed further improvement. The Continue Watching seed-based approach for badge calculation was unreliable. Additionally, series with only a TMDB ID (no IMDB ID) were never resolved against Trakt's `/shows/{id}/progress/watched` endpoint, since Trakt does not accept `tmdb:X` in path segments.

This is a continuation of #1074 and #1111.

## Changes

- Completely removed Continue Watching seed-based badge calculation. Badges are now computed solely from `/sync/watched/shows`, storing IMDB, TMDB, and Trakt IDs for broader compatibility.
- For every returned series, we fetch meta to determine total episode count. Meta lookups are throttled to avoid overwhelming the meta addon.
- Once a series badge status is determined (fully watched or not), the result is cached with a 12h TTL so we don't re-check all meta on every app launch.
- Improved TMDB-only ID fallback for Trakt API path endpoints. When `tmdbService.tmdbToImdb` returns nothing, we now:
  1. Check local `showIdToTraktPathId` cache (built from `/sync/watched/shows` - most compatible, instant)
  2. Fall back to `tmdbService.tmdbToImdb()` lookup
  3. Last resort: Trakt search API `/search/tmdb/{id}?type=show`
- Injected `TmdbService` into `TraktProgressService` for TMDB->IMDB resolution.

## Policy check

## Testing

- Tested with Nuvio Sync as watch progress source
- Tested separately with a Trakt account containing a large watch history (~200 watched series)
- Verified TMDB-only series (tmdb:225634, tmdb:286801) now correctly show watched badges
- Verified 12h TTL cache prevents redundant meta lookups on subsequent launches

N/A

## Breaking changes

Nothing should break 

## Linked issues

Continuation of #1074 and #1111

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/WatchedSeriesStateHolder.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/WatchedSeriesStateHolder.kt
@@ -1,8 +1,11 @@
 package com.nuvio.tv.data.local
 
 import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringPreferencesKey
 import androidx.datastore.preferences.core.stringSetPreferencesKey
 import com.nuvio.tv.core.profile.ProfileManager
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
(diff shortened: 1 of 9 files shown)


ram130:
Looks good but add a check every few mins to check for new changes and if any is found update new data only. Whether it's adding or removing.

skoruppa:
> Looks good but add a check every few mins to check for new changes and if any is found update new data only. Whether it's adding or removing.

I don't think this is needed. Series Watched Badges should barely change and 12h cache should be enough. Queering this every few minutes to only update checkmarks on home screen doesn't make any sense. I'm even considering increasing this cache into 24h

PR #1109, opened 2026-03-27
Feature/trailer youtube player fallback

halibiram:
## Summary

- add a YouTube iframe fallback for trailers when the current extraction flow hits rate limits or fails
- use `android-youtube-player` as the fallback player source: `https://github.com/PierfrancescoSoffritti/android-youtube-player`
- keep the existing trailer flow and UI unchanged, only switching to fallback when needed
- improve fallback startup sync by delaying unmute until the first playback progress signal arrives

## PR type

- Bug fix

## Why

- the current trailer system can fail after many trailer plays because YouTube-backed extraction starts hitting rate limits
- I have seen this problem mentioned many times in Discord, so this change addresses a recurring playback complaint from users
- this uses `android-youtube-player` as a safer fallback path for affected YouTube trailers
- the goal is to recover playback without changing the normal trailer experience

## Policy check

- [x] This PR is not cosmetic-only, unless it is a translation PR.
- [x] This PR does not add a new major feature without prior approval.
- [x] This PR is small in scope and focused on one problem.
- [ ] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

- passed `:app:testDebugUnitTest --tests "com.nuvio.tv.data.trailer.TrailerServiceYouTubeSessionCacheTest"`
- passed `:app:compileDebugKotlin`
- passed `:app:assembleDebug`
- manually tested trailer playback flow and confirmed the fallback still opens without changing the existing UI path
- manually tested the fallback startup behavior to verify audio/video startup feels more aligned after the delayed unmute change

## Screenshots / Video (UI changes only)

- None

## Breaking changes

- None

## Linked issues

- context only: this issue has been discussed repeatedly in Discord by users hitting trailer playback/rate-limit problems

PR #1186, opened 2026-03-31
Continue PR #1109 with selectable and iframe-only trailer playback

halibiram:
## Summary

- continue the trailer fallback work from #1109 by turning the YouTube iframe path into a selectable trailer source instead of a fallback-only path in the normal beta/app flow
- add an iframe-only Android flavor and release flow so a separate build can ship with iframe trailers only, without the trailer source selector in settings, and without the real in-app extractor implementation included
- preserve the existing standard debug/release workflow by keeping legacy APK names and output paths available for current scripts and automation
- merge the branch on top of latest `origin/dev` and keep the upstream trailer cache and home presentation changes compatible with this continuation work
- refined MetaDetails trailer seek behavior for iframe playback so long seek stays in sync with the custom seek overlay and commits only the final target position
- updated MetaDetails trailer button visibility so iframe trailers show the button only after iframe playback is actually ready, while the in-app trailer path keeps the existing button behavior

## PR type

## Why

- #1109 introduced the YouTube iframe fallback path for trailer playback, but users still could not explicitly choose that path when the in-app route was unstable for specific videos or environments
- we also need a dedicated iframe-only release track for cases where trailer playback should ship without the real in-app extraction implementation included in the APK
- in the normal beta/app flow, both methods stay available and the new settings option allows choosing between them; in the iframe-only build, that settings option is removed and only the iframe path is shipped
- verified MetaDetails iframe trailer seek stays synchronized with the custom seek overlay
- verified the MetaDetails trailer button appears only after iframe playback is ready, and the in-app tr [truncated]
- note: the current `values-pt-rBR/strings.xml` XML error fix is intentionally not included in this PR

## Policy check

## Testing

- built the normal app flow successfully with the regular build path
- ran the iframe-only release script successfully via `python scripts/release_beta_iframe.py 0.5.3-beta`
- verified the normal beta/app build exposes the trailer source setting and both trailer playback methods work there: the standard app/in-app trailer path works and the iframe trailer path works
- verified the iframe-only build uses only the iframe trailer path, does not expose the trailer source setting in the menu, and does not include the real in-app extractor implementation
- merged with latest `origin/dev` and resolved conflicts in `TrailerService` and `HomeViewModel`

- normal beta/app settings screen showing the new `Trailer Source` selector with both `In-App` and `YouTube IFrame` options available
- Screenshot_20260331_162137

## Breaking changes

- adds `standard` and `iframeOnly` product flavors, while preserving legacy `assembleRelease`, release output paths, and standard APK naming for existing tooling

## Linked issues

- continues #1109


skoruppa:
After testing:
- YouTube player from Details Screen works ✓
  - after seeking YouTube controls/channel title is visible. I think we can use [this](https://github.com/PierfrancescoSoffritti/android-youtube-player?tab=readme-ov-file#remove-annoying-views)
  - The trailer button is not visible right away - I think that with iframe we don't need to wait for `iframeTrailerPlaybackReady`, we should just show the button immediately and display a loading state in the player if needed
- Media Hero ✓
- Expanded Poster ✓
- Full Screen Poster ✓

Can we enforce quality? It seems the quality is tied to the player size - hero and poster players get worse quality than full screen. I saw in [documentation](https://github.com/PierfrancescoSoffritti/android-youtube-player?tab=readme-ov-file#change-video-quality) it should be possible 

The iframe also seems to load very slowly compared to regular YouTube browsing. I think the root cause is that every focus creates a new WebView and re-fetches `iframe_api` from scratch. We should keep a single singleton `YouTubePlayerView` alive for the whole app lifetime and reuse it - the IFrame API would load once at startup and video switches would just be a `loadVideo` call.

Maybe with iframe method we don't need to wait until poster is extended. With this method we shouldn't be afraid of rate limiting (I hope xD) so WebView can be prepared before poster expandes 

And although this is a lot of good work with splitting the release into two separate ones and making a script to prepare a "Google Play" version I assume this is too much for this PR ;) Leave only changes related to player, let people test it and only after that lets make decisions


I know we already picked one library - but we can compare it to https://github.com/IlyaPavlovskii/YouTubePlayer and see which one work for us better

halibiram:
You're right, creating separate builds was overkill. I'll just focus on the core problem now and see what I can do about the feedback. For the extractor, since it scans multiple CDNs, I think we can tweak it to automatically play the first 4K stream it finds. I'll also handle the quality settings for the other open-source provider. Lastly, I'll compare the new source you sent to see which one runs lighter.

skoruppa:
> You're right, creating separate builds was overkill. I'll just focus on the core problem now and see what I can do about the feedback. For the extractor, since it scans multiple CDNs, I think we can tweak it to automatically play the first 4K stream it finds. I'll also handle the quality settings for the other open-source provider. Lastly, I'll compare the new source you sent to see which one runs lighter.

Leave the extractor :D I think it works fine when not rate limited. The quality problem is only with the iframe one. So just focus on the iframe player ;)

halibiram:
> Leave the extractor :D I think it works fine when not rate limited. The quality problem is only with the iframe one. So just focus on the iframe player ;)

Alright, I'll keep my hands off before I accidentally nuke something. Sticking strictly to the iframe territory! 😂

halibiram:
closing this as it's unnecessary now.

PR #1190, opened 2026-03-31
fix crashes: Disable VideoEasy scrapers by default and require opt-in confirmation

### LINKED ISSUE #1187: [Bug]: VideoEasy can crash the whole app through QuickJS native failure during provider resolution
### Pre-flight checks

### App version / commit

All versions

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

All devices

### Android version

All versions

### Player mode

Internal player

### Area (tag)

Other

### Steps to reproduce

Open Spectacular Spider Man, with plugins enable and Videasy enable
Wait for app to crash.

### Expected behavior

App doesn't crash.

### Actual behavior

App crash for some series/movies with videasy.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

There is a crash path that is reproducibly triggered by the VideoEasy plugin.

This is not a normal plugin failure where a scraper simply returns no results. In this case, a provider failure inside VideoEasy can drive the QuickJS runtime into a native crash (SIGSEGV), which kills the entire app process.

That makes this more serious than a normal bad plugin response, because from the user perspective it looks like the whole app is unstable.

What is happening
The issue appears during plugin execution, not in the playback engine itself.

Observed behavior:

playback can already be running normally, or source resolution can still be in progress
VideoEasy continues executing provider logic in the background
when a provider path inside VideoEasy fails, QuickJS crashes natively
the app process dies
The important point is that the visible trigger is a provider failure inside VideoEasy, but the actual fatal failure is a native QuickJS crash while processing that async failure.

Why this matters
Telling users to manually disable VideoEasy is not an ideal solution.

Most users:

will not know which plugin is responsible
will not know how to identify it
will assume the app itself is broken
So even if VideoEasy is the trigger, the user-facing problem is still “the app crashes”, not “one plugin misbehaved”.

I investigated this in depth and narrowed it down to VideoEasy.

Reproduction pattern
The crash was reproducible with VideoEasy alone on the all devices.

That allowed to rule out “multiple plugins running at once” as the primary explanation.

What the logs showed
The native stack consistently pointed into QuickJS async job execution, in the path handling plugin promise failures.

In other words:

this is not a normal Kotlin/Java exception
this is not just playback failing
this is a native QuickJS crash triggered by a VideoEasy failure path
Attempts that did not solve it
Tried several mitigations inside the current architecture.

1. Handling timeout failures more defensively
Changing the runtime so timeout/cancel cases were converted into safer synthetic responses instead of raw transport failures.

Result:

not enough
the crash still happened
2. Killing VideoEasy after the first timeout
Then attempted to short-circuit further execution after the first timeout.

Result:

too late
the crash was happening in the failure path itself
3. Cancelling VideoEasy earlier with a fixed execution budget
Tried cancelling the plugin before the expected timeout window.

Result:

still not enough
QuickJS also crashed on cancellation/failure processing
4. Blocking a specific provider route (Astra / visioncine)
At one point this looked like the main trigger, so blocked that route.

Result:

not sufficient
the crash still reproduced through another VideoEasy provider path afterward
5. Sanitizing the scraper code to remove the suspected provider before execution
Removed the Astra / visioncine provider from the VideoEasy scraper code before running it.

Result:

still not sufficient
the crash moved to another provider path inside VideoEasy
Current conclusion
The evidence strongly suggests this is not just one bad endpoint.

The broader issue is:

VideoEasy can drive QuickJS into a native crash when certain async provider failure paths are hit
because QuickJS runs inside the main app process, that crash kills the entire app
So while VideoEasy is the trigger, the real architectural weakness is that a native plugin-runtime crash is not contained.

Practical mitigation
The most practical short-term mitigation is to disable VideoEasy or otherwise block its execution on affected configurations.

That would reduce user-facing crashes immediately.

However, this is still only a mitigation:

it does not solve the underlying architectural problem
it also puts too much responsibility on the user if exposed only as “disable this plugin”
Robust solution
The most robust solution would be to isolate the plugin runtime from the main app process.

How that would work
run QuickJS/plugin execution in a separate Android process
communicate with it through IPC
if QuickJS crashes, only the plugin process dies
the main app stays alive and can report “plugin failed” instead of crashing entirely
Advantages
native QuickJS crashes would no longer kill the whole app
plugin failures would be contained
user-facing stability would improve significantly
Disadvantages
much more architectural complexity
IPC/service lifecycle management
more serialization/deserialization code
more difficult debugging
some runtime overhead
So this is not a small fix, but it is the first option that looks truly robust against this class of native plugin-runtime failure.

Bottom line
This issue is currently tied to VideoEasy, but the user-visible problem is that a plugin-level failure can crash the whole app.

Short-term:

option 1 - Let it be, and tell users to disable VideoEasy 

option 2 - Ship VideoEasy disabled by default and show a warning when the user tries to enable it.

This practical mitigation would be to keep VideoEasy-based scrapers disabled by default and show a warning when the user tries to enable them.

This should happen at the scraper level, not at the repository level, since a single repository can contain multiple providers and only the VideoEasy-based one is the known problem here.

That would be much better than relying on users to discover the issue only after the app crashes. Most users will not know which provider is responsible, and they will assume the app as a whole is unstable.

A reasonable product-level mitigation would be:

install or refresh repositories normally
automatically keep VideoEasy-based scrapers disabled by default
when the user tries to enable one, show a clear warning such as:
“This provider is known to cause crashes on some content. Enable anyway?”
This would not solve the underlying QuickJS/process-isolation problem, but it would reduce confusion, improve communication, and make the risk explicit at the right moment.

Long-term:

option 3 - plugin runtime isolation is the robust fix, much more complex, with trade-offs.

Without containment, any future native QuickJS failure of this kind will continue to look like a full app crash to end users.

I think that option 2, is probably the best approach.

```shell

```

### Anything else? (optional)

_No response_

kernexshadow:
## Summary
This PR adds a safety mitigation for VideoEasy-based scrapers in the plugin flow.

Changes included:

VideoEasy-based scrapers are now installed as disabled by default.
Existing scraper enabled state is preserved on repository refresh (no forced re-enable/reset).
When a user tries to enable a VideoEasy-based scraper, the app now shows a confirmation warning dialog before enabling it.
This is intentionally a product-level safeguard, not a full runtime architecture rewrite.

## PR type
Bug fix

## Why
This is needed to reduce user-facing crashes tied to VideoEasy plugin execution paths and to make the risk explicit at the moment of enablement.

Disabling VideoEasy manually is not realistic for most users because they usually do not know why the app is crashing or even that a specific provider is causing crashes and will assume the app itself is unstable.

Notes on alternative approach (as discussed in #1187)
The more robust long-term direction is to run plugin/QuickJS execution in a separate process from the main app process.

Why this is more robust:

if QuickJS crashes natively, only the plugin process dies
the main app process stays alive
the app can report a plugin failure instead of hard-crashing

Why this is not included in this PR:

requires a significant architecture change (process boundary + IPC layer)
needs request/result marshalling and lifecycle handling for the plugin process
requires binder/process-death handling and restart/recovery logic
increases complexity in timeout/cancellation semantics across process boundaries
adds substantial QA/regression surface for all plugin execution flows
can introduce overhead/latency tradeoffs and more complex debugging/observability
Given those risks and scope, this PR applies a focused mitigation now and leaves process isolation as a future, larger change.

Context:

Bug request: #1187
Policy check
[X] This PR is not cosmetic-only, unless it is a translation PR.
[X] This PR does not add a new major feature without prior approval.
[X] This PR is small in scope and focused on one problem.
[] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

Installed/updated plugin repositories containing VideoEasy and confirmed those scrapers default to disabled.
Tried enabling a VideoEasy scraper and confirmed warning dialog appears.
Confirmed enabling proceeds only after explicit confirmation.
Confirmed non-VideoEasy scrapers are unaffected.

Videasy disable by default:

image

UI dialog:

image

## Breaking changes

None.

## Linked issues
Fixes #1187

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/plugin/PluginManager.kt
+++ b/app/src/main/java/com/nuvio/tv/core/plugin/PluginManager.kt
@@ -547,6 +547,13 @@ class PluginManager @Inject constructor(
                 
                 // Create scraper info
                 val scraperId = "$repoId:${info.id}"
+                val existingScraper = existingScrapers.firstOrNull { it.id == scraperId }
+                val defaultEnabled = info.enabled &&
+                    !PluginSafety.isVideoEasyScraper(
+                        scraperId = info.id,
+                        scraperName = info.name,
+                        filename = info.filename
+                    )
                 val scraper = ScraperInfo(
                     id = scraperId,
                     repositoryId = repoId,
@@ -555,7 +562,7 @@ class PluginManager @Inject constructor(
                     version = info.version,
                     filename = info.filename,
                     supportedTypes = info.supportedTypes,
-                    enabled = true,
+                    enabled = existingScraper?.enabled ?: defaultEnabled,
                     manifestEnabled = info.enabled,
                     logo = info.logo,
                     contentLanguage = info.contentLanguage ?: emptyList(),
--- a/app/src/main/java/com/nuvio/tv/core/plugin/PluginSafety.kt
+++ b/app/src/main/java/com/nuvio/tv/core/plugin/PluginSafety.kt
@@ -1,0 +1,13 @@
+package com.nuvio.tv.core.plugin
+
+internal object PluginSafety {
+    fun isVideoEasyScraper(
+        scraperId: String?,
+        scraperName: String? = null,
+        filename: String? = null
+    ): Boolean {
+        return listOf(scraperId, scraperName, filename).any { value ->
+            value?.contains("videasy", ignoreCase = true) == true
+        }
+    }
+}
--- a/app/src/main/java/com/nuvio/tv/ui/screens/plugin/PluginScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/plugin/PluginScreen.kt
@@ -282,6 +282,18 @@ fun PluginScreenContent(
             }
         }
     }
+
+    if (uiState.pendingScraperEnable != null) {
+        Popup(properties = PopupProperties(focusable = true)) {
+            uiState.pendingScraperEnable?.let { pending ->
+                ConfirmScraperEnableDialog(
+                    scraperName = pending.scraperName,
+                    onConfirm = { viewModel.onEvent(PluginUiEvent.ConfirmPendingScraperEnable) },
+                    onDismiss = { viewModel.onEvent(PluginUiEvent.DismissPendingScraperEnable) }
+                )
+            }
+        }
+    }
     }
 }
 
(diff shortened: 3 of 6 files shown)


ram130:
Looks good. Ready to merge. 

@tapframe

kernexshadow:
One of the many examples: https://discord.com/channels/1379902184207941732/1458552195904573513/threads/1488558157713379512

PR #1215, opened 2026-04-02
Add configurable app DNS routing and caching

halibiram:
## Summary

- Introduce app-level DNS routing with `AppDnsManager`, `AppDnsProvider`, and `buildWithAppDns()` so shared OkHttp clients, plugin fetches, subtitle hashing, player probing, trailer extraction, and image requests resolve hosts through the same runtime-configurable resolver.
- Add DNS settings persistence and UI wiring through `NetworkDnsSettingsDataStore`, `NetworkDnsInitializer`, and `NetworkSettingsViewModel`, including provider selection, IPv4-first ordering, and a new DNS cache toggle exposed from `NetworkSettingsScreen`.
- Replace the old `IPv4FirstDns` path with provider-aware plain DNS and DoH resolvers, evict pooled connections after DNS config changes, keep fallback-to-system behavior when custom resolution fails, and preserve legacy `network_dns_mode` values when reading stored settings.
- Add smarter in-memory DNS caching with provider-aware TTL handling, negative caching, stale-while-revalidate refreshes, and debug logging for cache hits, TTL source, and TTL remaining values.
- Align the network diagnostics screen with the active app resolver so the speed test uses the selected DNS provider and closes the DNS section when rerunning the test.
- ~~Migrate image loading from Coil 2 to Coil 3 by updating dependency coordinates, switching the application image loader to `SingletonImageLoader.Factory`, wiring Coil's OkHttp fetcher through app DNS, and updating Compose image call sites to the `coil3` API surface, including painter-state collection and `BitmapImage` handling where Coil 3 changed behavior. This also helps keep image loading aligned with the current stack and provides a small UI-side performance improvement on image-heavy surfaces.~~
- Revert the Coil 3 migration after it caused regressions, while preserving app DNS integration for image loading through the Coil 2-compatible OkHttp setup.
- Surface DNS TTL information in resolver logs for dnsjava and DoH-backed lookups so cache behavior is easier to inspect during debugging.

## PR type

## Why

- The app previously relied on a narrower DNS setup centered around IPv4-first behavior, while several network call sites created their own clients or used `HttpURLConnection`, which meant resolver behavior was inconsistent across playback, plugins, subtitles, trailer extraction, and image loading.
- This change centralizes DNS handling so a user-selected provider applies across app traffic, adds optional caching to reduce repeated lookups, and makes the diagnostics screen reflect the same active resolver path used by the rest of the app.
- ~~The Coil update is also included to modernize the image pipeline around the newer API surface and to capture a small but meaningful UI responsiveness improvement in image-heavy screens.~~
- Coil 3 introduced regressions in this branch, so the image stack was rolled back to Coil 2 while keeping the DNS integration in place.

## Policy check

## Testing

- Manually tested the DNS settings flow by switching providers in settings, toggling `IPv4 First`, toggling `DNS Cache`, and confirming the selected values persist after re-opening the screen.
- Smoke tested network-backed flows after the resolver changes by opening image-heavy screens, starting playback, exercising subtitle-related fetches, and verifying plugin / trailer requests still resolve successfully with the shared app DNS client.
- Verified that the speed test now runs through the selected app DNS provider instead of forcing `Dns.SYSTEM`, while keeping the UI focused on latency and download metrics.
- Verified debug logging for DNS cache hits, TTL source, TTL remaining values, and stale refresh behavior.
- ~~Verified the Coil 3 migration path by checking Compose image surfaces that use async painters and `AsyncImage`, including the updated painter-state observers used in screens that now consume Coil 3 state as a `Flow`, with a small perceived performance improvement on image-heavy UI surfaces.~~
- Verified the Coil rollback path by checking that image loading stays on Coil 2 while still using the shared app DNS-aware OkHttp client.
- Relevant debug logs observed during testing are expected to look like this:

```text
DNS lookup host=live.metahub.space provider=cloudflare_dns ipv4First=true cache=true cacheHit=false stale=false fallback=false elapsedMs=357 ttlMs=300000 ttlRemainingMs=300000 staleRemainingMs=75000 ttlSource=dnsjava addresses=[92.223.124.62, 2a03:90c0:41:2801::62]
DNS lookup host=live.metahub.space provider=cloudflare_dns ipv4First=true cache=true cacheHit=true stale=false fallback=false elapsedMs=0 ttlMs=300000 ttlRemainingMs=298421 staleRemainingMs=0 ttlSource=dnsjava addresses=[92.223.124.62]
DNS lookup host=image.tmdb.org provider=quad9_doh ipv4First=true cache=true cacheHit=true stale=false fallback=false elapsedMs=0 ttlMs=300000 ttlRemainingMs=297820 staleRemainingMs=0 ttlSource=doh addresses=[143.244.56.50, 2600:9000:26e2:9600:14:41bb:de80:93a1]
```

tv_screenshot_2026-04-03_0209

Screenshot_20260402_180506

## Breaking changes

- No explicit API, schema, or configuration break is intended. Existing installs keep safe defaults by mapping legacy DNS mode values, keeping `SYSTEM` as the fallback provider, leaving IPv4-first enabled by default, and defaulting the new DNS cache setting to enabled.
- Runtime behavior does change in one important way: app traffic can now follow the selected DNS provider consistently across more networking surfaces.
- ~~Image loading now uses Coil 3 internals instead of Coil 2.~~
- Image loading remains on Coil 2 for now, but still uses the shared app DNS-aware client.

## Linked issues

- Continuation of #699.


halibiram:
~~I had to upgrade to Coil 3 to ensure image loading also uses the app-level DNS configuration consistently.~~

skoruppa:
I will later test this to check if any regression was not introduced by new coil.

@tapframe you can leave feedback here instead of the previous PR. I think functionality is useful, but as the last PR was for long time not touched I want to be sure you also agree to have this in ;)

@halibiram I'm afraid of one thing - I think those settings should not be synced to our cloud. People who use same account on multiple devices may not necessarily want same dns config everywhere

halibiram:
> I will later test this to check if any regression was not introduced by new coil.
> 
> @tapframe you can leave feedback here instead of the previous PR. I think functionality is useful, but as the last PR was for long time not touched I want to be sure you also agree to have this in ;)
> 
> @halibiram I'm afraid of one thing - I think those settings should not be synced to our cloud. People who use same account on multiple devices may not necessarily want same dns config everywhere

Thanks for the review! I'm dropping this part.

halibiram:
> I will later test this to check if any regression was not introduced by new coil.
> 
> @tapframe you can leave feedback here instead of the previous PR. I think functionality is useful, but as the last PR was for long time not touched I want to be sure you also agree to have this in ;)
> 
> @halibiram I'm afraid of one thing - I think those settings should not be synced to our cloud. People who use same account on multiple devices may not necessarily want same dns config everywhere

I checked the sync path and couldn't find any place where DNS settings are uploaded. ProfileSettingsSyncService only syncs the explicit syncedFeatures list, and DNS is stored separately in NetworkDnsSettingsDataStore, outside that flow. I still excluded it from Android backup/device transfer as an extra safeguard so it stays device-only.

skoruppa:
`dnsjava` was built for full Java (desktop/server) and references classes that don't exist on Android, for example JNA for Windows API calls. These code paths are never reached on Android but R8 fails the build because it can't resolve them. To build I had to add these dontwarn rules:


```
-dontwarn com.sun.jna.**
-dontwarn javax.naming.**
-dontwarn lombok.Generated
-dontwarn sun.net.spi.nameservice.**
-dontwarn org.xbill.DNS.spi.**
-dontwarn org.xbill.DNS.config.**
```



After you changes you not possible to **Speed Test Results** when dns settings are open. 
The option description "Run Network Speed Diagnostic" can be changed ;)



But most important - coil upgrade introduced regression and all icons are now pixelated on my device with your changes (sidebar, playe controlls)

halibiram:
> `dnsjava` was built for full Java (desktop/server) and references classes that don't exist on Android, for example JNA for Windows API calls. These code paths are never reached on Android but R8 fails the build because it can't resolve them. To build I had to add these dontwarn rules:
> 
> ```
> -dontwarn com.sun.jna.**
> -dontwarn javax.naming.**
> -dontwarn lombok.Generated
> -dontwarn sun.net.spi.nameservice.**
> -dontwarn org.xbill.DNS.spi.**
> -dontwarn org.xbill.DNS.config.**
> ```
> 
> After you changes you not possible to **Speed Test Results** when dns settings are open. The option description "Run Network Speed Diagnostic" can be changed ;)
> 
> But most important - coil upgrade introduced regression and all icons are now pixelated on my device with your changes (sidebar, playe controlls)

Got it, thanks for the review! I'll revert the Coil upgrade to fix the pixelated icons and resolve the Speed Test navigation bug. I'll also rename the string. Really appreciate the help with the R8 rules for dnsjava!

halibiram:
> But most important - coil upgrade introduced regression and all icons are now pixelated on my device with your changes (sidebar, playe controlls)

~~Just a quick question: If we stop using Coil for the icons and render them using Canvas instead, the pixelation issue would be solved. Do you think Coil provides a significant performance benefit here, or should I just revert the Coil update rather than spending time fixing the icons~~

halibiram:
@skoruppa I've reverted Coil 3. I also added a logic to the Running Speed Test button that closes the DNS settings if they're open, so it's properly visible while running the speed test now. Also, I think the DNS system definitely improves the network speed. It's ready for merge!"

skoruppa:
Yes, now pixelated icons are gone. Maybe we need to slightly change the speed test results size now, it doesn't fit its place >< 

@tapframe what you think? Looks solid

halibiram:
> Yes, now pixelated icons are gone. Maybe we need to slightly change the speed test results size now, it doesn't fit its place >

For the speed test results overflowing, I'm a bit stuck on how to arrange the UI. Should we consider moving the DNS settings to a separate menu to free up space?

halibiram:
@tapframe I had this tested, and the system really works. I got feedback from a test user saying the trailer load time dropped from 6 seconds down to just 1 second. Could you please take a look at this whenever you have a moment?

skoruppa:
> For the speed test results overflowing, I'm a bit stuck on how to arrange the UI. Should we consider moving the DNS settings to a separate menu to free up space? 

I think we can just make font there a little smaller end it should be good :p

halibiram:
advanced-speedtest-tv

-  @skoruppa Reduced the gap between Settings and Results cards to 10dp to match the left menu, and made the Results card fill the remaining vertical space for better bottom alignment

halibiram:
I hadn't gotten prior approval for this work, so I closed the PR

Issue #1066, opened 2026-03-25
[Bug]: Continue watching up next


gecXMAN:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

0.5.2

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Android TV Philips

### Android version

12

### Player mode

Internal player

### Area (tag)

Watch Progress (resume, watched state, history)

### Steps to reproduce

1.I'm using nuvio sync, not trakt
2. episode is finished on TV app
3. series disappears from continue watching
4. on mobile app it's ok, app shows up next episode after pull from the cloud
5. on nuvio web is ok alos

It only happens on primary profile, I have other two and they are working fine, up next is shown for the same series and episodes.

### Expected behavior

Show up next episode

### Actual behavior

Mobile app works fine "up next" episode shows but doesn't show on TV app. If I finish an episode on the TV app the series just disappears from continue watching

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

it's the same for all the series, Shrinking,Outlander etc.

### Logs (required for crash reports)

```shell

```

ram130:
I have this issue too. Seems the app stops tracking progress. 

**Example**

I finished an episode. Now I'm on the Easy Slider episode. 

_Android TV doesn't update accordingly._ 

![image](https://github.com/user-attachments/assets/3b186444-2cd1-4473-88ca-9f9812ec09dd)

![image](https://github.com/user-attachments/assets/e79d38bc-979f-4833-baec-802186e8efc5)


Trakt has is marked correctly

![image](https://github.com/user-attachments/assets/30ecafcd-8016-456c-a6ed-59452ea44c88)
![image](https://github.com/user-attachments/assets/750dc244-1014-455f-8dba-f570447d0102)

And likewise Nuvio mobile has it correct. 

![image](https://github.com/user-attachments/assets/e2512884-e6a8-421f-83f6-966c2ab967a2)

![image](https://github.com/user-attachments/assets/bfc65cb1-5a25-4b84-9087-3c680641ede4)


Rebooting the app doesn't update with the data either. 

@skoruppa can you take a look.

gecXMAN:
[...]

PR #1218, opened 2026-04-02
Try to match CW to the Mobile Implementation

skoruppa:
## Summary

Align Continue Watching (CW) pipeline on TV with the mobile implementation to fix missing/disappearing items.

Changes:
- Raise `CW_MAX_RECENT_PROGRESS_ITEMS` from 150 -> 300 and `CW_MAX_NEXT_UP_LOOKUPS` from 15 -> 24 (matching mobile constants)
- ~~Add watched episodes skip in Next Up resolution - previously TV picked the very next episode after the seed, even if it was already watched; now it skips watched episodes like mobile does via `getWatchedShowEpisodes()`~~
- Add 7-day window for unaired next-season episodes - mobile shows upcoming season premieres within 7 days, TV was dropping them entirely
- Add 5-minute TTL on negative meta/next-up cache - previously a failed meta resolve (addon timeout) cached `null` forever, permanently dropping items until app restart; now retries after 5 minutes

## PR type

- Bug fix

## Why

Multiple users reported that CW on TV is missing items that appear correctly on mobile. After comparing both implementations, possible causes were: lower item caps cutting off entries, no watched-episode skipping causing wrong Next Up picks, no unaired season window, and permanent negative caching silently dropping items when addons were temporarily slow.

## Policy check

## Testing

[Test build](http://morigal.pl/upload/app-universal-benchmark-cw.apk) provided for users that reported issues with CW for confirmation if changes do anything for them

no visual changes, only data pipeline fixes.

## Breaking changes

nothing should break, although there is a chance CW might work worse, not better ><

## Linked issues

Maybe #1066

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/HomeViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/HomeViewModel.kt
@@ -135,8 +135,10 @@ class HomeViewModel @Inject constructor(
     internal var pendingExternalMetaPrefetchItemId: String? = null
     internal val prefetchedTmdbIds = Collections.synchronizedSet(mutableSetOf<String>())
     internal val cwMetaCache = Collections.synchronizedMap(mutableMapOf<String, Meta?>())
+    internal val cwMetaNegativeCacheTimestamps = Collections.synchronizedMap(mutableMapOf<String, Long>())
     internal val cwTmdbIdCache = Collections.synchronizedMap(mutableMapOf<String, String?>())
     internal val cwNextUpResolutionCache = Collections.synchronizedMap(mutableMapOf<String, NextUpResolution?>())
+    internal val cwNextUpNegativeCacheTimestamps = Collections.synchronizedMap(mutableMapOf<String, Long>())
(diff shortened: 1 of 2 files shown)


mistersandman-64:
> * Add watched episodes skip in Next Up resolution - previously TV picked the very next episode after the seed, even if it was already watched; now it skips watched episodes like mobile does via `getWatchedShowEpisodes()`

Would this cause the Next Up episode to be incorrect for users rewatching a series? For instance - user watches season 1 then goes back and rewatches S1E1. The Next Up episode should then be S1EP2 instead of S2EP1. I know this was an issue a few versions ago.

skoruppa:
> > * Add watched episodes skip in Next Up resolution - previously TV picked the very next episode after the seed, even if it was already watched; now it skips watched episodes like mobile does via `getWatchedShowEpisodes()`
> 
> Would this cause the Next Up episode to be incorrect for users rewatching a series? For instance - user watches season 1 then goes back and rewatches S1E1. The Next Up episode should then be S1EP2 instead of S2EP1. I know this was an issue a few versions ago.

I do agree, rewatching is important so I can't implement it. Updated. New build replaced the old one

PR #1219, opened 2026-04-03
P2p support

### LINKED ISSUE #1220: [Feature]: Add p2p support
### Area (tag)

Playback

### Problem statement

Currently no support for p2p content.

### Proposed solution

Giving the possibility to play p2p content.

_No response_

_No response_

CrissZollo:
## Summary

Add P2P streaming support, piece prioritization, and buffered playback.

## PR type

## Why

High demand of implementation and being more competitive on the market

## Policy check

## Testing

- Tested selecting p2p streams from addons that provide them. metadata resolves, buffering progress shows, playback starts after 30s buffer
 - Tested seeking to unbuffered regions, player stays visible, buffering spinner shows torrent stats (peers/seeds/speed/progress), playback resumes once pieces arrive
 - Tested seeking back to already-watched sections, instant playback from cached pieces
 - Tested episode switching with torrent streams for series content
 - Tested player exit and re-entry, cached pieces persist and are reused
 - Tested retry on error lightweight re-prepare without loading overlay for torrent streams

## Breaking changes

No breaking changes

## Linked issues

Fixes #1220


CrissZollo:
@skoruppa Would you be able to have a look at this PR if you feel it is solid enough let me know.

We can then merge it and take with the next update.

skoruppa:
Yup, it works, and there is no regression with web based player. I think we can merge it 

This is going to be a big release ;)

skoruppa (approved):

tapframe (commented):
I do not have the required hardware to test this on, but I'm requesting gathering feeback from users who are willing to test before going beta.

Pentaphon:
@CrissZollo for people who don't want p2p support on Nuvio, we can disable p2p entirely if we want, right?

CrissZollo:
> @CrissZollo for people who don't want p2p support on Nuvio, we can disable p2p entirely if we want, right?

No need, it is based on the users added addons and config there.

Wekmed:
Files larger than 1.5GB usually don't open and are much slower compared to Stremio.

kernexshadow:
> > @CrissZollo for people who don't want p2p support on Nuvio, we can disable p2p entirely if we want, right?
> 
> No need, it is based on the users added addons and config there.

That's true, it's not needed. But maybe having the ability disable p2p on a global setting, might give nuvio an advantage over Stremio.

Sometimes people complain that they receive an email from their ISP, because of p2p.  They don't know how correctly configure to exclude uncached content, even if they thought they were doing the correct thing. This could give an extra peace of mind for the less tech savy.

jeromedevine:
I tried to open a 3.5Gb file (as a third attempt after getting stuck several minutes into Buffering in the previous 2 attempts on a 7Gb file and on this same file). It also kept at the 'Buffering' stage on this file. After about 5 to 7 minutes buffering it finally started the movie, and scrolling through the whole movie was instant like playing a local copy. I think it buffered the entire movie file into the cache before starting playing the movie. I checked the app's cache and it was at about 7Gb. I had to clear the app's cache as I did not want the Android box' strorage to fill up with torrent cache. I have extended storage and the app is installed on this extended storage ssd disk. Is there no limit to the buffering percentage before the movie starts streaming/playing? And does it then delete the cached buffered file after the streaming is closed?

CrissZollo:
> > > @CrissZollo for people who don't want p2p support on Nuvio, we can disable p2p entirely if we want, right?
> > 
> > 
> > No need, it is based on the users added addons and config there.
> 
> That's true, it's not needed. But maybe having the ability disable p2p on a global setting, might give nuvio an advantage over Stremio.
> 
> Sometimes people complain that they receive an email from their ISP, because of p2p. They don't know how correctly configure to exclude uncached content, even if they thought they were doing the correct thing. This could give an extra peace of mind for the less tech savy.

I agree on that point. I think I will make it so it is off by default and when a users tries to start a file that needs it a warning message will pop up to prompt the user to enable it. This pop up and information will probably contain that this is activated on the users own risk and they confirm they legally own the content.

CrissZollo:
> I tried to open a 3.5Gb file (as a third attempt after getting stuck several minutes into Buffering in the previous 2 attempts on a 7Gb file and on this same file). It also kept at the 'Buffering' stage on this file. After about 5 to 7 minutes buffering it finally started the movie, and scrolling through the whole movie was instant like playing a local copy. I think it buffered the entire movie file into the cache before starting playing the movie. I checked the app's cache and it was at about 7Gb. I had to clear the app's cache as I did not want the Android box' strorage to fill up with torrent cache. I have extended storage and the app is installed on this extended storage ssd disk. Is there no limit to the buffering percentage before the movie starts streaming/playing? And does it then delete the cached buffered file after the streaming is closed?

I will have to take a look at that. It should only buffer the first 30 seconds and not the whole file. Then it should also buffer 30 seconds from where the selected time on the timeline.

jeromedevine:
When it is saying 'Buffering...' can you include besides it a value in MB of the filesize being downloaded as a buffer file, that is updated as the buffer is downloaded? Like for example Buffering... 50MB. I think it would help the user understand whether the connection's buffering has become stuck especially if there were few seeds. Stremio has this information visible (see screenshot) and I think it helps.

Screenshot_20260404-175319

LumeraD3v:
## Suggestion: TorrServer as an alternative P2P backend for Android TV

Hey! I've been looking at the `p2pSupport` branch and the jlibtorrent integration is really well thought out, the piece prioritization strategy with the sliding 30s window, resume data persistence, and the NanoHTTPD local server approach all look good.

I work on [Lumera](https://github.com/LumeraD3v/Lumera), another Stremio-compatible Android TV app, and I went through the same problem space. I wanted to share my experience because I ended up going with a different approach that solved some pain points I hit on actual TV hardware.

## What I found with in-process libtorrent on Android TV

Low-end Android TV boxes (2GB RAM, quad-core ARM) struggle with:

- **Memory pressure**: a full libtorrent session with DHT, peer management, and piece hashing competing with ExoPlayer for RAM causes OOM kills on cheaper devices
- **Storage consumption**: the current implementation uses libtorrent's default sparse storage mode, which creates the file at full size on disk. A 10GB movie immediately claims 10GB on the filesystem, even before pieces are written. On TV boxes with 8-16GB total storage, that alone can fill the device. And even if you switched to a truly incremental allocation where only downloaded pieces consume space, it's still a fundamental problem, pieces accumulate as playback progresses, so by the end of a 2-hour movie you've consumed most of the file size anyway. I actually tried implementing a storage cap where playback would jump to a RAM buffer while old pieces were cleaned up on disk, then fall back to storage, it was impractical and full of edge case bugs. Cleanup on playback exit works, but during playback there's no great solution when libtorrent is writing pieces to disk.
- **Lifecycle complexity**: managing the session, resume data, cache eviction, and the HTTP server is a lot of surface area for bugs

## The TorrServer approach

I use [TorrServer](https://github.com/YouROK/TorrServer) as an external binary. The entire integration is ~4 files and around 400 lines of Kotlin:

1. **`TorrServerEngine`**: manages the binary lifecycle. The `libtorrserver.so` binary is bundled in `jniLibs/`, launched via `ProcessBuilder`, and health-checked via `/echo`. Start takes ~2 seconds.

2. **`TorrServerApi`**: thin REST client. Three endpoints do everything:
   - `POST /torrents` with `action: "add"` + magnet link
   - `POST /torrents` with `action: "get"` for file list and stats
   - `GET /stream?link={magnet}&index={fileIdx}&play`  this is what ExoPlayer points at directly

3. **`TorrentService`**: foreground service that orchestrates startup, file resolution (by filename → by index → largest video fallback), and progress polling.

4. **`TorrentProgress`** simple data class for UI updates (status, speed, peers, seeds).

What makes it better: **ExoPlayer just plays a localhost HTTP URL**. TorrServer handles all the buffering, piece prioritization, and caching internally with a memory-mapped ring buffer, no files growing on disk as playback progresses.

## Why this works better on TV hardware

- **~5MB preload** before playback starts (vs 30 seconds of pieces)
- **RAM-based cache with a fixed ceiling**: TorrServer uses a ring buffer in memory that wraps around, so a 3-hour 4K movie uses the same amount of storage as a 20-minute episode. No disk space consumed during playback at all.
- **Process isolation**: if the torrent engine crashes, the app doesn't. OOM killer targets the server process first, not the UI
- **Tiny integration surface**:no NanoHTTPD, no piece priority management, no resume data handling in app code. 
- **No native library bloat**: single binary per arch in jniLibs, no libtorrent4j + 4 arch-specific `.so` files


Happy to share more details on my implementation or help out if you're interested. The Lumera repo is public if you want to look at the code directly: the p2p stuff lives in `app/src/main/java/com/lumera/app/data/torrent/`.

CrissZollo:
> ## Suggestion: TorrServer as an alternative P2P backend for Android TV
> 
> Hey! I've been looking at the `p2pSupport` branch and the jlibtorrent integration is really well thought out, the piece prioritization strategy with the sliding 30s window, resume data persistence, and the NanoHTTPD local server approach all look good.
> 
> I work on [Lumera](https://github.com/LumeraD3v/Lumera), another Stremio-compatible Android TV app, and I went through the same problem space. I wanted to share my experience because I ended up going with a different approach that solved some pain points I hit on actual TV hardware.
> ## What I found with in-process libtorrent on Android TV
> 
> Low-end Android TV boxes (2GB RAM, quad-core ARM) struggle with:
> 
>     * **Memory pressure**: a full libtorrent session with DHT, peer management, and piece hashing competing with ExoPlayer for RAM causes OOM kills on cheaper devices
> 
>     * **Storage consumption**: the current implementation uses libtorrent's default sparse storage mode, which creates the file at full size on disk. A 10GB movie immediately claims 10GB on the filesystem, even before pieces are written. On TV boxes with 8-16GB total storage, that alone can fill the device. And even if you switched to a truly incremental allocation where only downloaded pieces consume space, it's still a fundamental problem, pieces accumulate as playback progresses, so by the end of a 2-hour movie you've consumed most of the file size anyway. I actually tried implementing a storage cap where playback would jump to a RAM buffer while old pieces were cleaned up on disk, then fall back to storage, it was impractical and full of edge case bugs. Cleanup on playback exit works, but during playback there's no great solution when libtorrent is writing pieces to disk.
> 
>     * **Lifecycle complexity**: managing the session, resume data, cache eviction, and the HTTP server is a lot of surface area for bugs
> 
> 
> ## The TorrServer approach
> 
> I use [TorrServer](https://github.com/YouROK/TorrServer) as an external binary. The entire integration is ~4 files and around 400 lines of Kotlin:
> 
>     1. **`TorrServerEngine`**: manages the binary lifecycle. The `libtorrserver.so` binary is bundled in `jniLibs/`, launched via `ProcessBuilder`, and health-checked via `/echo`. Start takes ~2 seconds.
> 
>     2. **`TorrServerApi`**: thin REST client. Three endpoints do everything:
>        
>        * `POST /torrents` with `action: "add"` + magnet link
>        * `POST /torrents` with `action: "get"` for file list and stats
>        * `GET /stream?link={magnet}&index={fileIdx}&play`  this is what ExoPlayer points at directly
> 
>     3. **`TorrentService`**: foreground service that orchestrates startup, file resolution (by filename → by index → largest video fallback), and progress polling.
> 
>     4. **`TorrentProgress`** simple data class for UI updates (status, speed, peers, seeds).
> 
> 
> What makes it better: **ExoPlayer just plays a localhost HTTP URL**. TorrServer handles all the buffering, piece prioritization, and caching internally with a memory-mapped ring buffer, no files growing on disk as playback progresses.
> ## Why this works better on TV hardware
> 
>     * **~5MB preload** before playback starts (vs 30 seconds of pieces)
> 
>     * **RAM-based cache with a fixed ceiling**: TorrServer uses a ring buffer in memory that wraps around, so a 3-hour 4K movie uses the same amount of storage as a 20-minute episode. No disk space consumed during playback at all.
> 
>     * **Process isolation**: if the torrent engine crashes, the app doesn't. OOM killer targets the server process first, not the UI
> 
>     * **Tiny integration surface**:no NanoHTTPD, no piece priority management, no resume data handling in app code.
> 
>     * **No native library bloat**: single binary per arch in jniLibs, no libtorrent4j + 4 arch-specific `.so` files
> 
> 
> Happy to share more details on my implementation or help out if you're interested. The Lumera repo is public if you want to look at the code directly: the p2p stuff lives in `app/src/main/java/com/lumera/app/data/torrent/`.

This is great feedback and I will definitively take a look at this! I think the most of what you mention is great. Why reinvent the wheel.

CrissZollo:
@LumeraD3v see if the new version feels better now.

LumeraD3v:
> @LumeraD3v see if the new version feels better now.

I just digged into the code, the implementation is clean and the architecture (Binary → Api → Service with StateFlow) is solid. 
And here's what I found when I was going through the code:

---

### 1. File resolution needs filename matching (high priority)

`resolveFileIndex` currently has two strategies: match by `requestedIdx + 1`, or fall back to the largest video file. This breaks for season packs in two ways:

**a) Missing filename strategy.** Some addons provide `behaviorHints.filename` but no `fileIdx`. In that case, the code skips straight to "largest video file" and plays the wrong episode. You already carry `filename` through `StreamBehaviorHints` and `PlayerNavigationArgs` , it just needs to be passed into `resolveFileIndex` and matched against the file paths from `getTorrentStats()`.

**b) TorrServer sorts files alphabetically before assigning IDs.** The Go source does:
```go
sort.Slice(files, func(i, j int) bool {
    return utils2.CompareStrings(files[i].Path(), files[j].Path())
})
// then assigns:
st.FileStats = append(st.FileStats, &state.TorrentFileStat{
    Id: i + 1, // in web id 0 is undefined
})
```

So Stremio's `fileIdx` (based on original torrent order) can disagree with TorrServer's ID (based on alphabetical order). Example: a torrent with `S01E01.mkv`, `S01E10.mkv`, `S01E02.mkv`,  Stremio says `fileIdx: 1` meaning `S01E10.mkv`, but TorrServer assigns ID 2 to `S01E02.mkv` (sorted alphabetically), so `fileIdx + 1 = 2` plays the wrong episode.

Filename matching as the first strategy makes this very reliable. In Lumera I use this order:
1. Match by `behaviorHints.filename` against TorrServer's file paths
2. Match by ID offset (`fileIdx + 1`)
3. Match by positional index in the file list
4. Fallback: largest video file

---

### 2. Metadata race condition (high priority)

`resolveFileIndex` calls `getTorrentStats()` once immediately after `addTorrent()`. If TorrServer hasn't fetched the torrent metadata yet (very common with magnet links), `files` is empty and the code falls through to `return requestedIdx?.plus(1) ?: 1` which is a blind guess with no validation.

This needs a polling loop. In Lumera I poll `getFileList()` every second with a 15-second deadline:
```kotlin
val deadline = System.currentTimeMillis() + 15_000
while (System.currentTimeMillis() < deadline) {
    val files = api.getFileList(magnetLink)
    if (files.isNotEmpty()) // proceed with resolution
    delay(1000)
}
```

---

### 3. Addon trackers are dropped (medium priority)

`TorrentService.startStream()` accepts a `trackers` parameter but it's always called with the default empty list. The Stremio addon JSON often includes trackers in the `sources` field (prefixed with `"tracker:"`), but these never reach `buildMagnetUri()`.

The `Stream` domain model needs to carry the `sources` list through from the DTO, and the player integration needs to pass them to `startStream()`. Without addon-specific trackers, some torrents will have worse peer discovery or fail to start entirely.

---

### 4. No foreground service (medium priority)

`TorrentService` is a `@Singleton` class, not an Android Service. On low-memory TV boxes (2GB RAM), the system can kill the app process while the user is watching. A foreground service with a low priority notification prevents this.

---

### 5. Async cleanup on stop (low priority)

`stopStream()` launches a fire-and-forget `scope.launch` to drop the torrent:
```kotlin
currentHash?.let { hash ->
    scope.launch {
        try { api.dropTorrent(hash) } catch (e: Exception) { ... }
    }
}
```

If this runs during process death (which is exactly when cleanup matters most), the coroutine may never execute. The torrent stays active in TorrServer until the next restart. Consider making the drop synchronous, `runBlocking` in a cleanup path is fine since it's a single quick HTTP call.

---

### 6. Missing `save_to_db: false` (low priority)

The `addTorrent` request doesn't include `"save_to_db": false`. TorrServer's default may persist every added torrent to its internal database, so the config directory accumulates stale entries across sessions.

---

### 7. Missing `setExecutable()` (low priority)

Some Android devices/ROMs don't mark native libraries as executable by default. Calling `binaryFile.setExecutable(true)` before launching the process prevents silent failures on those devices.

---

### 8. Health check polling interval (minor)

The current 1-second polling interval for startup health checks means up to 1 second of wasted time after TorrServer is ready. A 200ms interval makes startup feel snappier.

---

Overall the foundation is solid, the StateFlow-based state management is cleaner than what I have in Lumera, and the sealed class approach for TorrentState is nice. The file resolution and metadata race condition are the two things I'd prioritize fixing since they directly cause wrong episode playback and startup failures.

Happy to answer questions or share more details on any of these.

mobo4368-netizen:
Hi how can i test this?

Wekmed:
> Hi how can i test this?

Income will be generated soon it's working quite stably we just need to wait.

skoruppa:
[new build](http://morigal.pl/upload/app-universal-benchmark-p2p.apk) for testing

jeromedevine:
> [new build](http://morigal.pl/upload/app-universal-benchmark-p2p.apk) for testing

First of all thank-you for your time and effort in creating this. It is greatly appreciated. The P2P component works really well. It is very fast in loading when there are seeds. The information presented when it is loading is helpful with the MB buffered. 
Two things that I noticed: When the stream is loaded and the stream is playing, if the user seeks forwards/backwards or the connection hickups, the MB buffered entry goes missing and is no longer shown.

I also noticed a bug when the app is closed (not put into the background, just completely closed) and then the user attempts to continue watching the same movie with a P2P past stream (automatically loaded). It attempts to buffer, and then it gives out the error as per the screenshot that I am enclosing.

Wekmed:
p2p and cloudstream. have any test build?

skoruppa:
> Remove 'reuse last link' for p2p links

Maybe instead of disabling it, we could save magnet in cache and reuse it when someone would like to continue with the same source?

CrissZollo:
@tapframe @skoruppa 

This is in my opinion now done and is ready to be merged.

skoruppa:
> @tapframe @skoruppa
> 
> This is in my opinion now done and is ready to be merged.

Yup, I agree. It works great 

@tapframe ready for merging 

//EDIT - binge group seems to not work

tapframe (approved):

Issue #1172, opened 2026-03-30
[Bug]: Trailer/Preview not auto-playing after 7 seconds on home screen (v0.5.3)


laphp1k-hash:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [ ] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

0.5.3

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Tcl c 835 

### Android version

11

### Player mode

Internal player

### Area (tag)

Playback (start/stop/buffering)

### Steps to reproduce

Open the NuvioTV app
Go to the home screen
Select any movie or TV show
Stay focused on the item for about 7 seconds without pressing anything
Observe that no trailer starts playing

### Expected behavior

A trailer/preview should automatically start playing after hovering on a movie or TV show for around 7 seconds.

### Actual behavior

The trailer/preview does not play or appear at all after waiting, even after more than 7 seconds. This issue started after updating to version 0.5.3-

### Frequency

Always

### Did this work before?

Not sure

### Logs (required for crash reports)

```shell

```

Shunsui7:
Can you share the settings you have under Layout > Focused Poster?

SkyDiver7199:
I have the same issue in 0.5.6.

The autoplay trailer setting under focused poster was enabled.

Also, the trailer button no longer shows... it's like it couldn't find the trailer anymore hence it can't autoplay? 

Used to work before this

donjengibre:
This happens to me too, the trailer auto-playing is very hit and miss. Sometimes it works, sometimes it doesn’t. I’ve tried toggling VPN, different trailer settings and can’t determine the reason.

Shunsui7:
They have been playing.. IMO unless someone from the team can comment.. Sometimes a fresh install works. It was part of the requirement for 0.5 (I think) second is purely settings.

SkyDiver7199:
Still not working in 0.6.0 beta

Issue #807, opened 2026-03-13
[Bug]: Trailer Shows Black Screen When Autoplaying From Detail Screen from Search Results


PBslayer:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

Beta 0.4.18

### Install method

GitHub Release APK

### Platform

Fire TV

### Device model

Fire TV Stick 4K Max

### Android version

Fire OS 8

### Player mode

Internal player

### Area (tag)

UI / Layout

### Steps to reproduce

1. Open the app.
2. Make Sure Auto-Play Trailers on detail screen is enabled.
3. Search for a movie or show using the search feature.
4. Select the movie or show from the search results.
5. Observe the trailer autoplay behavior on the details page.

### Expected behavior

The trailer should autoplay normally, just like it does when the movie or show is opened from the catalog. When selecting a movie/show from the catalog, the trailer plays as expected. However, when a movie/show is selected from search results, the trailer fails to autoplay and instead shows a black screen.

### Actual behavior

The trailer attempts to autoplay but only displays a black screen from a movie/show selected from search results.

### Frequency

Always

### Did this work before?

Not sure

### Logs (required for crash reports)

```shell

```

y-a13k:
I've also experienced this. Sometimes, the trailer doesn't work and shows a black screen or it doesn't play on the Home Screen on hero media or expanded card.

PR #1222, opened 2026-04-03
fix: cache YouTube visitor_data + API key to prevent 429 rate limiting

halibiram:
## Summary

Cache YouTube `visitor_data` and `API key` from the watch page instead of fetching them on every trailer extraction. The watch page is now fetched only once (on first use) and cached for 3 hours. On `LOGIN_REQUIRED` errors, the cache is invalidated and both credentials are refreshed together before retrying.

## PR type

- Bug fix

## Why

Heavy trailer browsing triggers YouTube's rate limiting (HTTP 429) on the watch page endpoint (`/watch?v=`), which causes all subsequent trailer extractions to fail. 

Investigation revealed:
- The **watch page** is the bottleneck — it gets rate-limited under load, while the Player API (`/youtubei/v1/player`) is much more resilient
- The **API key** (`INNERTUBE_API_KEY`) is not actually validated by YouTube — requests work without it
- The **`visitor_data`** is the critical credential — without it, most videos return `LOGIN_REQUIRED`
- A single `visitor_data` works across all videos and doesn't change between requests

By caching `visitor_data` + API key and only refreshing on errors, watch page requests are reduced by ~99%, effectively eliminating the 429 rate-limit problem.

## Policy check

## Testing

**Manual testing:**
1. Installed debug build on Android TV emulator
2. Browsed home screen trailers — trailers load on first focus with only 1 watch page request
3. Verified subsequent trailer loads use cached config (no additional watch page fetches in logcat)

**Automated stress testing (Python scripts):**
1. Sent 100 concurrent requests per wave × 6 waves (1200+ HTTP calls) to reproduce 429
2. Confirmed watch page gets 429 while Player API stays healthy
3. Verified `visitor_data` works across all videos with a single fetch
4. Confirmed API key is not required (YouTube ignores it)
5. Tested `LOGIN_REQUIRED` detection triggers cache refresh

N/A — no UI changes.

## Breaking changes

None. The `InAppYouTubeExtractor` public API is unchanged. The only behavioral change is internal: watch page is now cached instead of fetched per-extraction.

## Linked issues

Fixes YouTube 403/429 rate limiting when browsing trailers.
#1172  #807

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/trailer/InAppYouTubeExtractor.kt
+++ b/app/src/main/java/com/nuvio/tv/data/trailer/InAppYouTubeExtractor.kt
@@ -9,6 +9,8 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import kotlinx.coroutines.withTimeout
@@ -18,6 +20,7 @@ import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
 import java.net.URL
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicReference
 import javax.inject.Inject
 import javax.inject.Singleton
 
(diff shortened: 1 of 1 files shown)


halibiram:
Finally sorted out the trailer bug! It turns out that firing visitor and API requests continuously was triggering YouTube's login block. I optimized the flow to only make these requests on-demand.

halibiram:
@skoruppa  I don't think the trailer will cause any more issues. Since that's sorted, can we finally add the 'trailers and behind the scenes' row from the mobile app to the details page?

skoruppa:
Looks good. Lets hope this will finally resolve our trailer problems :)

PR #1237, opened 2026-04-04
Improve playback recovery, mime detection and aspect ratio handling

halibiram:
## Summary

- Reworks source-aware playback recovery end to end. `Stream` now carries `sources`, player navigation forwards `sourceUrls`, the stream link cache persists them, and media source creation plus mime detection now use `filename` and response header data. Playback failures caused by mime type parsing were resolved by improving mime detection, and tests were added for stream URL handling and mime inference.
- Adds a silent PCM fallback for ExoPlayer audio track initialization failures and a one-shot DV7 -> HEVC fallback flow for decoder initialization failures. These fallback paths are designed to kick in even when decoder priority is set to `Prefer device decoder`, so the player can try to recover through the loading recovery flow instead of failing straight to the error screen.
- Connects these recovery paths to runtime playback. This brings together automatic next-source fallback, mime detection based on response header data, and updated player loading messages; in practice it provides recovery paths for three different classes of failures: mime type parsing errors, audio playback failures, and DV7 decoder failures.
- Fixes an ExoPlayer aspect ratio issue where video size changes could reset or incorrectly override the user's selected aspect ratio. The selected resize mode is now reapplied after video size changes and first-frame rendering so the chosen aspect ratio remains active during playback.
- Replaces the older recovery toggles with a unified `SeamlessPlaybackMode` model in settings, covering stream-only, player-only, and stream-then-player recovery flows.
- Fixes same-addon fallback expansion by carrying full per-stream metadata into `RecoverySourceCandidate`, prefetching source pools for recovery, keeping fallback order scoped to the selected addon's streams, and preserving recovery overlay text during re-init.
- Adds and updates regression coverage for stream parsing, media source handling, autoplay selector compatibility, and same-addon recovery candidate ordering.

## PR type

- Approved larger change 

## Why

- Playback recovery had drifted across several code paths: stream fallback only knew about bare URLs, engine failover lived separately from stream failover, and some startup failures retried the same broken source instead of continuing through valid candidates.
- These commits make recovery source-aware, keep it scoped to the selected addon, preserve stream metadata while switching sources, and unify how the player decides between retrying, switching source, and switching engine.
- They also improve user feedback during recovery with explicit loading states instead of dropping back to generic buffering text too early.

## Policy check

## Testing

- Added `app/src/test/java/com/nuvio/tv/domain/model/StreamTest.kt` to verify `getStreamUrls()`, `getStreamUrl()`, and `isExternal()` behavior for streams with multiple sources.
- Added `app/src/test/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactoryTest.kt` to verify mime inference from `Content-Type`, `Content-Disposition`, query parameters, and redirected file types.
- Added `app/src/test/java/com/nuvio/tv/ui/screens/player/RecoverySourceCandidateTest.kt` to verify same-addon fallback ordering and metadata carry-over.
- Updated `app/src/test/java/com/nuvio/tv/core/player/StreamAutoPlaySelectorTest.kt` to align with the new `sources` field.
- Manually validated the debug build with decoder priority set to `Prefer device decoder`.
- Retested three scenarios that were not playing successfully in the main build: a source failing because of mime type parsing, a source failing with audio playback errors, and a source failing with a DV7 decoder error.
- The source affected by the mime type parsing issue played successfully after the mime detection logic was improved.
- PCM fallback was manually validated on the source that was failing with audio playback errors.
- DV7 -> HEVC fallback was manually validated on the source that was failing with the DV7 decoder error.
- Seamless mode transitions, same-addon source fallback ordering, and recovery overlay messages were manually validated through the relevant scenarios.

-
Screenshot_20260404_234142

Screenshot_20260404_234206

-

- Seamless playback **off**

https://github.com/user-attachments/assets/3cde95f1-7f37-4cab-817d-11e82e442550

- Seamless playback **Auto stream only**

https://github.com/user-attachments/assets/92847d49-8092-4c0a-87b2-b6e3a94ff957

- Seamless playback **Auto player only**

https://github.com/user-attachments/assets/63e38b94-410f-4a3b-9a03-41937eed2f52

## Breaking changes

- None.

## Linked issues

- None linked.


skoruppa:
I like those fixes and silent fallbacks for exo. But I’m not sure about the option you implemented.

Look at my fallback retry mechanism. It tries to reload player silently, after that more aggressively and only after that display an error.  I think we can have similar logic here - if there is an error that we can silently try to omit, lets do it once, if it fails - show an error or switch to mpv if user has an option for this enabled. Jumping through other sources seems like an overkill. 

If we want user to be aware that some kind of omitting is happening - we could maybe show an OSD message

Do you agree with me @tapframe ?

kernexshadow:
Love this PR.

This will make the experience much more smooth for the user, and less complaints in reddit and discord, my stream has this error, and don't open.

kernexshadow:
> I like those fixes and silent fallbacks for exo. But I’m not sure about the option you implemented.
> 
> Look at my fallback retry mechanism. It tries to reload player silently, after that more aggressively and only after that display an error. I think we can have similar logic here - if there is an error that we can silently try to omit, lets do it once, if it fails - show an error or switch to mpv if user has an option for this enabled. Jumping through other sources seems like an overkill.
> 
> If we want user to be aware that some kind of omitting is happening - we could maybe show an OSD message
> 
> Do you agree with me @tapframe ?

I prefer to have some info of whats happening. But I guess some people just prefer a seamless experience.

Why not an option?

skoruppa:
> > I like those fixes and silent fallbacks for exo. But I’m not sure about the option you implemented.
> > Look at my fallback retry mechanism. It tries to reload player silently, after that more aggressively and only after that display an error. I think we can have similar logic here - if there is an error that we can silently try to omit, lets do it once, if it fails - show an error or switch to mpv if user has an option for this enabled. Jumping through other sources seems like an overkill.
> > If we want user to be aware that some kind of omitting is happening - we could maybe show an OSD message
> > Do you agree with me @tapframe ?
> 
> I prefer to have some info of whats happening. But I guess some people just prefer a seamless experience? Why not an option?

You know what... after thinking more about this. Yeah, you are right. This makes experience more seamless. Those option are not bad. And "auto stream only" could be the default.

I didn't test this yet so let me ask - is the binge group updated when switching to new stream? Are new stream details saved if user has "reause last link" enabled and want to continue watching?

halibiram:
I see your point, but different users have different use cases. Some might actually prefer to try every single stream. How about I add a separate toggle for the fallback behavior? That way, users who don't want the silent retries can just bypass them and switch straight to MPV.

kernexshadow:
> You know what... after thinking more about this. Yeah, you are right. This makes experience more seamless. Those option are not bad. And "auto stream only" could be the default.

For some people mpv might be slow on their devices so player switch is not an option.

I agree that the auto stream only should be the default.

> 
> I didn't test this yet so let me ask - is the binge group updated when switching to new stream? Are new stream details saved if user has "reause last link" enabled and want to continue watching?

@halibiram you are the one who can answer this.

halibiram:
> 
> You know what... after thinking more about this. Yeah, you are right. This makes experience more seamless. Those option are not bad. And "auto stream only" could be the default.
> 
> I didn't test this yet so let me ask - is the binge group updated when switching to new stream? Are new stream details saved if user has "reause last link" enabled and want to continue watching?

 Looks like I might have broken this. I'll take a look

halibiram:
Fixed a bug preventing 'reuse last link' from functioning while 'Auto Stream Only' is enabled. Additionally, I simplified the text in the loading overlay.

halibiram:
@skoruppa If this auto stream feature is too much for now, I can open a separate PR with just the ExoPlayer improvements.

halibiram:
I'm closing this PR because moving the stream list from the navbar introduced a bug on certain devices, so a different approach is required. The issue is that if you open a 'continue watching' movie and dismiss the loading overlay early, the homescreen appears completely blank.

PR #1246, opened 2026-04-05
Improve ExoPlayer playback networking and recovery

halibiram:
## Summary

- ~~Improve ExoPlayer playback resilience by adding a dedicated playback networking layer with Cronet/OkHttp selection, richer MIME detection, and shared connection handling.~~
- Keep a dedicated ExoPlayer playback networking layer with stable OkHttp handling, richer MIME detection, shared connection handling, and header-aware data source creation.
- ~~Preserve and reuse alternate stream source URLs across mapping, navigation, caching, and runtime recovery so the player can retry more intelligently.~~
- ~~Tighten player recovery behavior by keeping the loading/recovery overlay active during retries and engine failover,~~ improve aspect scaling updates, keep decoder/audio fallbacks when `Prefer device decoders` is selected, and add a dedicated startup recovery path without changing the existing runtime retry flow.

## PR type

## Why

Some streams were failing or recovering poorly because playback setup depended on limited MIME inference, ~~a single source URL,~~ and fragile startup behavior. This change keeps the player more tolerant of inconsistent stream metadata and networking conditions, improves MIME detection from headers, filenames, query params, and probes, and adds a startup recovery path so TLS and some other startup playback failures before the first rendered frame can retry cleanly without changing the existing runtime recovery behavior.

When `Prefer device decoders` is selected, the fallback flow explicitly handles these cases:

- `ERROR_CODE_AUDIO_TRACK_INIT_FAILED (5001)`: retries with a PCM audio fallback when tunneling is off.
- `ERROR_CODE_DECODER_INIT_FAILED (4003)`: retries with DV7-to-HEVC decoder mapping for affected Dolby Vision playback.

## Policy check

## Testing

- I manually tested the updated ExoPlayer recovery and playback flows, including ~~source retry behavior,~~ startup recovery behavior, overlay visibility, and MIME detection improvements.
- Tester also validated the affected playback scenarios and confirmed the fallback flow works as expected with `Prefer device decoders` selected.
- Added `app/src/test/java/com/nuvio/tv/domain/model/StreamTest.kt` for stream URL prioritization, deduplication, and external URL behavior.
- Added `app/src/test/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactoryTest.kt` for MIME inference from headers, content disposition, paths, and query parameters.

None.

## Breaking changes

None.

## Linked issues

None.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerErrorRecovery.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerErrorRecovery.kt
@@ -12,6 +12,19 @@ import kotlinx.coroutines.launch
 
 private const val MAX_AUTO_RETRIES = 2
 private const val RETRY_DELAY_MS = 1_500L
+
+internal fun PlayerRuntimeController.showRecoveryOverlay() {
+    _uiState.update { state ->
+        state.copy(
+            error = null,
+            isBuffering = true,
+            showLoadingOverlay = true,
+            loadingMessage = context.getString(R.string.player_loading_buffering),
+            showPauseOverlay = false
+        )
+    }
+}
+
 /**
  * Determines whether the given [PlaybackException] is transient and worth retrying.
  *
@@ -112,9 +125,8 @@ private fun Throwable.findMostRelevantCauseMessage(): String? {
 /**
  * Attempts an automatic retry of the current stream, preserving the playback position.
  *
- * The player is fully torn down and re-initialised so that internal ExoPlayer state
- * (extractors, loaders, renderers) is clean - this is the most reliable way to recover
- * from the class of errors reported by users (corrupt parser state after pause/seek).
+ * The first retry re-prepares the current player, and the second retry fully rebuilds it,
+ * so recovery stays on the loading overlay until playback succeeds or finally fails.
  *
  * Returns `true` if a retry was scheduled, `false` if the error should be shown to the user.
  */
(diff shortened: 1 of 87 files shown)


app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerErrorRecovery.kt
```diff
@@ -76,24 +148,14 @@ internal fun PlayerRuntimeController.attemptAutoRetry(
...
     errorRetryJob = scope.launch {
-        _uiState.update {
-            it.copy(
-                error = null,
-                // Only show loading overlay on full teardown (second attempt).
-                showLoadingOverlay = if (isFirstAttempt) false else it.loadingOverlayEnabled,
-                showPauseOverlay = false
-            )
-        }
+        showRecoveryOverlay()
```
skoruppa: The first retry, without full teardown should be without showLoadingOverlay. The first retry I implemented was for simple network errors. This change seems to apply showLoadingOverlay here

halibiram: I reverted that part so the first retry no longer shows the loading overlay

ram130:
Interesting. Though may advise to await a little.

halibiram:
Cronet handles connection losses much better; it can recover as much as possible without affecting the player.

skoruppa:
> Cronet handles connection losses much better; it can recover as much as possible without affecting the player.

I didn't said my approach was the best :v it "works" and I don't want to break it :) but if there is a better way, we can explore it

halibiram:
> I didn't said my approach was the best :v it "works" and I don't want to break it :) but if there is a better way, we can explore it

So what you're saying is: 'If it works, don't touch it!

skoruppa:
I don't understand changes related to 
" Returns all playable URLs for the stream in priority order."

What URLs? Each stream is always one url. That did not change. So for what are those "multiple" sources?

I'm also not a fan of changing Okhttp to Cronet. I know it adds HTTP/3 (QUIC) support which can save ~100-200ms on initial connection to servers that support it. But for video streaming there should be no noticeable difference. . It also adds a dependency on Google Play Services, increases APK size, and adds complexity with the OkHttp fallback path. I also don't think it changes that much with connection loses. For mobile where connections changes between cellular and wifi this could help, but on TV we are not gaining anything. 

Handling ERROR_CODE_PARSING_CONTAINER_UNSUPPORTED, ERROR_CODE_PARSING_MANIFEST_UNSUPPORTED, ERROR_CODE_AUDIO_TRACK_INIT_FAILED  and ERROR_CODE_DECODER_INIT_FAILED is nice indeed but everything else seems not necessary

PR #1262, opened 2026-04-07
feat: add modern topbar option

SelfDL:
## Summary

Added a modern topbar option to layout settings.

## PR type

- Approved larger change 

## Why

Just an extra customization choice for users.

## Policy check

## Testing

Swapping between pages with the topbar enabled, blur on/off, using the default sidebar options to see if topbar changes their functionality. Everything seems to be working as intended with no bugs.

poc

## Breaking changes

None

## Linked issues

N/A


RSAK2105:
Wow! Nice. More or less similar to AppleTV top bar. Btw, Profile icon isn't missing at Top Left corner?

skoruppa:
Looks nice, but if we are adding another sidebar option I think we should have only one option with choices instead of 4 different switches. Something like in language/audio selection. We could have
- Classic
- Modern
- Moder Top 
and one switch to control the blur, visible only when Modern ones are selected. 

Also, don't commit changes to files in the .idea folde.

If you don't mind I can make some amendments

SelfDL:
> Looks nice, but if we are adding another sidebar option I think we should have only one option with choices instead of 4 different switches. Something like in language/audio selection. We could have
> 
> - Classic
> 
> - Modern
> 
> - Moder Top 
> 
> and one switch to control the blur, visible only when Modern ones are selected. 
> 
> 
> 
> Also, don't commit changes to files in the .idea folde.
> 
> 
> 
> If you don't mind I can make some amendments 

Yeah go for it! I probably could have just made the blur toggle shared now that you mention it, and sorry, first time using git feature and android studio this in depth. Didn't really know what to commit.

tsoltan:
So, what’s the reason for keeping just black empty space on the left if the toolbar would be on the top?
image

SelfDL:
> So, what’s the reason for keeping just black empty space on the left if the toolbar would be on the top? image

That specific movie hero is black on the edge, it's not just a blank space. As for the padding it's just cleaner with it pushed in. Would be ugly if everything spanned to the edge.

PR #1264, opened 2026-04-07
Bypass hard-strict subtitle parsing to reduce malformed subtitle startup delays

halibiram:
## Summary

- Disable subtitle parsing during extraction for the player media source path so malformed text subtitles do not slow down startup as early.
- Restrict the ASS subtitle parser compatibility path to actual SSA/ASS formats instead of attempting to participate in other text subtitle formats.
- Instantiate the text subtitle renderer with legacy decoding enabled so malformed subtitles can fall back to the more tolerant decoding path during playback setup.

## PR type

- Bug fix

## Why

Some streams include malformed text subtitles that trigger stricter parsing behavior during player setup. That can introduce subtitle-related startup delays before playback is ready. This change keeps subtitle parsing on a more tolerant path and limits the custom ASS handling to the formats it is intended for.

## Policy check

## Testing

- assembleDebug was successful. checked both video playback and subtitle functionality, and everything seems to be working perfectly

Not applicable. This PR does not change UI.

## Breaking changes

None expected. The change only adjusts subtitle parsing and text renderer setup for playback initialization.

## Linked issues

No linked issue.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerLibassCompat.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerLibassCompat.kt
@@ -43,7 +43,7 @@ internal fun ExoPlayer.Builder.buildWithAssSupportCompat(
     val mediaSourceFactory = DefaultMediaSourceFactory(
         dataSourceFactory,
         assExtractorsFactory
-    )
+    ).experimentalParseSubtitlesDuringExtraction(false)
     mediaSourceFactory.setSubtitleParserFactory(assSubtitleParserFactory)
 
     val player = this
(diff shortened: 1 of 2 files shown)


skoruppa:
I'm not sure if I understand changes here. Was anyone reporting such issues? Does this also help with subtitles from addons (for example .ass subtitles without .ass in the url)?

halibiram:
> I'm not sure if I understand changes here. Was anyone reporting such issues? Does this also help with subtitles from addons (for example .ass subtitles without .ass in the url)?

I haven't seen a specific report about this, but I noticed it myself: Media3's new strict subtitle extractor struggles with slightly malformed community subtitles and delays initial video startup. And yes, reverting to legacy decoding also fixes the issue with addon subtitles failing when the URL lacks an `.ass` extension, since it bypasses URL checks and reads the payload directly

halibiram:
I'm closing this as it makes no difference regarding the subtitles.

PR #1199, opened 2026-04-01
Support per-profile Trakt accounts with comments fallback

kernexshadow:
## Summary
This PR updates Trakt behavior for multi-profile setups:

Trakt auth is now profile-scoped (each profile can have its own Trakt account).
Trakt comments now support a fallback for secondary profiles:
If the active profile has no Trakt auth, comments are fetched using profile 1 auth (comments-only fallback).
Primary-only Trakt gating was removed where it blocked multi-profile behavior.
Trakt-related caches/state are reset on profile switch to avoid cross-profile data bleed.

## PR type
Small Impromevement

## Why

There was clear demand from users (Reddit and Discord) for proper multi-profile Trakt support.
The core issue was that Trakt was effectively behaving as a single-account setup in multi-profile usage, which did not match how profiles are expected to work.

This PR prioritizes that need by enabling Trakt per profile (separate Trakt auth/state per profile), so each profile can use its own Trakt account independently.

Additionally, to preserve usability for households already relying on shared behavior, this includes a comments-only fallback: if a secondary profile has no Trakt login, Trakt comments can still be read via profile 1 auth.
Progress/scrobble/library remain profile-scoped.

## Policy check
[X] This PR is not cosmetic-only, unless it is a translation PR.
[X] This PR does not add a new major feature without prior approval.
[X]  This PR is small in scope and focused on one problem.
[]  If this is a larger or directional change, I linked the issue where it was approved.

## Testing

Tested multi profiles with Trakt enabled sync.

## Screenshots / Video (UI changes only)
N/A

## Breaking changes

None. 

## Linked issues

On Discord and reddit

PR #1280, opened 2026-04-08
Fix DataStore corruption recovery

halibiram:
## Summary

- Add corruption handlers to the app's `Preferences DataStore` entry points so invalid protobuf state no longer crashes startup.
- Track corrupted synced profile stores and recover them by pulling the remote settings blob before an empty local state can be pushed back.
- Keep the change focused on DataStore crash recovery and settings preservation for already-synced profile preferences.
- Move watch progress storage off the single `watch_progress_preferences` JSON blob into sharded `Proto DataStore` files so corruption is isolated to a shard instead of taking down the whole store.
- Migrate legacy watch progress data into the new sharded store on first access while keeping the existing repository and sync-facing APIs unchanged.
- Clear profile-specific watch progress shard files alongside the rest of the profile-local DataStore cleanup.
- Prevent corrupted watch progress shards from being pushed back to Supabase until a recovery pull repopulates them, so temporary empty local state does not overwrite valid remote progress.

## PR type

- Bug fix

## Why

`Unable to parse preferences proto` was crashing the app when a `Preferences DataStore` file became unreadable. This change makes DataStore startup resilient to corruption and avoids overwriting remote profile settings with empty local data during recovery.

Watch progress was still stored as a single frequently-rewritten JSON blob inside `Preferences DataStore`, which left playback progress especially exposed to corruption. Moving it to sharded `Proto DataStore` files reduces the blast radius of a bad local file and keeps the watch progress recovery path compatible with the existing sync setup.

The follow-up problem was sync safety: once a corrupted shard was replaced with an empty local shard, the app could have treated that temporary state as valid and pushed it back to Supabase. The recovery flow now treats corrupted watch progress shards as invalid local state, suppresses outgoing watch progress pushes, and prefers a remote pull first so valid synced data can repopulate the local shard before sync resumes.

## Policy check

## Testing

- `./gradlew :app:compileDebugKotlin`
- I could not fully test the recovery flow end to end because I do not have access to the backend/API environment needed to validate the remote sync path.
- Logically, this should resolve the crash because unreadable `Preferences DataStore` files are now handled through corruption handlers instead of throwing during parse.
- For synced profile settings, the recovery path now marks corrupted stores, suppresses pushing the temporary empty state, and forces a remote pull first so valid server-side settings can repopulate the local store.
- Watch progress now migrates from the legacy preferences blob to sharded proto files without changing the existing repository or sync contract, so the runtime blast radius of corruption is smaller even before remote recovery kicks in.
- For watch progress sync, corrupted local shards are now treated like recovery-only state: push paths bail out, the repository requests a remote pull instead, and corruption markers are only cleared after remote data has been merged back in.

None. No UI changes.

## Breaking changes

None.

## Linked issues

- [Discord issues](https://discord.com/channels/1379902184207941732/1458552195904573513/threads/1491268578941468732)


skoruppa:
Although it is nice that the crash will no longer happen, we don't know why corrupted data happens - this will result in clearing user settings each time, instead of crashing

halibiram:
> Although it is nice that the crash will no longer happen, we don't know why corrupted data happens - this will result in clearing user settings each time, instead of crashing 

it's a bit random, but this is a temporary fix. it’s better to just clear the corrupted part and re-sync rather than forcing the user to clear the entire app cache. i suspect that trying to delete a profile from outside the flow might be causing this; triggering it through the android lifecycle instead would probably be safer.

halibiram:
@skoruppa  I took a closer look at the logs to analyze the issue in more detail, and the biggest suspect seems to be watch progress. In the current setup, while playback is active, watch progress was rewriting a large JSON blob into Preferences DataStore roughly every 10 seconds. The "Unable to parse preferences proto" / "Protocol message contained an invalid tag (zero)" error in the logs also points to that file becoming corrupted.

So the most likely problem is that the `watch_progress_preferences` store, which was being written very frequently during playback, became corrupted at some point, and then the app crashed because it could no longer parse that file.

To address this, I moved watch progress away from the single large JSON blob approach. It now uses sharded Proto DataStore files, so corruption is isolated to an individual shard instead of taking down the entire store. Legacy data is migrated into the new structure, the existing repository/sync contract stays intact, and profile cleanup now removes the shard files as well.

I also added sync-side protection: if a watch progress shard is marked as corrupted, the app will not push that temporary empty/local state. Instead, it tries a remote pull first; once valid remote data repopulates the local shards, the corruption markers are cleared and normal sync resumes. This also prevents broken local state from overwriting healthy remote progress data.

When you have a moment, could you take a look at the changes I made?

halibiram:
tv_screenshot
tested it by connecting the trakt api, cw is working exactly as it should

halibiram:
This proto shard change makes it way easier to implement profile-based trakt support in the future

skoruppa:
> This proto shard change makes it way easier to implement profile-based trakt support in the future

hmm  #1199 ;) 

We still need some testing. Especially adding, removing things from watched, in progress ect ect. I don't have time now, but I made a [test build](http://morigal.pl/upload/app-universal-benchmark-data.apk) for you so you can log in and check.

Btw, for this to even work with release builds there are again needed R8/Progroud entries
```
# ── DataStore / Protobuf ──────────────────────────────────────────────────────
# Protobuf-lite uses reflection to access fields by name with trailing underscore
# (e.g. entries_, key_, value_). R8 must not rename or remove these fields.
-keep class androidx.datastore.preferences.protobuf.** { *; }
-keep class * extends androidx.datastore.preferences.protobuf.GeneratedMessageLite {
    <fields>;
    <methods>;
}
# Safety net: keep all fields ending with _ in all classes (protobuf convention).
-keepclassmembers class ** {
    *** *_;
}
```
And I'm not sure if I like that

halibiram:
> > This proto shard change makes it way easier to implement profile-based trakt support in the future
> > hmm  #1199 ;)
> 
> We still need some testing. Especially adding, removing things from watched, in progress ect ect. I don't have time now, but I made a [test build](http://morigal.pl/upload/app-universal-benchmark-data.apk) for you so you can log in and check.
> 
> Btw, for this to even work with release builds there are again needed R8/Progroud entries
> 
> ```
> # ── DataStore / Protobuf ──────────────────────────────────────────────────────
> # Protobuf-lite uses reflection to access fields by name with trailing underscore
> # (e.g. entries_, key_, value_). R8 must not rename or remove these fields.
> -keep class androidx.datastore.preferences.protobuf.** { *; }
> -keep class * extends androidx.datastore.preferences.protobuf.GeneratedMessageLite {
>     <fields>;
>     <methods>;
> }
> # Safety net: keep all fields ending with _ in all classes (protobuf convention).
> -keepclassmembers class ** {
>     *** *_;
> }
> ```
> 
> And I'm not sure if I like that

I get your point about r8 entries. they are a bit too broad, but necessary for protobuf-lite to function correctly in release builds. i'll look for a more specific way to keep only the required fields. let me check the test build for sync.

halibiram:
@skoruppa I tested the watch progress flows in the test build. Adding/removing watched and in-progress items, mark watched/unwatched actions, batch operations, and persistence after app restart all behave the same as on main, so I did not notice any regression from these changes.

I also tightened the ProGuard/R8 fix to a narrower protobuf-lite keep rule instead of the broader catch-all rules from the review, so the release fix is in place without overkeeping unrelated classes and fields.

PR #1400, opened 2026-04-16
Persist player aspect mode preference

halibiram:
## Summary
~~Persist the player's selected aspect mode across playback sessions. Add debug logs for aspect mode toggle, save, and restore flow.~~

Persist the player's selected aspect mode across playback sessions using **device-local storage** instead of profile-based DataStore. This ensures each device keeps its own aspect ratio preference independently — no cross-device or cross-profile sync. Add debug logs for aspect mode toggle, save, and restore flow.

## PR type

## Why
~~Users want the selected aspect ratio to stay the same until they change it again.~~

Users want the selected aspect ratio to stay the same until they change it again. Per review feedback, aspect mode is now stored in a dedicated `DeviceLocalPlayerPreferences` DataStore (`device_local_player_prefs`) that is **not** tied to any profile and never synced to cloud. This way, different devices can have different aspect ratio preferences without interfering with each other.

## Policy check

## Testing
Tested locally by changing aspect ratio during playback, closing the player, and reopening playback. Confirmed that the selected aspect mode is restored and stays unchanged until the user switches it again. Aspect mode is now stored per-device and does not sync across profiles.

None.

## Breaking changes
None.

## Linked issues
#1433

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/DeviceLocalPlayerPreferences.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/DeviceLocalPlayerPreferences.kt
@@ -1,0 +1,45 @@
+package com.nuvio.tv.data.local
+
+import android.content.Context
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.PreferenceDataStoreFactory
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringPreferencesKey
+import androidx.datastore.preferences.preferencesDataStoreFile
+import com.nuvio.tv.ui.screens.player.AspectMode
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import javax.inject.Inject
+import javax.inject.Singleton
+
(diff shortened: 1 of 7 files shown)


skoruppa:
I'm not sure if we want this in `PlayerSettings`. This will result in setting being synced to cloud and if someone has two devices on which he prefers other aspect ratios, it will be bothersome

Additionally, should we wait with merging this if you are going to revert your aspect ratio changes?

halibiram:
> I'm not sure if we want this in `PlayerSettings`. This will result in setting being synced to cloud and if someone has two devices on which he prefers other aspect ratios, it will be bothersome
> 
> Additionally, should we wait with merging this if you are going to revert your aspect ratio changes?

No need to revert just yet. I found the reason behind that constant buffering behavior

skoruppa:
Merging :)
`DeviceLocalPlayerPreferences.kt` may come handy later if we decide that Advanced Player options (like DV/HDR fallbacks) should also not be synced between devices (as one device may have different compatibility than the other)

halibiram:
> Merging :) `DeviceLocalPlayerPreferences.kt` may come handy later if we decide that Advanced Player options (like DV/HDR fallbacks) should also not be synced between devices (as one device may have different compatibility than the other)

Good idea, we can store the device specific flags here

PR #1403, opened 2026-04-16
Fix modern home performance regression

halibiram:
## Summary

- Move modern home row mapping and lookup building out of `ModernHomeContent` and into the presentation pipeline.
- Reuse precomputed presentation rows in the UI instead of rebuilding carousel rows during composition.
- Keep continue watching, catalog rows, collection rows, and focus lookups intact while reducing duplicate work on the home screen.

## PR type

- Bug fix

## Why

The modern home screen was rebuilding row presentation data inside Compose even though similar mapping was already available from the presentation pipeline. That duplicate work increases allocations and raises the risk of jank during recomposition, especially on Android TV home navigation. This change makes the presentation layer the single source of truth so the UI only consumes prepared rows and lookups.

## Policy check

## Testing

- Manual testing completed on the modern home screen.
- Verified initial load, continue watching rendering, catalog rows, collection rows, focus restore, and general home navigation after the refactor.
- Verified the UI still consumes precomputed presentation rows and preserves hero/trailer behavior during navigation.

None. This change is performance-focused and does not intentionally alter UI design.

## Breaking changes

None.

## Linked issues

None.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/HomeScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/HomeScreen.kt
@@ -563,8 +563,7 @@ private fun ModernHomeRoute(
             { item -> viewModel.onItemFocus(item) }
         },
         onPreloadAdjacentItem = preloadAdjacentItem,
-        onSaveFocusState = saveModernFocusState,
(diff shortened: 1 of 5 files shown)


skoruppa:
Good change. Tested it, and it is indeed smoother. I will think what we can do about trailer job

PR #1407, opened 2026-04-16
Adjustments to scroll focus in Modern View

SelfDL:
## Summary

Changed the way scrolling calculates the next focused index (e.g., Continue Watching down to catalogs row, and expanded poster up/down to different rows.)

## PR type

## Why

Currently when you scroll up/down from an expanded poster, it calculates the next index as the closest point (up or down and to the right). Similar to scrolling down from any index greater than 1 from Continue Watching, instead of scrolling directly down to the previously focused index, it'll scroll down and shoot over right a few indexes.

## Policy check

## Testing

Basic scrolling up, down, left, right between different rows and continue watching indexes with expanded posters to ensure it functions as intended and with expanded posters disabled to ensure that doesn't affect normal use case.

N/A

## Breaking changes

None

## Linked issues

None


skoruppa:
Good Job! Tested changes and now focus out of extended poster behaves as expected

PR #1412, opened 2026-04-17
fix: keep Modern home hero backdrop fade on cache hits

halibiram:
## Summary

Replace `crossfade(400)` with a custom `AlwaysCrossfadeTransitionFactory` on the Modern Home hero backdrop so the fade-in animation plays even on Coil memory-cache hits.

## PR type

- Bug fix

## Why

Coil's built-in `crossfade()` skips the transition when the image is served from the memory cache (`isSampled = false`). On the Modern Home screen this causes the hero backdrop to pop in instantly after the first load — breaking the smooth fade effect users expect when swiping between hero items.

`AlwaysCrossfadeTransitionFactory` forces `CrossfadeTransition` for every `SuccessResult`, regardless of cache status.

## Policy check

## Testing

- Manual: installed debug build on physical Android TV device, navigated to Modern Home, swiped through hero items multiple times. Verified crossfade animation plays consistently on every backdrop change (first load and subsequent cache hits).

N/A — animation behavior, no layout change.

## Breaking changes

- None

## Linked issues

- None

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/AlwaysCrossfadeTransitionFactory.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/AlwaysCrossfadeTransitionFactory.kt
@@ -1,0 +1,44 @@
+package com.nuvio.tv.ui.screens.home
+
+import coil.drawable.CrossfadeDrawable
+import coil.request.ImageResult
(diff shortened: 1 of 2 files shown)


skoruppa:
Tested. Works great. Now "crossfade" is applied always :) I don't see any performance regression

Issue #1428, opened 2026-04-17
[Bug]: subtitle loading clutters playback thread


harrydbarnes:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

0.6.3

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Google TV Streamer

### Android version

Android 14 or 15

### Player mode

Internal player

### Area (tag)

Playback (start/stop/buffering)

### Steps to reproduce

Open app, find a show or film with a lot of subtitles and start playback. Users have reported slowness and stuttering on Reddit.

### Expected behavior

No stutter, smooth playback

### Actual behavior

Lag until subtitles load

### Frequency

Often (more than 50%)

### Did this work before?

Not sure

### Logs (required for crash reports)

```shell

```

### Anything else? (optional)

Will provide a fix shortly.

PR #1429, opened 2026-04-17
Perf: Enable asynchronous subtitle parsing in ExoPlayer

harrydbarnes:
## Summary

Updates PlayerRuntimeControllerInitialization to use `DefaultSubtitleParserFactory` instead of `null` when configuring subtitle parsing on `DefaultMediaSourceFactory`. This delegates text extraction to a background thread instead of the playback/rendering thread, which significantly reduces UI stutter and improves playback startup times, especially for large, text-heavy subtitle files (like those found in certain streaming setups as reported on Reddit.

## PR type

- Bug fix

## Why

Reddit users had reported lag when playback starts

## Policy check

- [Y] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

I am unable to build an apk at the moment due to my setup environment, but the idea makes sense in principle. If someone can provide an APK, I am happy to test.

image

## Breaking changes

Shouldn't be any as far as I know, so marking this as none.

## Linked issues

https://github.com/NuvioMedia/NuvioTV/issues/1428

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactory.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactory.kt
@@ -104,24 +104,28 @@ internal class PlayerMediaSourceFactory {
         return when {
             isHls && !forceDefaultFactory -> HlsMediaSource.Factory(httpDataSourceFactory)
                 .setAllowChunklessPreparation(true)
-                .apply {
-                    customSubtitleParserFactory?.let { parserFactory ->
-                        setSubtitleParserFactory(parserFactory)
-                    }
-                }
+                .applySubtitleParserFactory(customSubtitleParserFactory)
                 .createMediaSource(mediaItem)
             isDash && !forceDefaultFactory -> DashMediaSource.Factory(httpDataSourceFactory)
-                .apply {
-                    customSubtitleParserFactory?.let { parserFactory ->
-                        setSubtitleParserFactory(parserFactory)
-                    }
-                }
+                .applySubtitleParserFactory(customSubtitleParserFactory)
                 .createMediaSource(mediaItem)
             else -> defaultFactory.createMediaSource(mediaItem)
         }
     }
 
     fun shutdown() = Unit
+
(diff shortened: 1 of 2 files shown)


halibiram:
Can you test this out? We need to be absolutely sure that it works correctly

harrydbarnes:
> Can you test this out? We need to be absolutely sure that it works correctly

I can build an apk yes. I'm trying to get info from the Redditor on their device to rule out other issues to ensure a comprehensive fix.

halibiram:
> > Can you test this out? We need to be absolutely sure that it works correctly
> 
> I can build an apk yes. I'm trying to get info from the Redditor on their device to rule out other issues to ensure a comprehensive fix.

ExoPlayer is tricky. Often, what feels like a logical fix just gets in the way of it working right

skoruppa:
What I'm afraid of are ass subtitles. Both built in and from external source. We don't want to break that

If you want a build in which you can login, I can create a test build later

harrydbarnes:
Thanks both. I'm trying out a build now but already run into some issues regarding this version of the update. Libass is one of them, HLS streams is another. Will have another go.

harrydbarnes:
With someone to test who had the issue

harrydbarnes:
@skoruppa the user in question has tried to test this, however their setup doesn't allow them to install add ons easy (keyboard doesn't work apparently, and isn't linked to Google Home so cannot remotely type). Is there a way you can build a version with login enabled? Would be great to know how to do this on my side for some other test, so far only this one appears affected, but would help test playback and replicating scenarios a bit easier

skoruppa:
@harrydbarnes you can't make a build that will log in to our system ;) I would need to publish our API keys which for obvious reason I can't. But each time you want such build, just ping me, and I will try to provide one 

http://morigal.pl/upload/app-universal-benchmark-sub.apk

skoruppa:
@harrydbarnes how testing concluded?

harrydbarnes:
> @harrydbarnes how testing concluded? 

No, I focussed on some other quick wins. Will look again at this this week.

skoruppa:
This seems to not be needed anymore

PR #1452, opened 2026-04-19
Add Reset button and DPAD_CENTER shortcut to clear subtitle delay (fixes #1329)

### LINKED ISSUE #1329: [Bug]: Cannot reset subtitle delay to 0 after using "Sync by line"
### Pre-flight checks

### App version / commit

0.6.0

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

NVIDIA Shield TV

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

### Steps to reproduce

1. Open a video with subtitles.
2. Navigate to the subtitle delay settings.
3. Use the "Sync by line" function.
4. Attempt to manually adjust the subtitle delay back to 0.
5. Observe that the delay cannot be reset to 0.

### Expected behavior

I should be able to manually reset the subtitle delay back to exactly 0 after using the 'Sync by line' feature. Ideally, there would also be a dedicated 'Reset to 0' button and the ability to set the delay beyond the current 60-second limit.

### Actual behavior

After utilizing the 'Sync by line' feature, it becomes impossible to manually adjust the subtitle delay back to exactly 0. The offset gets stuck and cannot be returned to its default state.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

Add a "Reset to 0, Default" button: Adding a dedicated button to instantly reset the delay would solve this issue and make it much more user-friendly.

Increase the 60-second limit: It would be very helpful if the manual offset limit could be increased beyond the current maximum of 60 seconds for subtitles that are severely out of sync.

https://github.com/user-attachments/assets/3b3b518b-c981-4162-99f1-3af5fa77ec1f

Sanjays2402:
## Summary

Adds a reset-to-zero mechanism for subtitle delay. Users can now clear an auto-sync offset (e.g. +238ms from "Sync by line") via **DPAD_CENTER / Enter** on the delay slider or a new **Reset** button in the subtitle delay overlay.

## PR type

- Bug fix

## Why

After using "Sync by line" to auto-sync subtitles, the delay is set to an arbitrary millisecond value (e.g. +238ms). Since manual adjustment uses 100ms steps, users can never return to exactly 0ms. There's no reset mechanism at all, so an incorrect auto-sync effectively locks the user out of the original subtitle timing.

## Policy check

## Testing

- Manually tested on device: opened a video, ran "Sync by line" to produce a non-zero delay (e.g. +238ms), pressed DPAD_CENTER / Enter on the slider → delay returned to 0ms and the overlay refreshed correctly.
- Manually tested the new **Reset** button in the subtitle delay overlay — same behavior as the key binding.
- Verified the existing `adjustSubtitleDelay()` side effects (MPV property update, overlay refresh) still run, since the reset reuses that path by applying the negative of the current delay.

## Breaking changes

None.

## Linked issues

Fixes #1329

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerPlaybackEvents.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerPlaybackEvents.kt
@@ -813,6 +813,9 @@ fun PlayerRuntimeController.onEvent(event: PlayerEvent) {
         is PlayerEvent.OnAdjustSubtitleDelay -> {
             adjustSubtitleDelay(event.deltaMs, event.showOverlay)
         }
+        PlayerEvent.OnResetSubtitleDelay -> {
+            adjustSubtitleDelay(-_uiState.value.subtitleDelayMs, true)
+        }
         PlayerEvent.OnShowSpeedDialog -> {
             val state = _uiState.value
             if (state.tunnelingEnabled) {
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerScreen.kt
@@ -400,7 +400,10 @@ fun PlayerScreen(
                                     return@onKeyEvent true
                                 }
                                 KeyEvent.KEYCODE_DPAD_CENTER,
-                                KeyEvent.KEYCODE_ENTER,
+                                KeyEvent.KEYCODE_ENTER -> {
+                                    viewModel.onEvent(PlayerEvent.OnResetSubtitleDelay)
+                                    return@onKeyEvent true
+                                }
                                 KeyEvent.KEYCODE_DPAD_UP -> {
                                     return@onKeyEvent true
                                 }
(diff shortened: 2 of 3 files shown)


skoruppa:
Can you please apply our PR template?

Sanjays2402:
@skoruppa I've updated the description to match the PR template of this repo. I'll remember this :)

skoruppa:
Can you also show where you put the button? :)

Sanjays2402:
The new **Reset** button lives inside the existing **Subtitle Delay overlay** (the one with the delay slider), directly to the **left of the existing "Sync Line" button**. They're siblings in a new `Row` with 10 dp spacing.

In `PlayerScreen.kt → SubtitleDelayOverlay()`:

```kotlin
Row(horizontalArrangement = Arrangement.spacedBy(10.dp)) {
    Card(onClick = onResetDelay, ...) { Text("Reset", ...) }          // ← new
    Card(onClick = onOpenSyncByLine, ...) { Text("Sync Line", ...) }  // existing
}
```

So the overlay now shows `[ Reset ] [ Sync Line ]` under the delay slider. Reset calls `adjustSubtitleDelay(-currentDelay, true)` via the new `PlayerEvent.OnResetSubtitleDelay`, which zeros out the delay and refreshes the overlay.

There's also a non-UI path: pressing **DPAD_CENTER / Enter** while the delay slider is focused fires the same event, so users on TV remotes can reset quickly without visual button focus.

Also corrected the PR title — the original "fast auto scrolling button" was a copy-paste slip; the actual change is the subtitle delay Reset for issue #1329.

skoruppa (approved):

Issue #1093, opened 2026-03-26
[Bug]: Age Rating Localization Failure (Simple Language Codes)


SK-Gits:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

Beta 0.6.3

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

All

### Android version

12/14

### Player mode

Internal player

### Area (tag)

UI / Layout

### Steps to reproduce

1. Change the application or system language: Set the interface to French (fr).
2. Search for a recent movie: Navigate to the movie page of a film with a known French rating (e.g., Superman 2025 or any released film with a TP (Tous Publics) rating).
3. Observe the metadata: Check the displayed Age Rating badge.

### Expected behavior

The application should correctly identify the region from the ISO language code "fr" and display the local French rating (e.g., TP, 12).

### Actual behavior

The application fails to extract the region from the short ISO code "fr", as the current logic expects a hyphenated string. Since the French region (FR) is missed, the search priority defaults to US and GB, displaying an American rating (e.g., PG-13) instead of the local French one.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

<img width="1575" height="951" alt="Image" src="https://github.com/user-attachments/assets/926d7b7b-3b49-424e-89f8-d66e3c800e2c" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/84a53059-b13a-4831-ae45-d4b728a339d6" />

### Logs (required for crash reports)

```shell

```

### Anything else? (optional)

1. Problem SummaryThe age rating system fails to identify the user's region when a simple language code is used (e.g., "fr" instead of "fr-FR"). Consequently, the system ignores local maturity ratings (such as -12 or -16 in France) and defaults to American (US) or British (GB) classifications.

[...]

PR #1454, opened 2026-04-19
Fix original audio language resolution for CW and detail screen playback

skoruppa:
## Summary

Resolve original content language from addon meta `language` and `country` fields so the player can select the correct audio track even when TMDB enrichment is disabled.

Previously `contentLanguage` was only populated by TMDB enrichment. When CW enrichment was off, the player received `null` and fell back to device/secondary language instead of original.

Changes:
- Add `Meta.resolveContentLanguage()` with fallback chain: `language` -> `country` -> `null`, normalizing ISO 639-2/3 codes and full names to ISO 639-1
- Detail screen now uses `meta.resolveContentLanguage()` instead of raw `meta.language`
- CW enrichment applies the same `normalizeLanguageCode` / `countryToLanguageCode` fallback from addon meta

## PR type

- Bug fix

## Why

Users with "Preferred audio: Original" and TMDB CW enrichment disabled reported that anime/foreign content played in English instead of the original language. The player had no `contentLanguage` to resolve "Original" and fell back to device language.

Also fixes stale next-up items persisting in CW after unmarking all watched episodes of a series.

## Policy check

## Testing

- Anime with TMDB CW enrichment disabled: play from CW -> audio correctly selects Japanese
- Anime from detail screen with addon meta `country: "jpn"`, no `language` field -> player selects Japanese
- TMDB enrichment enabled: still uses TMDB language (no regression)
- Tested both ExoPlayer and mpv engine paths

No changes

## Breaking changes

Nothing should break 

## Linked issues

Reported on discord

### DIFF
--- a/app/src/main/java/com/nuvio/tv/domain/model/Meta.kt
+++ b/app/src/main/java/com/nuvio/tv/domain/model/Meta.kt
@@ -85,6 +85,85 @@ data class Meta(
     }
 }
 
+/**
+ * Resolves the original content language as an ISO 639-1 code.
+ * Falls back to [country]-based inference when [Meta.language] is absent.
+ */
+fun Meta.resolveContentLanguage(): String? {
+    normalizeLanguageCode(language)?.let { return it }
+    countryToLanguageCode(country)?.let { return it }
+    return null
+}
+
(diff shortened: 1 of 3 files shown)


SK-Gits:
Will this also fix the age display issue related to #1093?

PR #1482, opened 2026-04-21
feat: seek-preview thumbnails (YouTube/Netflix style scrubbing)

### LINKED ISSUE #1481: [Feature]: Seek preview thumbnails on scrub
### Area (tag)

Playback

### Problem statement

When scrubbing through a video with the D-pad, the player only shows the target timestamp — there's no visual indication of what scene I'm on. 

## Feature: Seek Preview Thumbnails (Opt-in)

Add a new setting under **Playback** called **"Seek preview thumbnails"** (disabled by default).

When enabled, the app generates small preview images at intervals, allowing users to see a snapshot of the scene while fast-forwarding or scrubbing (similar to YouTube/Netflix).

---

### Behavior

- On video start, thumbnails for the **first 25%** are generated in the background.
  - Does **not block playback**
  - Takes a few seconds

- As playback continues, thumbnails are generated progressively in **25% chunks** until the full video is covered.

- Thumbnails are **cached to disk**:
  - Max size: **200MB**
  - Oldest entries are automatically removed

- On rewatch/resume:
  - Thumbnails load instantly
  - No additional bandwidth is used

---

### Technical Details

- Thumbnail interval: **every 20 seconds**

- Supported:
  - Debrid-resolved **MP4 / MKV streams**

- Not supported:
  - **HLS / DASH streams** (silently skipped, no errors)

- Playback engines:
  - **ExoPlayer**
  - **MPV**

- Format:
  - **160×90 JPEG**
  - Optimized for fast decoding and TV display

_No response_

_No response_

AKhalil609:
## Summary

Adds YouTube/Netflix-style seek-preview thumbnails. When the opt-in setting is enabled, the app pre-generates JPEG frames from the video in the background and displays them above the progress bar while the user scrubs. Progress bar tick marks show which positions are already cached.

The core generation engine has been extracted into a standalone Android library — **[FrameScout](https://github.com/AKhalil609/FrameScout)** (`io.framescout`, via JitPack). This keeps the NuvioTV diff focused on integration code rather than the full engine implementation.

**Generation pipeline:**
- Two-pass approach: sparse pass (one frame every 3 min, whole movie) runs first for instant coarse coverage, then dense passes (one frame every 30 s, 25% chunks) fill in detail
- MP4 + H.264 only — MKV excluded (Amlogic SoC all-black bug), HEVC/x265 MP4 excluded (silent grab failure on some hardware)
- Source selection scans all available addon streams for the best MP4 candidate (composite score: release group > source type > platform > quality), probes each candidate's container duration to match the playing file (±3 s), avoiding alternate cuts
- Current stream fast-path: if the active playback URL is already MP4+H.264, it is used directly without probing alternatives
- `OPTION_CLOSEST_SYNC` — snaps to nearest keyframe, no forward decoding; ~10–15× faster than `OPTION_CLOSEST` on remote HTTP streams
- 2 parallel workers with round-robin timestamp sharding
- HDR/10-bit content handled: `RGBA_F16`, `HARDWARE`, `RGBA_1010102` bitmaps converted via `Canvas.drawBitmap()` onto `ARGB_8888` (color-managed, avoids near-black output)
- 20 s per-frame timeout prevents worker stalls on large I-frames over HTTP
- Bad debrid source detection: rejects sources shorter than 60 s (error clips from uncached debrid content) and retries with next candidate; zero-frame sparse pass also triggers retry (codec-level failure)
- Progress bar tick marks: `cachedFractions` StateFlow drives 2 dp white marks at each cached position; updated per chunk commit, no per-frame recomposition cost
- CDN mirror deduplication via `distinctBy { filename }` before probing

**Settings (Settings → Playback → Seek Preview Thumbnails):**
- Off by default
- Clear Thumbnail Cache (shows current disk usage)
- Cache Size Limit: 50 / 100 / 200 MB (default 200 MB)

Closes #1481

## PR type

## Why

Scrubbing through a movie on a TV with only a progress bar and timestamp gives no visual feedback about what scene you are jumping to. This is a common UX gap compared to streaming services. Linked feature request: #1481

## Policy check

Approved in #1481

## Testing

**Manual:**
- Tested on Android emulator (API 34) and Amazon Fire Stick 4K (Android 9 / API 28)
- Verified thumbnails appear while scrubbing with ExoPlayer engine
- Verified sparse pass delivers coarse full-movie coverage within ~2–3 min for a 2h film
- Verified dense chunks continue after sparse pass without time-gate issues
- Verified HDR content produces correctly exposed thumbnails (not near-black)
- Verified HEVC/x265 MP4 sources are skipped and next candidate is tried
- Verified debrid error clips (short duration) are rejected and retried
- Verified HLS/DASH/torrent streams silently skip generation with no crash
- Verified cache persists across app restarts and is reused on rewatch
- Verified progress bar tick marks appear immediately when a cached file is found on resume
- Verified toggle off disables all thumbnail activity

**Automated:**
- Unit tests for `SeekPreviewCacheKey`, `SeekPreviewThumbnailStore`, and `SeekPreviewGenerator` live in the [FrameScout](https://github.com/AKhalil609/FrameScout) library repository

Screenshot 2026-04-27 at 12 04 09 PM

Screenshot 2026-04-27 at 12 25 15 PM

Seeking-ezgif com-video-to-gif-converter (2)

_(scrubbing demo recorded on Fire Stick 4K — thumbnail tracks scrubber position above progress bar)_

## Breaking changes

None

## Linked issues

Closes #1481


EzzatBoukhary:
Please follow the PR body template

EzzatBoukhary:
I've been wanting this feature. Excited to see how well this will work. What's the performance like with the setting enabled? Any noticeable performance hits in playback or while scrubbing?

AKhalil609:
@EzzatBoukhary Nothing that I experienced at least as long that you have a device that is at least 1GB of memory it should be fine.

skoruppa:
Here is a test build including latest develop and your changes. We need thorough tests for as many streams, types, stream providers, p2p, plugins and devices before merging this. Aslo, can you remove your .idea files from the PR? 

http://morigal.pl/upload/app-full-universal-benchmark-seek.apk

I tested this without any luck. Tried few http streams, few links from debris and the thumbnails were not present while seeking. I'm not sure how I feel about feature that doesn't work most of the time...

mytee12:
Just did some quick tests to see this in action. Setting on seek thumbnails on Sparse setting with MKV with RD and TB. 3 out of 6 worked on first result.  

The ones that failed, when i tried the next link they worked. I feel like this is likely to be issue with the "(±3 s)" requirement for the videos to scrub for thumbs.  Worked somewhat with Usenet/NZBDav too but not sure what container they were and some thumbs in a video showed the purple aiostreams error video. 

Sometimes it felt like the aspect ratio of the thumbs were stretch high, not 100% though. Maybe someone can find a workaround the the MKV bug, would be so clean then.

AKhalil609:
@skoruppa @mytee12 

Thank you both for testing. There are a few hard requirements for thumbnails to generate — these come from how Android's `MediaMetadataRetriever` works over HTTP, not something we can configure away:

- **Feature must be enabled**: Settings → Playback → **Seek Preview Thumbnails** (off by default)
- **MP4 + H.264 only**: MKV is excluded (seek index at end of file breaks on many Android TV SoCs), HEVC/x265 is excluded (frame extraction silently fails on some hardware), HLS/DASH are excluded entirely
- **No P2P**: `MediaMetadataRetriever` jumps to random byte offsets — a torrent stream can't serve that reliably and would kill playback performance
- **Content ≥ 3 min**: Short clips and debrid "not cached yet" error clips are rejected via a duration probe — @mytee12 the purple AIOStreams video is exactly this case, worth investigating why the probe didn't catch it
- **Source duration must match within ±3 s**: The format of the stream you're watching doesn't matter — even a high-bitrate MKV will get thumbnails as long as the addon's source list contains at least one MP4 x264 stream whose container duration matches within ±3 s. @mytee12 the 3/6 failure rate is likely this — the second link worked because a matching candidate was found

**Known issue**: On some devices `getScaledFrameAtTime` stretches frames to 160×90 without preserving aspect ratio — most visible on 2.39:1 content. Fix is tracked as a follow-up.

---

Ideally this would work for everyone, but for this first iteration I scoped it to debrid users since debrid reliably exposes direct MP4 HTTP links — a solid baseline to build on. Every user's setup is different (debrid, P2P, direct streaming) and it's hard to cover all cases. One possible direction is sourcing a free MP4 stream independently of the user's addons purely for thumbnail extraction, though that still risks a duration mismatch and wouldn't be guaranteed to work either.

If you know of a better approach or a more reliable way to get MP4 streams for extraction, I'd be happy to explore it — open to any suggestions.

I'll remove the `.idea` files from the PR.

haveAnIssue:
UP

PR #1483, opened 2026-04-21
fix: wait for resolved profile before rendering home

halibiram:
## Summary

That does not just create a visual flash. Even when the wrong layout is not visibly shown, the app still does work for two layouts in sequence in the background: first for the temporary layout, then again for the actual resolved layout. This adds unnecessary startup cost and slows down initial app launch.

- Prevents Home from briefly rendering the wrong layout during cold start.
- Waits for the active profile and layout preferences to resolve before Home content is allowed to render.
- Avoids doing startup work twice when Modern Home is selected.

## PR type

- Bug fix

## Why

When Modern Home is selected, the app can briefly render another layout first during cold start. This happens because Home starts building UI from a temporary default profile/layout before the real active profile is loaded from datastore.

That does not just create a visual flash. It also adds unnecessary startup cost because the app begins work for the wrong layout first, then repeats layout-dependent work again once the real profile and preferences arrive.

## Policy check

## Testing

- Ran `./gradlew :app:compileDebugKotlin`
- Manually verified the Home startup flow to ensure Home waits for resolved profile/layout state before rendering.
- Confirmed the incorrect temporary layout is no longer shown first when Modern Home is selected.

- Problem 

https://github.com/user-attachments/assets/7fe90289-3150-4795-ac83-597aeeba5f9b

## Breaking changes

- None

## Linked issues

- None

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/profile/ProfileManager.kt
+++ b/app/src/main/java/com/nuvio/tv/core/profile/ProfileManager.kt
@@ -10,6 +10,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
 import java.io.File
 import javax.inject.Inject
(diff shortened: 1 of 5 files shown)


skoruppa:
Another small piece ;) Not the sole reason, but it is now a bit better

PR #1490, opened 2026-04-22
fix(detail): keep TMDB Collection tab inline with other integration tabs (refs #1488)

### LINKED ISSUE #1488: [Bug]: TMDB Integration Collection tab misaligned - can’t browse other sections
### Pre-flight checks

### App version / commit

0.6.5

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Nvidia Shield Pro 2019

### Android version

Android 11

### Player mode

Internal player

### Area (tag)

UI / Layout

### Steps to reproduce

Starting with 0.6.5 - The Collection TMDB integration tab is pushed below the rest of the tabs which is blocking navigation to the content of the Creator and Cast & More like this tabs

1. Enable TMDB integration Creator and Cast, More like this, Trailer, and Collection
2. Open Detail page of movie that has Collection
3. Attempt to navigate content of Creator and Cast or More like this

### Expected behavior

Selecting one of the tabs and then clicking down should take you to the content of the tabs to scroll through

### Actual behavior

Clicking down from Creator and Cast & More like this takes you to the Collection tab and its content - unable to browse content of other sections

Image

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

Sanjays2402:
## Summary

Puts the Collection tab back on the same row as Creator & Cast / More Like This / Trailer. Since 0.6.5 it's been rendering in a second row below, which breaks D-pad focus on Android TV — pressing DOWN from any top-row tab skips the Collection tab button and drops you straight into Collection's content.

## PR type

- Bug fix

## Why

Fixes #1488.

The culprit is `6aaa286e` ("feat: adding trailersection"). That commit restructured `PeopleSectionTabs` into a `Column` that splits the row whenever `tabs.size > 3 && COLLECTION is present`. Once Trailer became a regular tab, any movie/show with a Collection hits that condition and Collection gets pushed to its own standalone row. The new row is outside the main tab row's focus group, so D-pad DOWN skips past it.

This PR reverts `PeopleSectionTabs` to the single-Row layout it had before 0.6.5. Nothing else changes.

## Policy check

## Testing

Opening as a **draft** because I don't have my Android TV set up locally to verify D-pad focus. What I did check:

- Compared against the pre-regression code shape from 0.6.4-beta — same structure.
- All symbols used in the revert are already imported in `MetaDetailsScreen.kt`.
- Diff is 1 file, +17/−33.

I'll mark ready-for-review once I've confirmed on device that:
1. Collection tab button sits inline with the other three.
2. D-pad DOWN from each sibling tab lands on that tab's own content.
3. Phone layout still looks fine (narrow screens rely on the parent `LazyRow` scrolling, which the revert preserves).

@mistersandman-64 — if you've got the Shield Pro handy, a quick sanity check on this would be super helpful.

Before is documented by the screenshot on #1488 (Collection button on its own second row).
After should have all four tabs on one row.

Will drop a before/after once tested on device.

## Breaking changes

None.

## Linked issues

Refs #1488. Keeping it as `refs` rather than `fixes` until device verification.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsScreen.kt
@@ -1968,48 +1968,32 @@ private fun PeopleSectionTabs(
 
     val defaultRequester = tabs.first().focusRequester
     val restorerRequester = tabs.firstOrNull { it.tab == activeTab }?.focusRequester ?: defaultRequester
-    val shouldSplitCollection = tabs.size > 3 && tabs.any { it.tab == PeopleSectionTab.COLLECTION }
-    val firstRowTabs = if (shouldSplitCollection) tabs.filterNot { it.tab == PeopleSectionTab.COLLECTION } else tabs
-    val secondRowTabs = if (shouldSplitCollection) tabs.filter { it.tab == PeopleSectionTab.COLLECTION } else emptyList()
 
-    Column(
+    Row(
         modifier = Modifier
             .fillMaxWidth()
             .padding(top = 20.dp, start = 48.dp, end = 48.dp)
             .focusRestorer(restorerRequester),
-        verticalArrangement = Arrangement.spacedBy(8.dp)
+        verticalAlignment = Alignment.CenterVertically
     ) {
-        @Composable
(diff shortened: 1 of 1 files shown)


skoruppa:
Hello! This was already fixed upstream and PR is no longer needed :)

Issue #1365, opened 2026-04-13
[Bug]: Slow start (stream RD) compared to starting release


Syntist:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).

### App version / commit

0.6.1-beta

### Install method

GitHub Release APK

### Platform

Fire TV

### Device model

Fire TV Lite

### Android version

Android 12

### Player mode

Internal player

### Area (tag)

Playback (start/stop/buffering)

### Steps to reproduce

1. Open any show/movie
2. Select stream
3. Wait for it to start

### Expected behavior

In older version of NuvioTV, one of the major reason I loved nuvio was how fast it start the stream compared to stremio, but Now I guess it's in same zone, it takes alot of time to start stream. mostly take about 30+ Sec.

### Actual behavior

Should be fast, but i guess it does alot of things before starting the stream?

### Frequency

Always

### Did this work before?

Yes was faster in older versions

skoruppa:
@Syntist can you check if disabling AFR and libass changes anything? Also, set subtitles load method to fast. With those changes there should be nothing slowing down stream loading

Syntist:
Subtitles load method to fast did the tricks other two were already disable, thanks

One of the reason i love this app is it's really fast in running stream compared to stremio, even Netflix lol

PR #1549, opened 2026-04-25
Optimize player startup loading

halibiram:
## Summary

- Optimized ExoPlayer startup setup by running stream MIME detection and startup subtitle preparation concurrently instead of serializing both pre-playback tasks.
- Kept the existing 10 second startup subtitle prefetch timeout, but skipped OpenSubtitles hash calculation during the startup prefetch path so extra stream `HEAD` / range requests do not add blocking work before playback preparation.
- Preserved the existing full addon subtitle fetch behavior outside startup, including hash calculation when subtitles are fetched normally after player setup.
- Replaced hardcoded ExoPlayer `DefaultLoadControl` values with the `BufferSettings` values loaded through `PlayerSettingsDataStore`.
- Applied the buffer values loaded from `PlayerSettingsDataStore` for min buffer, max buffer, initial playback buffer, rebuffer playback buffer, target buffer size, back buffer duration, and keyframe back-buffer retention.
- Lowered the default initial playback buffer threshold in `BufferSettings` from 2500ms to 1000ms so fresh installs can start playback sooner.

## PR type

## Why

Player startup was doing multiple pre-playback tasks in sequence before preparing playback. In addition, ExoPlayer was using hardcoded load control values instead of the `BufferSettings` values loaded by `PlayerSettingsDataStore`, so DataStore-backed buffer defaults and saved values were not reflected in the actual `DefaultLoadControl` configuration.

This change reduces unnecessary serialized startup work while keeping the existing subtitle timeout behavior intact. It also applies the DataStore-backed buffer values to ExoPlayer and lowers the default initial playback buffer threshold, which makes playback start sooner in practice.

## Policy check

Not required; this is a small focused maintenance optimization for player startup latency.

## Testing

- Ran `./gradlew.bat :app:compileDebugKotlin` successfully.
- Ran `./gradlew installDebug` successfully.
- Manually tested playback startup and confirmed the player opens faster.

| `main`  | `debug`  |
| :---: | :---: |
| <video src="https://github.com/user-attachments/assets/6c052959-eeec-4cd5-a0a9-d2e23c781542" width="300" /> | <video src="https://github.com/user-attachments/assets/f753f0e2-3d5a-4d36-8551-15e8af0145f1" width="300" /> |

*For my tests, I specifically picked the slowest loading source on my end*

## Breaking changes

None.

## Linked issues

Relates to #1365
Relates to #1428

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
@@ -177,12 +177,18 @@ internal fun PlayerRuntimeController.initializePlayer(
                 }
                 return@launch
             }
-            resolveCurrentStreamMimeType(
-                url = url,
-                headers = headers
-            )
+            val streamMimeTypeJob = async {
+                resolveCurrentStreamMimeType(
+                    url = url,
+                    headers = headers
+                )
+            }
             mpvInitializationInProgress = false
-            val startupSubtitlePreparation = prepareStreamStartSubtitles(playerSettings, showLoadingStatus)
+            val startupSubtitlePreparationJob = async {
+                prepareStreamStartSubtitles(playerSettings, showLoadingStatus)
+            }
+            streamMimeTypeJob.await()
+            val startupSubtitlePreparation = startupSubtitlePreparationJob.await()
             afrJob.await()
             requestedUseLibassByUser = playerSettings.useLibass
             val useLibass = when {
@@ -198,17 +204,7 @@ internal fun PlayerRuntimeController.initializePlayer(
                     libassRenderType = playerSettings.libassRenderType
                 )
             }
-            val loadControl = run {
-                DefaultLoadControl.Builder()
-                    .setTargetBufferBytes(100 * 1024 * 1024)
-                    .setBufferDurationsMs(
-                        DefaultLoadControl.DEFAULT_MIN_BUFFER_MS,
-                        70_000,
-                        DefaultLoadControl.DEFAULT_BUFFER_FOR_PLAYBACK_MS,
-                        5_000
-                    )
-                    .build()
-            }
+            val loadControl = buildLoadControl(playerSettings)
 
             
             trackSelector = DefaultTrackSelector(context).apply {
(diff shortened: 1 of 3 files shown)


harrydbarnes:
Love this. Should really help.

app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
```diff
@@ -630,6 +656,7 @@ internal suspend fun PlayerRuntimeController.prepareStartupSubtitles(
 
     val fetchedSubtitles = withTimeoutOrNull(STARTUP_SUBTITLE_PREFETCH_TIMEOUT_MS) {
         fetchAddonSubtitlesNow(
+            computeHash = false,
```
skoruppa: Why are you disabling hash computing? If I'm not wrong we already skip hash calculation when streaming addon provides videoHash. In other case we should still do it

halibiram: You're right. My intention was only to skip hash calculation during startup prefetch to avoid delaying playback. If the stream already provides videoHash, it is still used. But when videoHash is missing and the startup prefetch succeeds, we may skip the later full fetch and never compute the hash.

halibiram: Thoughts on making the MIME resolution and subtitle prep run in parallel? It brings a small improvement, but if you guys don't like the approach, I can just go ahead and close the PR

skoruppa: Well, it helps only with one specific scenario. When someone has subtitles obtaining method not set to fast (so we are waiting for them on loading screen) and hash is provided by addon so we don't need to calculate it before requesting subtitles. In other scenarios this should not make any difference if I'm not mistaken

halibiram: Yep, you're totally right. Scope of this improvement is so small that we probably don't even need it

ram130:
Looks good. Now I got to reconsider my buffer approach for dv 7

halibiram:
> Looks good. Now I got to reconsider my buffer approach for dv 7

I'm really looking forward to your DV7 work. Even though my TV supports DV7, it still struggles with playback. I tried your test build and it's awesome

ram130:
> > Looks good. Now I got to reconsider my buffer approach for dv 7
> 
> I'm really looking forward to your DV7 work. Even though my TV supports DV7, it still struggles with playback. I tried your test build and it's awesome

I had no idea you were monitoring the thread haha. 🤣🤣🤣. I appreciate you. Thank you so much.

PR #1550, opened 2026-04-25
Memorize last selected profile

iamohdisa:
## Summary

To memrozie the last selected profile by the user and select it by default on the next application launch. The user can turn this feature on or off in Settings > Advanced. By default it is off.

## PR type

## Why

To enhance the UX of the application so the user does not have to always select the mostly used profile everytime they launch the application.

## Policy check

## Testing

- Turned this feature off
1. Selected profile 1, then relaunched the app, the profile selection screen appears.
2. Selected profile 2, then relaunched the app, the profile selection screen appears.

- Turned this feature on
1. Selected profile 1, then relaunched the app, the app loads profile 1
2. Selected profile 2, then relaunched the app, the app loads profile 2

image

## Breaking changes

None

### DIFF
--- a/app/src/main/java/com/nuvio/tv/MainActivity.kt
+++ b/app/src/main/java/com/nuvio/tv/MainActivity.kt
@@ -251,6 +251,7 @@ class MainActivity : ComponentActivity() {
 
             val activeProfileId by profileManager.activeProfileId.collectAsState()
             val profiles by profileManager.profiles.collectAsState()
+            val hasEverSelectedProfile by profileManager.hasEverSelectedProfile.collectAsState()
             val activeProfile = remember(activeProfileId, profiles) {
                 profiles.firstOrNull { it.id == activeProfileId }
             }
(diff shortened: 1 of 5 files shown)


jammekod:
Great feature, however as someone who shares Nuvio with my partner it would be annoying to get her CW on my profile :/ forcing choosing profile actually avoids that.

iamohdisa:
@jammekod That's a good point. Maybe a toggle option in the settings to whether enable or disable this feature would be better

chayons-source:
Inserting the profile name in the 'Continue Watching' row could help here. When you open the app, you see 'Continue Watching for xxx'.

iamohdisa:
Updated the PR to allow users to either enable or disable this feature in Settings > Advanced

skoruppa:
What about profiles with PIN? What will happen if last selected profile has a PIN?

iamohdisa:
> What about profiles with PIN? What will happen if last selected profile has a PIN?

@skoruppa It will not bypass the profile selection screen in this case. 
Flag !activeProfileHasPin will be false and it will prompt the user to select a profile

skoruppa:
@iamohdisa can you rebase to the latest develop? We changed a bit how ADnvaced Settings look, so you could move the setting above the "Clear Continue Watching Cache". I really don't want to people to press this button if they don't need to :D

iamohdisa:
@skoruppa done

image

Felitendo:
This is awesome, thanks :)

PR #1551, opened 2026-04-25
Feat: TMDB as a source for collections

tapframe:
## Summary

Adds TMDB backed sources to Collections. Folders can now use TMDB public lists, movie collections, production companies, networks, presets, and custom discover queries together with existing addon catalogs.

## PR type

- Approved larger change (DISCUSSED INTERNALLY) 

## Why

Collections were only using addon catalogs before. This lets users build more useful live collections from TMDB.

## Policy check

- [x] If this is a larger or directional change, I linked the **approved** feature request issue below. [DISCUSSED INTERNALLY]

Discussed internally.

## Testing

- Manually checked TMDB source creation for presets, lists, collections, production companies, networks, and custom discover.
- Verified existing addon catalog collections still load after migration.

Screenshot 2026-04-25 at 8 04 25 PM

Screenshot 2026-04-25 at 8 04 47 PM

## Breaking changes

None for normal upgrades. Existing addon catalog collections are migrated for backward compatibility.

## Linked issues

N/A


skoruppa:
1. When going to the TMDB Sources, the focus should be on the "Presets" tile
2. Groin back from there should move you back to "Edit Folder" view (but this is an existing bug - "Add Catalog" behaves the same)
3. As we are reusing the existing "Add catalog" view, the "Done" in right top corner is a bit misleading. In catalogs, we select more than one and need to confirm with it, here it just goes back 
4. We should filter out "planned" and "in production" things for Networks just like we do when going to Networks from Details Screen
5. Maybe a Movie/Series option should also be present in network? I don't know if network lists can have movies. And type displayed in Collection should be "Series" not "TV" to stay coherent. 
6. After adding TMDB Source it is not possible to edit it 
7. Collections and Public lists should also have sort options ("original" for Lists must be present)
6. TMDB enrichment could be disabled for items 

But overall, it works great. Sharing jsons with this will make a way more sense.

EzzatBoukhary:
I was working on this feature on a fork of mine before it was discussed internally because I saw users requesting it but I haven't reached the final point. However, I'd suggest an easier way to use the presets like how I was implementing it. 

I had a button in the Collection page that said "Add presets" or something like that. It would add collections for all the common collections like streaming services, genres, studios, decades, directors, and actors automatically with all the catalogs of each. The user can then decide if they want to edit, delete, or add things to them.

I would suggest having this flow available to the user because it makes it much easier to use.

skoruppa:
Now it works amazing. 

The editor in itself could see some small improvements 
- going back from "Folder Editor" should not move you to "Collections"
- when going back (for example from Add Catalog) it should remember the scroll position and last focused elemnt (restore focus)

But that not necessary must be part of this PR

As to what @EzzatBoukhary wrote, I don't agree with such complex "presets" that include many folders (at least not as a part of this PR - but with TMDB implemented a "collection presets" is a valid feature), but we can add more regular presets to TMDB. I imagine that "Latest" and "Popular" Networks will be the most popular folders, so networks can have those as well. 

What about QR Code Webserver Configuration Editor? I assume it will be done later

PR #1556, opened 2026-04-25
Refactor lastFocusedIdx handling in ModernHomeRows

SelfDL:
## Summary
Updated the row logic to track the user's focus position in real-time. Instead of just setting an initial starting point, the row now continuously updates its "current position" state as the user navigates.

## PR type

## Why
Prevent unexpected index jumping when navigating out of sidebar back to content.

## Policy check
- [ ] If this is a larger or directional change, I linked the **approved** feature request issue below.

N/A

## Testing
Verified both normal and fast paced vertical/horizontal navigation for both normal and expanded cards function as intended. Tested sidebar-to-content focus (at the start and middle of rows) across CW and category rows.

None.

## Breaking changes
None.

## Linked issues
None.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeRows.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeRows.kt
@@ -693,7 +693,8 @@ internal fun ModernRowSection(
         }
 
         CompositionLocalProvider(LocalBringIntoViewSpec provides horizontalBringIntoViewSpec) {
-            val lastFocusedIdx = focusedItemByRow[rowKey] ?: 0
+            val initialIdx = remember(rowKey) { focusedItemByRow[rowKey] ?: 0 }
+            var lastFocusedIdx by remember { mutableIntStateOf(initialIdx)}
             val restoreIdx = lastFocusedIdx.coerceIn(0, (row.items.size - 1).coerceAtLeast(0))
             val restoreStableKey = "${row.key}_$restoreIdx"
             val restoreFocusRequester = uiCaches.requesterFor(rowKey, restoreStableKey)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nuvio/tv/ui/screens/home/ModernHomeRows.kt
```diff
@@ -700,10 +700,19 @@ internal fun ModernRowSection(
             val restoreStableKey = "${row.key}_$restoreIdx"
             val restoreFocusRequester = uiCaches.requesterFor(rowKey, restoreStableKey)
 
+            val isExpandedOrVerticalScrolling = effectiveExpandEnabled && (
+                    !isActiveRow ||isVerticalRowsScrolling ||
+                            (!isRowScrolling && expandedCatalogFocusKey != null)
+                    )
+
```
skoruppa: This sounds more like "isExpandedAndScrolling". You don't have OR there. 

can you check if isVerticalRowsScrolling is not enough for everything to work? We also have same logic on the Classic View

SelfDL: Huh.. yeah I probably could've changed that name. Also yes I did try that beforehand. It does work fine with just isVertical, but I added !isActive because if you were to scroll fast enough through the row then vertical scroll before the horizontal scroll finished, it wouldn't use last known index and would jump the next row a fex indexes. Was just to prevent that one specific scenario.

skoruppa: Have you verified with Classic Mode? Is the issue present there? And yeah, please change the variable name ;)

SelfDL: Tested it just the make sure and no, the problem is not present in Classic (both focus and that one specific index skip quirk). I also remember from my previous PR regarding focus that it wasn't an issue when I tested then; only a Modern layout thing.

skoruppa:
Sorry for dragging this for so long but I'm still a bit unsure about the change >< can you test latest develop? I removed catalog cache there thay I think caused some issues. We no longer need it with lazy loading

SelfDL:
Just pulled changes. Doesn't seem like anything changed regarding the focus behavior. Issue is still present. @skoruppa

PR #1588, opened 2026-04-28
Add automatic skip options for detected segments

### LINKED ISSUE #1585: [Feature]: Auto Skip Intro
### Area (tag)

Playback

### Problem statement

The ‘Skip Intro’ feature currently requires manual action. It would be better to include an option for automatically skipping both intro and outro, particularly for anime content.

@haaihondschildpad

### Proposed solution

Add Auto skip intro Outro Anime

No, this is only a proposal

_No response_

_No response_

haaihond:
## PR type

## Why
Users now have to press a button to skip segments (intro, recap and outro). This addition will allow users to chose per segment if they want to keep skipping manually or skip automatically.

## Policy check

## Testing

Tested manually. With Anime Skip and IntroDB
image
image

## Breaking changes

None
## Linked issues

Fixes https://github.com/NuvioMedia/NuvioTV/issues/1585

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
@@ -255,10 +255,10 @@ enum class AutoSkipSegmentType(val storedValue: String) {
         fun fromStoredValue(value: String): AutoSkipSegmentType? =
             values().firstOrNull { it.storedValue == value }
 
-        fun fromSkipIntervalType(type: String): AutoSkipSegmentType? = when (type.lowercase()) {
-            "op", "mixed-op", "intro" -> INTRO
+        fun fromSkipIntervalType(type: String): AutoSkipSegmentType? = when (type.trim().lowercase()) {
+            "op", "opening", "mixed-op", "intro" -> INTRO
             "recap" -> RECAP
-            "ed", "mixed-ed", "outro" -> OUTRO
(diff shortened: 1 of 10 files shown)


haaihond:
Test APK:
https://drive.google.com/file/d/1y8YTaN3G1YcJJj4X1AXHPZVUbeU3Zj13/view?usp=sharing

harrydbarnes:
I was also looking at this. This works for me. My idea was to have a way to do it globally, as you have done, or per show. So you can have an option for Once or Always on the skip intro toast.

haaihond:
> per show. So you can have an option for Once or Always on the skip intro toast.

I haven't given that any though. Sounds like a good idea. This can be hooked into the same logic, as this current PR.

I tested this implementation thoroughly already. I hope Tap will approve these changes soon. I'd prefer adding this addition, after this current PR is merged, instead of adding it to the current one. This would also allow you to add this, if you'd like.

harrydbarnes:
> > per show. So you can have an option for Once or Always on the skip intro toast.
> 
> I haven't given that any though. Sounds like a good idea. This can be hooked into the same logic, as this current PR.
> 
> I tested this implementation thoroughly already. I hope Tap will approve these changes soon. I'd prefer adding this addition, after this current PR is merged, instead of adding it to the current one. This would also allow you to add this, if you'd like.

Sounds great, happy to iterate after this has come in!

haaihond:
@harrydbarnes Just to let you know it has come in. Feel free to iterate.

PR #1593, opened 2026-04-28
feat: add subtitle background color control to in-player style panel

### LINKED ISSUE #1567: [Feature]: Transparency for Subtitiles.
### Area (tag)

Subtitles

### Problem statement

It would be a great feature if the users can set transparency for subtitles as per their choice.
Watching on OLED really make the subtitles pop against dark backgrounds.
Choosing a transparency would make the subtitles blend in with the content better.
I already use this feature with my Kodi setup and it helps immensely. 
Thank you. 

### Proposed solution

Alternatively more neutral darker colour options would be a welcome change too.

No, this is only a proposal

_No response_

_No response_

afda1:
## Summary

Exposes the existing `SubtitleStyleSettings.backgroundColor` field in the in-player `SubtitleStyleSidePanel`. The data model, DataStore key, and persist method all already existed, but the UI to control it while watching was missing.

A new **Background** card appears in the style panel with color swatches (black, navy, dark grey, white) and an opacity stepper, matching the existing Text Color section pattern.

## PR type

## Why

The subtitle background color was already stored in `SubtitleStyleSettings` and wired up in `PlaybackSettingsScreen`, but was never exposed in the in-player panel. Users had no way to add a background behind subtitles without leaving the player. Fixes #1567.

## Policy check

## Testing

- Opened in-player subtitle style panel
- Verified Background card appears in column 1 below Bold
- Selected each background color swatch and confirmed color applies
- Used opacity stepper to set 0%–100% and verified background renders correctly
- Confirmed Reset Defaults resets background to transparent (already handled by existing reset logic)

Panel gains a new Background card with color swatches and opacity stepper, visually identical in style to the existing Text Color card.

## Breaking changes

No breaking changes. New `OnSetSubtitleBackgroundColor` event added to `PlayerEvent`; all existing events and persisted settings are unaffected.

## Linked issues

Fixes #1567

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerPlaybackEvents.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerPlaybackEvents.kt
@@ -1012,6 +1012,9 @@ fun PlayerRuntimeController.onEvent(event: PlayerEvent) {
         is PlayerEvent.OnSetSubtitleTextColor -> {
             scope.launch { playerSettingsDataStore.setSubtitleTextColor(event.color) }
         }
+        is PlayerEvent.OnSetSubtitleBackgroundColor -> {
+            scope.launch { playerSettingsDataStore.setSubtitleBackgroundColor(event.color) }
+        }
         is PlayerEvent.OnSetSubtitleBold -> {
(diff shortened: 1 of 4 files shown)


skoruppa:
Any screenshots? :)

TedCreator:
@afda1 Does this add the ability to select black/darker subtitle colors? Or contribute to the implementation of that feature? Right now we have white and off white.

PR #1596, opened 2026-04-28
prototype: add essential experience mode

tapframe:
## Summary

Prototype pass for a simplified “essential” experience mode, with a lighter first-run path, reduced settings surface, and add-on-focused setup flow. advanced mode still keeps the full existing experience.

this is intentionally unfinished and mostly for discussion: what should be included in essential, what should stay advanced only, and whether the onboarding/settings split feels right. users can switch between essential and advanced later from settings.

## PR type

- Approved larger change

## Why

we’ve discussed internally that the app can feel a bit heavy up front, especially for users who mainly want to install addons and start watching. this explores a simpler default path without removing the advanced controls.

## what’s included

- new first-run experience mode selection: essential or advanced
- essential mode defaults to the modern layout
- essential addon setup screen when no addons are installed
- addons only phone/qr management for essential mode
- simplified settings surface in essential mode
- essential playback settings for common playback, subtitle, audio, and p2p options
- ability to switch from essential to advanced in settings
- existing advanced mode keeps the full settings/layout/addon management surface

## Policy check

discussed internally.

## Testing
- ran essential manual testing

## Breaking changes

none expected. this is a prototype and may change based on review/discussion.

## Linked issues

none


harrydbarnes:
I like the idea as is. I think for those who just want to install and crack on, this is the way. A lot of alternatives also just have settings in the background without choosing options during onboarding.

skoruppa:
- **Integrations** should be also present in Essential Mode. Even though it is a bit advanced this is something that most users will want to do. We don't want to force them to enabled Advanced Settings 
- Same situation with **Layout**. I don't know if we want to hide it and force user to modern view. UI adjustment is something that most users want to do and what Nuvio is praised for. Maybe if we don't want to show all options we should at least leave **Home Layout** (I also think extended posters could be turned on by default, without trailers)?
- The trailer button should be turned on by default 

I didn't notice that earlier but turning P2P from setting should have same warning as turning it on from Streams Screen

PR #1617, opened 2026-04-30
Remove periods from refresh addons subtitles

Stuey-81:
## Summary

Removes trailing periods from the Refresh Addons subtitles so they match the existing Addons screen row style.

## PR type

Small UI text consistency cleanup.

## Why

Existing Addons screen subtitles do not use trailing periods, so this keeps the new Refresh Addons row consistent with the surrounding UI.

## Policy check

- This PR is cosmetic-only / UI text consistency.
- This PR is small in scope and focused on one issue.

## Testing

Text-only change. Confirmed diff only removes trailing periods from the Refresh Addons subtitles.

Not needed for this text-only consistency cleanup.

## Breaking changes

None.

## Linked issues

None.


ram130:
Looks good.

Issue #1619, opened 2026-04-30
[Feature]: only show preferred languages in subtitle list


harrydbarnes:
### Area (tag)

Playback

### Problem statement

For some content there is an overwhelming amount of subtitles, so it would be good to only show selectable the ones that a user would need.

### Proposed solution

Use the existing language preferences settings on top of a new option to only show preferred languages in the selection list

### Are you planning to implement this yourself?

Yes, but I understand implementation still needs maintainer approval

PR #1621, opened 2026-04-30
Feat hide unpreferred languages subtitles

harrydbarnes:
## Summary

Adds a new `hideUnpreferredLanguages` setting to `SubtitleStyleSettings` allowing users to hide subtitle options that don't match their chosen preferred/secondary languages. Includes UI toggle in settings and filtering logic in the overlay.

## PR type

## Why

For some content there is an overwhelming amount of subtitles, so limiting those shown on screen helps reduce visual clutter.

## Policy check

## Testing

Generated apk, tested setting off/on and video playback whilst selecting options in the list

New settings and before and after of same content options (ignore the reflections please!):
PXL_20260430_133735044 MP
PXL_20260430_124828710 MP
PXL_20260430_124753975 MP

## Breaking changes

Nothing changed from my testing. Subtitle use and playback worked fine.

## Linked issues

Addresses: https://github.com/NuvioMedia/NuvioTV/issues/1619


harrydbarnes:
@skoruppa here is one part of the subtitle work as mentioned in other discussion. Only additional comment is should I auto change the addon subtitle loading method to be preferred languages? As is, this feature works.

app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
```diff
@@ -121,7 +121,8 @@ val AVAILABLE_SUBTITLE_LANGUAGES = listOf(
 data class SubtitleStyleSettings(
     val preferredLanguage: String = "en",
     val secondaryPreferredLanguage: String? = null,
-    val size: Int = 120, // Percentage (50-200)
+    val showOnlyPreferredLanguages: Boolean = false,
+    val size: Int = 100, // Percentage (50-200)
```
skoruppa: you should not change default size

harrydbarnes: Will revert later

skoruppa:
> should I auto change the addon subtitle loading method to be preferred languages

I think so. Of course, don't change if user has "Fast"

harrydbarnes:
@skoruppa done! Will change if new setting is on, and then change back to 'all' if was on that prior to the setting being on, if it's turned off.

PR #1624, opened 2026-04-30
Feat/plugin process isolation

### LINKED ISSUE #1604: [Bug]: Plugins make my firestick4k crash when i click on any movies
### Pre-flight checks

- [ ] I can reproduce this on the latest release or latest main build.

### App version / commit

Beta 0.6.9 Latest

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Fire TV Stick 4K

### Android version

Fire TV Stick 4K

### Player mode

Internal player

### Area (tag)

Streams / Sources (selection, loading, errors)

### Steps to reproduce

When i open the app and click on any movies , and   all the plugins loads simultaneously and when i click on any plugins streams it loads then crash, feels like too many plugins is making it crash,

### Expected behavior

it should play just fine as it plays the addons streams i have , like the cartoon add on, when i play , it works fine 

### Actual behavior

idk but i feel like the plugins just make it very unstable at the moment, 
let alone cloudstream extension, they dont even show up

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

Also, the plugins never work on the mobile version,  please help make the tv version work

halibiram:
## Summary

Isolate the JavaScript scraper runtime into a separate `:plugin` Android process and harden it against the failure modes that have been crashing the host app, killing scraper batches mid-execution, or returning empty results on Android TV. The host (UI/playback) process never executes plugin JS or HTTP work anymore — crashes, OOM kills, native leaks, or runaway DOM parses inside a third-party scraper now stay confined to the worker process and the user-facing app keeps running.

## PR type

- Bug fix
- Approved larger change 

## Why

Several user-visible problems all traced back to the JS scraper runtime running in the same process as the UI/player.

**1. App crashes from plugin code.** A buggy plugin or one pulling in a giant HTML page through a redirect chain could OOM the main process, taking the player and home screen down with it. With the runtime relocated to `:plugin`, the worst case is now an empty result list and a process restart on the next batch — no UI impact.

**2. `lmkd` kills during scraping on low-RAM TVs.** Redirect-heavy scrapers (e.g. `hubcloud`, `UHDMovies`) issued 200–300 fetches that each eagerly decoded the response body into a JS string AND inlined the whole body into a JSON bridge payload. That produced hundreds of MB of large-object allocations and triggered Android's `lmkd` swap thrasher, which then killed either the scraper job or the host app. Moving body decoding behind a lazy bridge plus bounding the DOM caches keeps the worker's heap small enough that `lmkd` no longer evicts it.

**3. TV cert chain failures.** TCL Android TV firmware ships an outdated CA bundle that rejects Cloudflare ECC and ISRG Root X2 chains, breaking `api.themoviedb.org` and other modern TLS endpoints from inside the plugin runtime. The main `OkHttpClient` already uses a trust-all `SSLContext`; the plugin runtime now matches.

**4. Stuck worker after a kill.** Once `lmkd` killed the `:plugin` process, the host kept holding stale `Messenger` references and every subsequent scraper just timed out until the user restarted the app. The host now rebinds on demand after process death.

**5. Memory not returned between batches.** Even with the worker isolated, OkHttp connection pools and QuickJS arenas were lingering across scraper sessions. The host now releases the binding once a batch finishes and no requests are in flight, letting the OS reclaim the worker process.

The PR is six focused commits on top of `dev`:

1. **`feat(plugin): isolate JS scraper runtime in :plugin sub-process via IPC`** — new `PluginRuntimeService` runs in a dedicated `:plugin` process, bound from the host via `Messenger`. Scraper requests cross the boundary as `MSG_EXECUTE` / `MSG_RESULT` / `MSG_ERROR` / `MSG_CANCEL` bundles. `RemotePluginRuntimeClient` handles bind/unbind, request correlation, and cancellation propagation. Cap of 10 concurrent scrapers is enforced both host-side (semaphore) and worker-side (fixed thread pool of `Thread.MIN_PRIORITY` daemons).
2. **`fix(app): skip Hilt injection in :plugin sub-process`** — the Hilt graph is host-only; the worker process boots a minimal `Application` to avoid duplicate DataStore / Coil / etc. initialization in `:plugin`.
3. **`fix(plugin): rebind on-demand after :plugin process death`** — `RemotePluginRuntimeClient` detects `onBindingDied` / `onNullBinding` and lazily rebinds on the next request instead of leaving every subsequent scraper hanging.
4. **`plugin: release :plugin process after each scraper batch`** — new `releaseIfIdle()` unbinds the service when no requests are pending, wrapped in `NonCancellable`. The worker process exits and returns its native arenas / OkHttp pools to the OS between batches.
5. **`plugin: trust-all TLS + lazy fetch body`** — plugin `OkHttpClient` now uses the same trust-all `X509TrustManager` + permissive hostname verifier as the main `NetworkModule` so TV firmwares with stale CA bundles can still hit Cloudflare ECC / ISRG Root X2 endpoints. Fetch responses no longer inline the body: bytes are stored in a per-execution `LruCache(cap=32)` and the bridge JSON only carries `__bodyId` + `__bodyLen`. The JS wrapper decodes on demand via `__native_fetch_body_text`, caches the decoded string in its closure, and exposes a `release()` method that frees the native bytes if the body was never consumed. `text()` / `json()` stay idempotent.
6. **`plugin: bound DOM caches, gate hot-path logs, drop redundant code read`** — `documentCache` / `elementCache` switched from unbounded `ConcurrentHashMap` to `LruCache(cap 8 / 8192)`. `console.log`, `console.debug` and fetch start/result `Log.d` calls are gated behind `BuildConfig.DEBUG`. `__native_fetch_body_text` warns when an id has been evicted from the response-body cache instead of silently returning `""`. `PluginManager` no longer reads the scraper code on the host side (the worker re-reads via `codePath`), saving ~100 KB IO and ~1 MB allocation per scraper invocation.

## Policy check

- [x] This PR is small in scope and focused on one problem. *(plugin runtime stability)*

## Testing

Manual on a TCL Android TV (the firmware that originally exposed the cert-chain failures and the `lmkd` kill loop). Loaded a NuvioTV repo with 10 scrapers including `UHDMovies`, `hubcloud`, and one deliberately throwing scraper to verify isolation, then ran scraper batches against several TMDB ids (movies and multi-season shows). Verified:

- Host process PID stays the same across batches; only the `:plugin` PID changes between idle releases (`adb shell ps -A | grep com.nuvio.tv`).
- Killing the worker mid-batch via `adb shell am kill com.nuvio.tv:plugin` no longer hangs subsequent scrapers; the host rebinds on the next request.
- A scraper that throws does not affect the player, the home screen, or other scrapers in the same batch.
- `api.themoviedb.org` and Cloudflare-fronted scraper hosts are reachable; previously these failed with "Chain validation failed" on this TCL firmware.
- `dumpsys meminfo com.nuvio.tv:plugin` after a batch confirms the worker process exits and RSS is released back to the OS once `releaseIfIdle()` runs.
- For redirect-heavy scrapers (`hubcloud`, `UHDMovies`), `dumpsys meminfo` peak heap during a batch is now bounded — the previous unbounded `ConcurrentHashMap` could hold 30+ Jsoup `Document`s (1–3 MB each) until plugin exit; LruCache cap of 8 keeps the documents working set at ~6–24 MB worst case.
- Concurrent playback during scraping no longer stutters; the host process is no longer competing with QuickJS interpretation or hundreds of MB of JSON serialization.
- Logcat in release builds is much quieter — hot-path `Log.d` calls in fetch and `console.log/debug` are dead-code-eliminated under `BuildConfig.DEBUG`.

Build: `./gradlew :app:compileFullDebugKotlin` and `./gradlew installFullDebug` succeed; only pre-existing unrelated warnings remain. Existing unit tests pass.

Not applicable — no UI changes.

## Breaking changes

None.

The `PluginRuntime` API gained no new required parameters and `RemotePluginRuntimeClient` is a drop-in replacement at every call site. The only manifest change is a new `PluginRuntimeService` declared in the `full` flavor `AndroidManifest.xml` with `android:process=":plugin"`.

## Linked issues

fixes #1604

### DIFF
--- a/app/src/full/AndroidManifest.xml
+++ b/app/src/full/AndroidManifest.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <service
+            android:name="com.nuvio.tv.core.plugin.ipc.PluginRuntimeService"
+            android:exported="false"
+            android:process=":plugin" />
+    </application>
+
+</manifest>
--- a/app/src/full/java/com/nuvio/tv/core/plugin/PluginManager.kt
+++ b/app/src/full/java/com/nuvio/tv/core/plugin/PluginManager.kt
@@ -7,6 +7,7 @@ import com.nuvio.tv.core.plugin.cloudstream.tvTypeFromString
 import com.nuvio.tv.core.plugin.cloudstream.ExternalExtensionLoader
 import com.nuvio.tv.core.plugin.cloudstream.ExternalExtensionRunner
 import com.nuvio.tv.core.plugin.cloudstream.ExternalRepoParser
+import com.nuvio.tv.core.plugin.ipc.RemotePluginRuntimeClient
 import com.nuvio.tv.data.local.PluginDataStore
 import com.nuvio.tv.domain.model.ExternalPluginEntry
 import com.nuvio.tv.domain.model.LocalScraperResult
@@ -65,6 +66,7 @@ private const val MANIFEST_SUFFIX = "/manifest.json"
 class PluginManager @Inject constructor(
     private val dataStore: PluginDataStore,
     private val runtime: PluginRuntime,
+    private val remoteRuntime: RemotePluginRuntimeClient,
     private val pluginSyncService: com.nuvio.tv.core.sync.PluginSyncService,
     private val authManager: com.nuvio.tv.core.auth.AuthManager,
     private val externalRepoParser: ExternalRepoParser,
@@ -781,9 +783,14 @@ class PluginManager @Inject constructor(
         episode: Int?
     ): List<LocalScraperResult> {
         return try {
-            val code = dataStore.getScraperCode(scraper.id)
-            if (code.isNullOrBlank()) {
-                Log.w(TAG, "No code found for scraper: ${scraper.name}")
+            val codeFile = dataStore.getScraperCodeFile(scraper.id)
+            if (!codeFile.exists() || codeFile.length() == 0L) {
+                Log.w(TAG, "No code file found for scraper: ${scraper.name} (${codeFile.absolutePath})")
+                return emptyList()
+            }
+            val code = withContext(Dispatchers.IO) { codeFile.readText() }
+            if (code.isBlank()) {
+                Log.w(TAG, "Empty code for scraper: ${scraper.name}")
                 return emptyList()
             }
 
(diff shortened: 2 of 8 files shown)


halibiram:
We don't need this anymore since the necessary optimizations were done in QuickJS

Issue #1589, opened 2026-04-28
[Feature]: long press subtitle icon to toggle subtitles on/off


harrydbarnes:
### Area (tag)

Subtitles

### Problem statement

Turning subtitles off requires the full menu UI showing up on screen. Turning it back on to what you had on/language you prefer requires the same menu

### Proposed solution

If subtitles are on, long pressing subtitles icon turns them off. If off, long pressing turns them back onto previous selected subtitle or if not on previously selects based on users language audio preferences. Additionally, update the Subtitle icon to be the Material Icon for Subtitles, not Closed Caption, and utilise the Off state to indicate when active/not active.

### Are you planning to implement this yourself?

Yes, but I understand implementation still needs maintainer approval

PR #1626, opened 2026-04-30
Subtitles long press and icon updates

harrydbarnes:
## Summary

Users can now long-press the subtitles button to quickly toggle subtitles on and off. To improve user feedback, the icon dynamically switches between Subtitles and SubtitlesOff based on the active state.

More details:

Long-press subtitle control now toggles subtitles:
- If subtitles are active, it disables them and remembers the active internal/addon selection.
- If subtitles are off, it restores the last enabled selection when possible.
- If nothing restorable exists, it opens the subtitle selector like a normal click.

Subtitle button now uses Material Icons:
- Icons.Default.Subtitles when active.
- Icons.Default.SubtitlesOff when inactive.

Long-press key handling consumes the release event so it won’t also fire the normal click.

## PR type

## Why

Without a quick toggle option by long pressing select, there isn't an unobtrusive way of quickly toggling subtitles off and on. This can be useful for if a particular scene requires them, someone is making noise in the house and you need to hear or you need to mute and put them on. Likewise, useful to turn off if you don't want them on but usually do. 

The icon update is necessary to enable this feature, and also the previous icon was not technically correct as not all subtitles are CC.

## Policy check

## Testing

Built APK, tested playing content with subtitles, long pressing icon to turn off, noticing icon update, long pressing to turn back on. Verified single click still works.

Before:
PXL_20260430_215235238 MP

https://streamable.com/hgh66y

After: 
PXL_20260430_215205647 MP

https://streamable.com/5xtcm3

## Breaking changes

Nothing broken in testing elements affected; all okay.

## Linked issues

Implements: https://github.com/NuvioMedia/NuvioTV/issues/1589

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
@@ -278,6 +278,7 @@ class PlayerRuntimeController(
     internal var pendingAudioSelectionAfterSubtitleRefresh: PendingAudioSelection? = null
     internal var rememberedTrackPreference: TrackPreference? = null
     internal var persistedTrackPreference: TrackPreference? = null
+    internal var lastEnabledSubtitleSelection: RememberedSubtitleSelection? = null
     internal var pendingEngineSwitchTrackPreference: PendingEngineSwitchTrackPreference? = null
     internal var explicitSubtitleSelectionForEngineSwitch: ExplicitSubtitleSelectionForEngineSwitch? = null
     internal var effectiveSubtitleSelectionForEngineSwitch: ExplicitSubtitleSelectionForEngineSwitch? = null
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerPlaybackEvents.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerPlaybackEvents.kt
@@ -713,6 +713,14 @@ fun PlayerRuntimeController.onEvent(event: PlayerEvent) {
                 ) 
             }
         }
+        PlayerEvent.OnToggleSubtitlesFromLongPress -> {
+            logSwitchTrace(
+                stage = "event-toggle-subtitles-long-press",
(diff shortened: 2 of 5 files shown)


kernexshadow:
Is this really useful? To turn on and off the subtitles?

kernexshadow:
I would prefer a customization. You could choose the action. For instance open subtitles delay, turn on/off subtitles, etc.

harrydbarnes:
> Is this really useful? To turn on and off the subtitles?

Yes, I'd argue. I've given examples of why in the feature proposal.

harrydbarnes:
> I would prefer a customization. You could choose the action. For instance open subtitles delay, turn on/off subtitles, etc.

I think for now this is the most obvious use of long pressing subtitles. If there were any tracking analytics of what setting users use, I'd add more in as an option, yes.

kernexshadow:
I never use the option of turning on/off subtitles because, I use always subtitles since I am not English country related. Most people who English is not their first language turning off subtitles is useless. And subtitle delay is more useful.

harrydbarnes:
> I never use the option of turning on/off subtitles because, I use always subtitles since I am not English country related. Most people who English is not their first language turning off subtitles is useless. And subtitle delay is more useful.

My partner works from home so sometimes I may watch something with subtitles on low volume, or whilst we are watching something in foreign language trying to understand the words better. There are use cases for it! Likewise, I've never had to adjust the subtitle delay. Raise an issue for it and it can be a further development of this potentially?

harrydbarnes:
@skoruppa my other subtitle improvement idea at the moment, let me know your thoughts on this one? Quick benefit I think. I can add a setting to alter what long pressing subtitles button does, if you like.

harrydbarnes:
@skoruppa any thoughts on this one when you have a moment? Thanks

skoruppa:
@harrydbarnes I think this is a quastion to @tapframe as this changes current behavior

harrydbarnes:
okay! @tapframe let me know what you think. There is a suggestion of a setting to handle what action is taken, happy to implement that if approval exists of the request.

tapframe:
Nice addition, but this can be revisted later. Not planning on merging this for now.

harrydbarnes:
> Nice addition, but this can be revisted later. Not planning on merging this for now.

All good. When you do revisit, let me know anything you'd like changing.

PR #1635, opened 2026-05-01
fix: recover play button focus after async state updates on detail sc…

halibiram:
## Summary

Recover play button focus on the detail screen when late-arriving async state updates (`nextToWatch`, `mdbListRatings`) cause structural recomposition that invalidates the focus node.

## PR type

- Bug fix

## Why

When the detail page opens, focus is correctly set on the play button. However, the ViewModel fires several async observers (`calculateNextToWatch`, `loadMDBListRatings`, etc.) that resolve ~100–200ms after the initial focus grant. These state changes trigger structural recomposition inside `AnimatedVisibility` (e.g. play button text changes from "Play" to "Play S2 E3", or `MDBListRatingsRow` becomes visible), which invalidates the existing focus node and causes focus to silently disappear.

## Policy check

## Testing

- Manually tested on Android TV device by navigating to detail screens for both movies and series.
- Verified focus stays on the play button after `nextToWatch` and `mdbListRatings` arrive.
- Verified trailer autoplay is not affected (guard: `!isTrailerPlaying`).
- Verified returning from playback with `pendingRestoreType` set does not trigger duplicate focus requests.

N/A — no visual change, focus behavior fix only.

## Breaking changes

- None

## Linked issues

- None

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsScreen.kt
@@ -1358,6 +1358,20 @@ private fun MetaDetailsContent(
         }
     }
 
+    // Re-request hero focus when late-arriving async state (nextToWatch, mdbListRatings)
+    // causes structural recomposition that invalidates the focus node.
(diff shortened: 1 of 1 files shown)


halibiram:
Since the root cause has been found, I'm closing this workaround

PR #1189, opened 2026-03-31
Add optional post-play Trakt rating prompts

konvar:
## Summary

- add optional post-play Trakt rating prompts for movies and episodes
- add a compact TV-friendly 10-star Trakt rating selector in the player
- add matching default-rating configuration in Trakt settings
- preload an existing Trakt rating so users can keep it or change it

## PR type

- Approved larger change (link approval below)

## Why

This change adds an optional Trakt rating flow after playback finishes, so users who connected Trakt in the app can rate titles without leaving the TV experience.

Context:
- Feature request / discussion: #1188

Note:
- Per `CONTRIBUTING.md`, this is a larger feature / UX change.
- This PR should stay in draft until a maintainer explicitly approves it in #1188.

## Policy check

- [x] This PR is not cosmetic-only, unless it is a translation PR.
- [x] This PR does not add a new major feature without prior approval.
- [x] This PR is small in scope and focused on one problem.
- [x] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

Manual:
- built a debug APK and tested on Android TV / Google TV
- connected a Trakt account from inside the app
- verified the rating prompt appears after playback completion
- verified the prompt uses the existing Trakt rating when one already exists
- verified the user can change the rating and submit it
- verified the user can dismiss the prompt
- verified the default rating setting updates the preselected value

Automated:
- `./gradlew assembleDebug --console=plain`
- `./gradlew testDebugUnitTest --tests com.nuvio.tv.data.repository.TraktRatingServiceTest --console=plain`

## Screenshots / Video (UI changes only)

Added screenshots of:
- the post-play Trakt rating popup
- the Trakt settings default rating selector

## Breaking changes

None

## Linked issues

Related to #1188

PR #1656, opened 2026-05-01
feat(tv): add Android TV Continue Watching preview channel

### LINKED ISSUE #886: [Feature]: Home Screen Channel - Continue Watching
### Area (tag)

Watch Progress

### Problem statement

Would it be possible to integrate the "Continue Watching" channel like Stremio and other apps?

### Proposed solution

On the GoogleTV home screen and Projectivy, you can add the "continue watching" channel from various streaming apps. This let's you pickup where you left off watching directly on your home screen. Stremio and other streaming apps have this feature integrated.  Would it be possible to integrate this for Nuvio as well?  

Even in beta, app is becoming nearly my daily driver. This is the only feature holding me back from using it as a primary app.

No, this is only a proposal

_No response_

https://github.com/spocky/miproja1/issues/436#issuecomment-4075079097

harshrs21:
Closes #886

## Summary

**Continue Watching preview channel for Android TV / Projectivy launcher**

- Adds an Android TV launcher "Continue Watching" preview channel that surfaces in-progress content on the home screen (tested with Projectivy launcher on Fire TV)
- Channel tiles display backdrop/poster art with a title-logo badge overlay (`COLUMN_LOGO_URI`), duration, progress bar, and season/episode metadata
- TMDB fallback when the primary addon returns no artwork for an item
- Logo/artwork hydration for both watch-progress and library items, with duration hydration for items previously missing runtime
- Stale Trakt metadata (entries with `runtimeMs == 0` from before runtime hydration was added) is now refreshed automatically; addon runtime takes precedence over cached values
- Fixes `BuildConfig.DEBUG` guard for the in-app updater prompt (debug buildType sets `isDebuggable = false`, so the existing check never fired; switched to the custom `IS_DEBUG_BUILD` field)

**Duration fix — movies showing incorrect 1hr progress bar**

- Removed hardcoded 1hr fallback that fired for Trakt-sourced movies: Trakt API returns `progressPercent` but not playback duration, so the fallback triggered every time after the optimistic local entry expired. Channel now omits duration/position entirely rather than displaying a fake bar, and clears stale values on each UPDATE
- Added TMDB runtime fallback in `TraktProgressService`: when the primary addon returns no runtime for a movie, the TMDB details response (already fetched for artwork) now also supplies `runtimeMinutes`, giving the channel an accurate duration and progress bar

**Channel sync reliability**

- Switched `reconcile()` from `UPDATE` to `DELETE + INSERT` for existing program rows: launchers that cache tiles on insert (e.g. Projectivy) now redraw tiles with updated progress immediately rather than serving stale data until the next home refresh
- Added `TvChannelRefreshJobService` — a `JobScheduler` periodic job (15 min, persisted across reboots) that reconciles the channel when the app is not in the foreground, so progress from other devices appears without requiring the TV app to be opened
- Added `TvChannelBootReceiver` handling `BOOT_COMPLETED` and `ACTION_INITIALIZE_PROGRAMS` to trigger an immediate reconcile after device boot or launcher channel initialisation
- Added Nuvio Sync periodic pull (every 5 min) in `WatchProgressRepositoryImpl` so progress from other devices appears without an app restart when Trakt is not the active source

## Test plan

- [x] Channel appears in Projectivy launcher after first app launch
- [x] Tiles show backdrop art with logo badge overlay on focus
- [x] Progress bar reflects accurate watch position and duration for Trakt and Nuvio Sync sources
- [x] Trakt movies with no addon runtime show correct duration via TMDB fallback
- [x] Removing an item from Continue Watching removes it from the channel within ~300 ms
- [x] Trakt items with previously stale duration now show correct runtime
- [x] Progress changes from another device (phone → TV) appear in the channel within ~15 min without opening the app
- [x] Channel repopulates after device reboot without opening the app
- [x] In-app updater prompt does not appear on debug builds
- [x] Channel survives app restart and profile switch
- [x] Channel is a no-op on non-leanback (phone/tablet) devices


konvar:
#1189 Hey, you may also want to take a look at my PR. It has some related work around Google TV Home integration, Trakt rating prompts, and Trakt / Continue Watching changes, so there might be some logic or implementation details you can copy or reuse here.
https://github.com/konvar/NuvioTV/tree/all-implementation-in-one

app/src/main/java/com/nuvio/tv/data/repository/WatchProgressRepositoryImpl.kt
```diff
@@ -222,11 +231,17 @@ class WatchProgressRepositoryImpl @Inject constructor(
...
 
+        val episodeRuntimeMs = episodeMeta?.runtimeMs ?: 0L
+        val runtimeMs = episodeRuntimeMs.takeIf { it > 0 } ?: metadata.runtimeMs
+
         return progress.copy(
             name = if (shouldOverrideName) metadata.name ?: progress.name else progress.name,
             poster = progress.poster ?: metadata.poster,
             backdrop = backdrop,
             logo = progress.logo ?: metadata.logo,
+            duration = if (progress.duration > 0) progress.duration
```
halibiram: duration priority here is opposite to what was fixed in TraktProgressService.kt, Is there a specific reason you used the opposite logic here

ram130:
Looks good

harryposter:
This is great to have, thank you for your work! However, I am seeing only just the one show I am currently watching out of the 29 I have in continue watching and it looks really weird.
Can we have all CW items showing in the channel please? And clear logos would be nice too as I don't use titles on PL.
Thanks!
Projectivy Launcher_20260506_133123

PR #1672, opened 2026-05-02
perf(home): reduce recompositions during D-pad navigation

wingoku:
## Summary

Refactor Modern Home screen state management to reduce unnecessary recompositions during D-pad navigation. Previously, reading dynamic navigation state (active indices, focus keys, metadata URLs) in the body of `ModernHomeContent` and `ModernHomeRowsList` triggered top-down re-evaluations on every focus move.

Key changes:
- Convert navigation delegated properties (`by`) to raw `MutableState` objects to isolate reads from Composable function bodies
- Refactor metadata parameters (trailer URLs, expansion states) from raw values to lambdas/State objects to defer reads to individual item level
- Stabilize row-level callbacks and `isActiveRow` checks via remembered lambdas so `ModernRowSection` remains skippable
- Decouple horizontal scroll state from vertical focus restoration to prevent horizontal movements from invalidating the parent LazyColumn
- Extract `ModernHomeRowsList` composable to isolate the rows list from the hero/backdrop state
- Added Compose Recomposition Highlighter toggle to Debug Settings (visible only in debug builds)
- Fix `GridHomeContent` compile error from `item()` calls outside `LazyGridScope`

## PR type

- Bug fix

## Why

D-pad navigation on the home screen caused full-tree recompositions, leading to frame drops and jank on lower-end Android TV devices. The unstable state reads at the top level invalidated the entire composition tree on every focus change.

## Policy check

## Testing

- Manual testing with Compose Recomposition Highlighter enabled to verify reduced recomposition counts on D-pad navigation
- Verified home screen scrolling, focus transitions, hero backdrop updates, and trailer autoplay still function correctly
- Build verified with `compileFullBenchmarkKotlin`

no visual changes, performance-only improvement.

## Breaking changes

Nothing should break

## Linked issues

performance improvement - no issues reported


wingoku:
image

skoruppa:
> image

Can you show how it looks after?

halibiram:
build is failing because ScrollStateRegistry.kt seems to be missing from the commit. Can you check this?

wingoku:
merge conflict issue. Let me fix the imports quickly

halibiram:
https://github.com/user-attachments/assets/83f94657-b41b-4e83-9b62-0f413e79f266

I just added a basic version of the missing file to make it compile, but something isn't working quite right. I'm not sure if the issue is caused by that missing file

wingoku:
image

halibiram:
@wingoku  Could you open a separate PR for the RecompositionHighlighter util? It might be a good idea to put it under 'debug' in the settings. This way, we can better see the difference between the normal version and your changes

skoruppa:
> It might be a good idea to put it under 'debug' in the settings. This way, we can better see the difference between the normal version and your changes

Separate PR is not needed but I agree. I already discussed this and moved the RecompositionHighlighter setting to debug options. 

New build for testing purpose (it is a "relese build" but with IS_DEBUG flag set to true so the RecompositionHighlighter option is present to let you quickly check recompositions) - http://morigal.pl/upload/app-full-universal-benchmark-PR1672.apk

skoruppa:
@wingoku issues introduced by this PR (not present in upstream/dev):

1. If you go quickly from CW one item right, then few items down, after focus stops the hero content will show CW item you went into at the beginning. Such issue does not exist on develop 
2. Focus restoration shifts row horizontally: When navigating right within a row, then down, then back up - focus lands on the first item (index 0) instead of the previously focused item, causing the row to scroll back to the start. The focusedItemByRow map appears to not be consulted correctly by the focusRestorer during vertical navigation
3. Hero does not update when returning to CW row: After scrolling down to catalog rows and then back up to Continue Watching, the hero backdrop/content remains frozen on the last catalog item instead of updating to the focused CW item. The CW item's onFocused debounce does not fire (or is suppressed) when focus returns via vertical navigation

so although the performance improvement is nice I can't merge this yet

skoruppa:
I noticed one more issue - the lazy row loader (the one that triggers catalog fetches for shimmer/placeholder rows as they become visible) could get "stuck." It would stop requesting new data for rows that scroll into view, leaving them as permanent shimmer placeholders. Navigating into an item and back would unstick it because the whole screen re-initializes

skoruppa:
I resolved all those issues:

- **Focus restoration jumping to index 0** - switched the row-level focusRestorer to a lambda that reads the actual last-focused index at invocation time instead of a stale value captured at composition.
- **Hero not updating when returning to CW row** - the hero settle delay constants were bumped way too high (250–320ms vs 90–170ms on develop), making the hero appear frozen. Restored them to match develop.
- **focusHolder getting out of sync** - the onActiveRowKeyChange/onActiveItemIndexChange callbacks weren't updating the focusHolder object, so subsequent recompositions could reset active row/item to stale values.
- **Lazy loader getting stuck** - the catalog load trigger was reading a stale carouselRows closure instead of the latest state, so newly visible placeholder rows never got their fetch kicked off.

Also merged upstream develop to bring the branch up to date.

Everything works correctly now, but I'm not 100% sure the changes don't negatively impact performance (especially the lambda-based focusRestorer which evaluates on every focus-restore event rather than being pre-computed). Definitely needs more testing on device - particularly rapid navigation and scroll smoothness.

skoruppa:
New test build: http://morigal.pl/upload/app-full-universal-benchmark-PR1672-fix.apk

wingoku:
Thanks @skoruppa for the hotfixes. 

The mutable states are causing the frequent recompositions. The way I was trying to get around it is by using derivedStates and lambdas. Having that reverted will cause frequent recompositions.

In compose the rule of thumb is to minimize mutable states and specially when they're tied to the Compose views.

One odd thing that i observed in the app is

scrolling horizontally on one row recomposes the other row which isn't being interacted with at all. This shouldn't happen. As a first step to resolve it this PR is submitted but unfortunately there are some bugs which need to be ironed out. Perhaps we can try patching those bugs without using mutable states.

On the note of higher delays for backdrop, We generally don't want herobackground to be constantly updated and/or image to be decoded when scrolling. Therefore I bumped that up to higher numbers, this can be brought down a bit but reducing it to the original values will have regressive perf impacts

In my testing, I found the current process of identifying focus and its position might be a bit hacky and contributing to perf regressions. I don't have the full background insight into why this particular process was implemented but I hope we employ and implement a more performant and robust focus solution

wingoku:
I tested the build you shared and it's a step increase in performance over the current official release. Although some of the perf gains are lost due to putting back the mutable states in UI for fixing the bugs which is understandable. I think it provides a solid foundatoin for further fixes going forward.

Thanks for testing and hotfixing the bugs @skoruppa

wingoku:
@skoruppa I'm wondeirng why we're prefetching 8 items instead of 3-4.

Once a row is populated, upon scroll/focus position N-2 where N is currently the last visible item in the horizontal row, we should start prefetch for 3-4 items.

Anything more then that is a serious CPU and Memory bottleneck which leads to frame drops and lag on mainstream GoogleTV devices. Nvidia shield and similar device can brute force their way through bigger prefetches, but mainstream can't.

This rule also applies to vertical rows. Only one row below current row should be prefetched and not more. 

Perhaps we can add this optimization to this commit and merge the PR>

wingoku:
Nice thanks for changing the prefetching.

Can you please share a build here

skoruppa:
Yeah, I think you made good points. Here is the build http://morigal.pl/upload/app-full-universal-benchmark-PR1672-3.apk

PR #1676, opened 2026-05-03
auto-play with source filters and locale not applying on language change

skoruppa:
## Summary

Fix two bugs:
1. **Auto-play stream selection fails when source scope/addon filters are configured** - `autoSelectTriggered` was set to `true` even when no matching stream was found, preventing retries when the filtered addon responds later.
2. **Language change not applied until full app restart** - `LocaleCache` wasn't updated before `activity.recreate()`, and Modern Home resolved strings from stale Application context with cached old-locale titles.

## PR type

- Bug fix

## Why

Users reported that configuring "auto-play source scope" or "allowed addons/plugins" caused auto-play (regex/first stream) to always fall back to manual source selection. Additionally, changing the app language in settings didn't reflect on Modern Home section titles ("Kontynuuj Oglądanie", type suffixes) until a full process restart.

## Policy check

## Testing

- Configured auto-play mode to FIRST_STREAM with source scope INSTALLED_ADDONS_ONLY and a single selected addon. Verified stream is auto-played after the addon responds (even if slower than timeout or other sources).
- Configured auto-play with multiple addons, selected only one in allowed addons. Verified the system retries auto-select when the allowed addon responds after timeout.
- Changed language from Polish to English in settings. Verified Modern Home immediately shows "Continue Watching" and correct type suffixes (Movies/Series) without needing a full app restart.
- Verified switching back to system language also applies immediately.

behavioral fix, no UI layout changes.

## Breaking changes

Nothing should break

## Linked issues

Reported on Discord


halibiram:
Mind doing a minor touch on the Play button? It's currently ignoring app language

Issue #1637, opened 2026-05-01
[Feature]: Soft gradient fade at top of rows (Modern layout) instead of hard cut


tsoltan:
### Area (tag)

UI / Layout

### Problem statement

When scrolling down through the rows, the top edge of the scrolling row has a hard horizontal boundary. As posters move upward toward that edge, the bottom portion of them remains fully visible right up until it hits the line and then disappears instantly. There's no gradual transition - one frame that visible slice is there at full opacity, and the next it's gone.
It's not a clipping issue per se, more that the cutoff is too "binary." The eye keeps tracking the lower part of the poster all the way up, and then it just pops out of view, which makes the whole scroll feel less smooth than it could.
This only happens on scroll down. Scrolling up doesn't have the same effect.

### Proposed solution

Add a soft alpha gradient along the top edge of the row so the visible portion of the poster gradually fades out as it approaches the boundary, instead of staying at 100% opacity until the very last pixel and then vanishing. Or maybe you see a better option. Or maybe you have/know a better approach.

### Are you planning to implement this yourself?

No, this is only a proposal

PR #1687, opened 2026-05-03
Perf Home Screen: Reduce recompositions significantly to improve responsiveness & frame loss

wingoku:
## Summary

Decomposed Complex Composables into atomic ones and enhanced performance by stabilizing lambda and refining prefetching

1.
Removed BoxWithConstraints: Replaced with more efficient screen size lookups to eliminate subcomposition during scrolling.
2.
Decomposed ModernHomeContent: Broke down the massive composable into smaller sections (ModernHeroSection, HeroInfoSection) to isolate recompositions and improve performance.

3.
Preserved Focus Logic: Ensured all existing focus and navigation logic remains unchanged and robust.

## PR type

## Why

Several recompositions are happening on home screen leading to lag and stutter. Current composables are also complex and need to be atomic. BoxWithConstraint can perf issues due to subcompositions, removing that. 
## Policy check

- [ ] This PR is not cosmetic-only, unless it is a translation PR.
- [ ] This PR is small in scope and focused on one problem.
- [ ] If this is a larger or directional change, I linked the **approved** feature request issue below.

## Testing

Home screen

After:
image

## Breaking changes
None found so far


halibiram:
I tested the optimizations in this PR. Compared to the latest commit on the dev branch, I got these values in my measurements: vertical scroll improved from 38fps -> 42fps, and horizontal scroll from 42fps -> 44fps

halibiram:
I observed some stuttering while long press scrolling from left to right on the CW row. focus calculation might not be keeping up. I think slowing down the scroll speed just a bit, specifically for the CW row, could fix this

wingoku:
Thanks Halil. Focus calculation needs to be rethought. The current approach is quite resource intensive and complicated. This will remain to be a bottleneck going forward.

A simple approach could be to keep the lastSelectedIndex in the data class per row and look at how other gridView composables are handling focus finding.

We should not introduce artificial delays to mask the issues. If focus calc isn't able to catch up to optimized composables then let's improve focus.

halibiram:
Focus is very fragile; even a minor change breaks it immediately. I had struggled with this, but I see that you're quite experienced in these matters. It would be great if you could take a look at this too

tsoltan:
Maybe in scope of such changes you can take a look at https://github.com/NuvioMedia/NuvioTV/issues/1637

skoruppa:
All changes included in #1699

PR #1690, opened 2026-05-03
fix: enable seamless frame rate switching for trailer player

obahj:
Trailers on devices locked to 23.976Hz (e.g., NVIDIA Shield) exhibited visible judder because the trailer ExoPlayer was configured with VIDEO_CHANGE_FRAME_RATE_STRATEGY_OFF, preventing any display refresh rate adaptation.

Switch to VIDEO_CHANGE_FRAME_RATE_STRATEGY_ONLY_IF_SEAMLESS so the platform can seamlessly adjust the display refresh rate to match the trailer frame rate when supported, eliminating judder without causing disruptive black screen flickers.

## Summary

Changed `VIDEO_CHANGE_FRAME_RATE_STRATEGY_OFF` to `VIDEO_CHANGE_FRAME_RATE_STRATEGY_ONLY_IF_SEAMLESS` in `TrailerPlayer.kt` (1 line).

## PR type

- Bug fix

## Why

On devices with a fixed 23.976Hz display mode (e.g., NVIDIA Shield), trailer playback exhibits visible judder/stutter. The main player already has full AFR (Auto Frame Rate) support, but the trailer player's ExoPlayer had frame rate switching explicitly disabled, causing a mismatch between the trailer's frame rate (typically 24fps/30fps from YouTube) and the display refresh rate.

## Policy check

## Testing

Tested on NVIDIA Shield Pro set to 23.976Hz. Trailer playback is now smooth without visible judder. No black screen flicker occurs during trailer transitions since the seamless strategy only switches when the platform supports non-disruptive mode changes.

## Breaking changes

None

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/components/TrailerPlayer.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/components/TrailerPlayer.kt
@@ -88,7 +88,7 @@ fun TrailerPlayer(
                 .build()
             ExoPlayer.Builder(context)
                 .setLoadControl(loadControl)
-                .setVideoChangeFrameRateStrategy(C.VIDEO_CHANGE_FRAME_RATE_STRATEGY_OFF)
+                .setVideoChangeFrameRateStrategy(C.VIDEO_CHANGE_FRAME_RATE_STRATEGY_ONLY_IF_SEAMLESS)
                 .build()
                 .apply {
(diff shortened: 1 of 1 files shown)


ram130:
Hmm. Shouldn't the seamless option also be added to the main player? We only have frame rate matching and not seamless support.

PR #1698, opened 2026-05-04
Feat/plugin duplicate detection

halibiram:
## Summary

Replace the videasy-specific risky-enable confirmation with a generic duplicate-scraper detection modal on the Plugins screen. The modal groups enabled scrapers that share a `name` (case-insensitive) or `filename` across repositories, picks one to keep enabled (alphabetic by scraper id), and offers a one-tap "Disable N" to switch off the rest. It opens the first time the Plugins screen has scrapers loaded, after a new repository is added, and again whenever a scraper goes from disabled to enabled (so toggling a repository off/on resurfaces it).

Two commits, in order:

- Remove the videasy `PluginSafety` gating, the `plugin_risky_enable_*` strings (16 locales), the pending-enable UI state and the `ConfirmScraperEnableDialog`. `PluginManager.downloadScrapers` now respects `info.enabled` from the manifest as the default for new scrapers instead of force-disabling videasy.
- Add `PendingDuplicateResolution` / `DuplicateGroup` / `DuplicateEntry` to `PluginUiState`, a union-find duplicate scan in `PluginViewModel` debounced 750 ms (so the per-repo emits during initial scraper download coalesce into a single comprehensive check), and `DuplicateResolutionDialog` rendered as a `Popup` next to the existing `ConfirmRepoChangesDialog`. The dialog list shows the keep entry in `titleSmall` Success green with the repository name appended in `bodySmall` `TextTertiary`, and each duplicate entry on its own line in the same compact "scraper - repository" form so multiple groups fit without scrolling.

## PR type

## Why

The videasy gating in `PluginSafety` existed to default-disable a scraper that crashed inside the QuickJS native library on certain content. That underlying crash was fixed in [#1685](https://github.com/NuvioMedia/NuvioTV/pull/1685) by rebuilding the QuickJS AAR, so the special-case gating is no longer protecting against anything and was just filtering a legitimate provider out of the manifest. While removing it, the more common real-world failure mode became visible: users with multiple repositories often end up with two or three copies of the exact same scraper enabled at once (same `name`, same `filename`, different repository), which causes redundant scraper calls and inconsistent stream lists. The new modal surfaces those duplicates and lets the user resolve them in one step instead of hunting through the list manually.

## Policy check

## Testing

Manual on an Android TV device (`./gradlew installFullDebug` against a TV):

- First open of the Plugins screen with three duplicate scrapers across two repos (SezonlukDizi, Webtelzle, FilmModu in `MoOnCrOwN-NUVİO` and `Erenxyz`) shows the modal with three groups and a `Disable 3` action; screenshot captured.
- `Disable 3` flips the three Erenxyz copies to disabled, scrapers list updates, modal dismisses, no re-trigger from the disable emits.
- `Later` dismisses the modal; toggling a repository off then on causes the same modal to reappear because the previously-disabled scraper ids re-enter the enabled set. Toggling individual scrapers off does not re-trigger.
- Adding a fresh duplicate repository while the screen is open causes the modal to open after the scraper download settles (the 750 ms debounce coalesces the per-repo emits so the check sees the full final list).
- Read-only profiles (non-primary using primary plugins) skip the duplicate scan entirely.
- `./gradlew :app:compileFullDebugKotlin` passes (only the pre-existing `@Inject` annotation warning).

tv_screenshot

## Breaking changes

None. `PluginSafety` and the `plugin_risky_enable_*` strings are removed because nothing else referenced them. Existing per-scraper `enabled` state is preserved by `PluginManager.downloadScrapers` (it still falls back to `existingScraper?.enabled` first), so users who had the videasy scraper manually enabled keep it enabled, and users who had it disabled keep it disabled. The new state fields and events on `PluginUiState` / `PluginUiEvent` are additive.

## Linked issues

Builds on top of [#1685](https://github.com/NuvioMedia/NuvioTV/pull/1685) (QuickJS AAR rebuild that resolved the videasy crash and made the `PluginSafety` gating obsolete). No tracking issue for the duplicate-detection modal — surfaced during internal multi-repository testing.


halibiram:
Looks like this approach isn't getting much like . I'm closed PR

PR #1738, opened 2026-05-07
fix(subtitles): prefer normal embedded subtitles over forced/CC

guuilp:
## Summary

When a release has multiple embedded subtitle tracks per language (e.g. forced, CC, normal), the auto-selection now prefers the normal one. Previously it picked whichever track came first in the file, which was often forced.

## PR type

- Bug fix

## Why

Some anime releases ship with 3 embedded subtitle tracks per language: forced (songs & signs), CC (closed captions), and normal (dialogue). The auto-selection was picking the first matching track, which was often the forced one. Users had to manually switch to the normal subtitle every time.

## Policy check

## Testing

Tested with anime releases containing 3 embedded subs per language across 7+ languages. Verified:
- Normal track is selected over forced and CC on fresh play
- CC is preferred over forced when no normal track exists
- PT/ES regional variant tie-breaking still works correctly
- MPV engine switch (auto internal engine) re-evaluates auto-selection when embedded tracks are discovered after an addon fallback

## Screenshots
image

## Breaking changes

No breaking changes


guuilp:
Replaced by new PR with rebased branch onto main

PR #1082, opened 2026-03-26
Fix forced subtitle selection logic (match by audio language + label)

craitose:
## Summary

A forced subtitle is selected only if:

It is marked as forced (isForced = true), and

It matches the audio language by:

language code (e.g., "en"), or

language name inside the label (e.g., "English [Forced]")

## PR type

- Bug fix

## Why

The previous implementation relied on the generic auto‑subtitle pipeline, which:

Always tries to pick something

Falls back to the first forced track

Does not consider audio language when selecting forced subtitles

## Policy check

- [x] This PR is not cosmetic-only, unless it is a translation PR.
- [x] This PR does not add a new major feature without prior approval.
- [x] This PR is small in scope and focused on one problem.
- [x] If this is a larger or directional change, I linked the issue where it was approved.

## Testing

Scenarios:

1.Audio: JA, Forced: JA + EN → selects Forced JA

2.Audio: JA, Forced: EN only → selects none

3.Audio: JA, no forced tracks → selects none

4.Audio: JA, Forced: JA → selects Forced JA

5.Audio: EN, Forced: ES only → selects none

## Breaking changes

No Breaking changes

## Linked issues

Fixes #662
Fixes #1200

PR #1801, opened 2026-05-10
feat: refactor forced subtitles

harrydbarnes:
## Summary

Users now set subtitle behaviour with two separate controls:

- Preferred Language: the normal subtitle language, e.g. English.
- Use Forced Subtitles: a toggle that only changes auto-selection behaviour around forced tracks.

Scenarios:

#### Preferred Language = English, Use Forced Subtitles = off

- Auto-selects normal non-forced English subtitles when available.
- Forced subtitles are not specially preferred.

#### Preferred Language = English, Use Forced Subtitles = on

- If selected audio is English:
  - Selects English forced subtitles if available.
  - If no English forced subtitles exist, selects nothing.

- If selected audio is not English:
  - Selects normal non-forced English subtitles.

#### Preferred Language = None, Use Forced Subtitles = off

- Selects no subtitles.

#### Preferred Language = None, Use Forced Subtitles = on

Forced-only mode, but only when the selected audio matches the user's resolved preferred audio language.

- If selected audio matches the resolved preferred audio language:
  - Selects a forced subtitle that matches the selected audio language.
  - If no matching forced subtitle exists, selects nothing.
  - Never falls back to normal subtitles.

- If selected audio does not match the resolved preferred audio language:
  - Selects no subtitles.

Example:

- Device language is English.
- Preferred audio is Device language.
- Preferred subtitle language is None.
- Use Forced Subtitles is on.
- The selected source only has French audio, so playback selects French audio.

Result:

- The app selects no subtitles.
- It does not select French forced subtitles, because French audio does not match the user's resolved preferred audio language.

Existing users who had Forced selected are migrated to `Use Forced Subtitles = on`.

- If they had a secondary subtitle language set, that becomes the preferred subtitle language.
- Otherwise they default to English plus the new forced behaviour.

This also preserves existing `Show Only Preferred Languages` behaviour: when Preferred Language is None but a secondary subtitle language is set, the secondary language is still used for filtering visible addon subtitles. The forced-subtitle audio-language target is layered on top only when the new forced-subtitle rules allow it.

## PR type

## Why

Previously, Forced was treated like a subtitle language option. That caused bad matches because “forced” is a track type, not a language.

Example issue:

- User selects Forced.
- Video audio auto-selects English.
- Available forced subtitle is marked as French forced, or just has forced in the label.
- The app could select that forced subtitle even though it does not match the actual audio language.

Another issue:

- User prefers original audio.
- Default subtitles are English.
- A video auto-selects Japanese audio.
- The old forced preference could still chase forced tracks instead of choosing useful English subtitles.

Now the app uses the actual selected audio track, not just the user’s configured audio setting. That matters because Original audio can resolve differently per video.

Example:

- User audio setting is Original audio.
- Subtitle preference is English.
- Use Forced Subtitles is on.

Results:

- Video’s original audio is English:
  - English forced subtitles are selected if present.
  - Otherwise nothing is selected.

- Video’s original audio is Japanese:
  - Normal English subtitles are selected.

For Preferred Language = None, forced subtitles are also gated by the resolved preferred audio language.

Example:

- Device language is English.
- Preferred audio is Device language.
- Preferred subtitle language is None.
- Use Forced Subtitles is on.
- A selected source only has French audio.

Result:

- The app selects no subtitles.
- It does not select French forced subtitles, because those are intended for French-audio viewers and would not help an English-device-language user.

That keeps forced subtitles for “foreign dialogue within my audio language” cases, while still giving full subtitles when the media audio is in another language.

Language matching is also stricter now: forced subtitle matching uses language codes as bounded tokens, so short codes like `en` do not accidentally match unrelated labels such as `French`.

## Policy check

## Testing

- Built and tested APK on a few sources to see selection logic in action, and to see the new settings plus migration work.
- Reviewed the affected Kotlin code paths for settings migration, subtitle settings UI, auto-selection logic, language matching, and preferred-language filtering.
- Verified `Forced` is no longer offered as a subtitle language option in the settings dropdowns.

Settings Before and After:
WhatsApp Image 2026-05-10 at 9 08 16 PM

WhatsApp Image 2026-05-10 at 9 08 16 PM (1)

## Breaking changes

None during testing. Existing `Forced` subtitle-language preferences are migrated to the new setting.

## Linked issues

Addresses: https://github.com/NuvioMedia/NuvioTV/pull/1082


oathzed:
Sounds like what we want. Good job Harry

harrydbarnes:
> Sounds like what we want. Good job Harry

Thanks!

PR #1805, opened 2026-05-11
skip full avatar catalog load on startup

harrydbarnes:
## Summary

Skips loading full avatar catalog on app open

- Normal app startup: skip avatar catalog unless the sidebar/profile chip needs an avatarId resolved.
- User opens avatar/profile selection: ProfileSelectionViewModel loads the avatar catalog then.
- If it was already loaded earlier: AvatarRepository has an in-memory cachedCatalog, so the profile screen can reuse it instead of fetching again.

## PR type

## Why

Previously the app fetched the full avatar catalog on first composition every time. Now it only fetches that catalog if the active profile needs an avatarId lookup and does not already have a direct avatarUrl. That removes an avoidable startup network call for profiles that do not need it.

## Policy check

## Testing

Built out apk and tested, no issues on my end

No visible UI changes to user

## Breaking changes

Nothing broken in my testing

## Linked issues

None, based on an idea to optimise initial loading times

### DIFF
--- a/app/src/main/java/com/nuvio/tv/MainActivity.kt
+++ b/app/src/main/java/com/nuvio/tv/MainActivity.kt
@@ -312,15 +312,19 @@ class MainActivity : ComponentActivity() {
             }
 
             var avatarCatalog by remember { mutableStateOf(emptyList<com.nuvio.tv.data.remote.supabase.AvatarCatalogItem>()) }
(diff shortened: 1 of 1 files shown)


harrydbarnes:
Whoops, deleted in error trying to clean things up my end. Still good to merge.

Issue #2004, opened 2026-05-21
[Bug]: Infinite retry loop when source-level extractor fails after first frame renders


LostInTyping:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

Beta 0.6.19

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Raspberry Pi 5

### Android version

Android 16

### Player mode

Internal player

### Area (tag)

Playback (start/stop/buffering)

### Steps to reproduce

1. Play a Matroska stream that triggers a source-level extractor error after the first frame renders (reproduced with *Parks and Recreation* S6E20 & S6E16 via a debrid source whose MKV trips Media3 1.8.0's EBML varint parser during cue-point indexing).
2. Let the stream fail and watch the player's automatic retry behavior.
3. Observe the player retries indefinitely with no error dialog ever shown.

### Expected behavior

The player retries up to the configured retry budget (2 attempts), exhausts it, and shows the existing "Playback Error" dialog with the "Go Back" button.

### Actual behavior

The player retries, renders one frame, resets the retry budget via `resetErrorRetryState()` in `onRenderedFirstFrame`, fails again at the same byte offset, retries again, renders one frame, resets the budget again, and loops forever. The error dialog is never shown because the budget never exhausts.

Logcat shows `Auto-retry 1/2` and `Auto-retry 2/2` cycling indefinitely.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

- Series: *Parks and Recreation* S6E20 / S6E16
- The MKV triggers Media3 1.8.0's MatroskaExtractor "No valid varint length mask found" error during cue-point indexing, after the first video frame has already been rendered.

### Logs (required for crash reports)
Sanitized logcat snippet showing the retry loop (repeats indefinitely):
```shell
onRenderedFirstFrame -> resetErrorRetryState() (budget back to 0/2)
MatroskaExtractor: No valid varint length mask found
onPlayerError -> Auto-retry 1/2
onRenderedFirstFrame -> resetErrorRetryState() (budget back to 0/2)
MatroskaExtractor: No valid varint length mask found
onPlayerError -> Auto-retry 1/2
```
repeats forever until user manually backs out.

[...]

PR #2005, opened 2026-05-21
fix(player): gate retry budget reset behind stable playback

LostInTyping:
## Summary
When MatroskaExtractor (or any source-level extractor) fails after the first frame has rendered, `resetErrorRetryState()` in `onRenderedFirstFrame` resets the retry budget on every player rebuild, creating an infinite retry loop. This PR gates the reset behind 5 seconds of continuous healthy playback so the budget can actually exhaust and surface the existing error dialog.

## PR type
- [x] Behavior bug/regression fix

## Why
The retry budget bounds repeated failures on the same stream. The bug: the budget was cleared every time a single video frame rendered, which happens after every re-prepare. For source-level parser errors that fail again at the same byte offset (Media3 1.8.0 MatroskaExtractor's "No valid varint length mask found" is the concrete case), the player retried, rendered one frame, got the budget reset, failed again, and looped endlessly with no error dialog ever shown.

The fix defers the reset until the player has been in `STATE_READY && isPlaying` continuously for 5 seconds. `onPlaybackStateChanged` and `onIsPlayingChanged` (re)arm or cancel the timer on state transitions, so any unhealthy interval (buffering, paused, error) restarts the window.

## Issue or approval
#2004 - reported and reproduced by me on my own Android TV device. Provider/source-dependent, only fires on certain files where Media3 1.8.0's EBML parser trips on cue-point indexing.

## UI / behavior impact
- [x] No UI change

## Policy check
- [x] This PR is small, focused, and limited to one problem.
- [x] This PR is not cosmetic-only.
- [x] Any UI change fixes a linked glitch/bug and includes visual proof, or this PR has no UI change.
- [x] Any behavior change fixes a linked bug/regression or has explicit approval, or this PR has no behavior change.
- [x] This PR does not add dependencies, architecture changes, migrations, or product-direction changes without explicit approval.

> UI polish, cosmetic-only changes, minor behavior tweaks, and unapproved product changes will be closed without review.

## Scope boundaries
Intentionally not in this PR:

- Fixing the upstream MatroskaExtractor varint bug (lives in Media3 1.8.0; a version bump would be a larger dependency change).
- Recovery via "retry from position 0 on detected cue-parse failure" (separate WIP branch for potential follow-up).
- Loosening auto-engine-switch to fire on post-first-frame errors (currently gated to startup-only by design).
- Any UI or copy changes to the error dialog.

Four files touched, all in `app/src/main/java/com/nuvio/tv/ui/screens/player/`:

- `PlayerRuntimeController.kt`, one new `Job` field for the deferred-reset timer.
- `PlayerRuntimeControllerErrorRecovery.kt`, three new helpers (`scheduleStableProgressReset`, `cancelStableProgressReset`, `refreshStableProgressResetGate`) and one delay constant.
- `PlayerRuntimeControllerInitialization.kt`, wires the helpers into existing `onRenderedFirstFrame`, `onPlayerError`, `onPlaybackStateChanged`, and `onIsPlayingChanged` callbacks.
- `PlayerRuntimeControllerLifecycle.kt`, cancels the timer on `releasePlayer`.

## Testing
Built `installFullDebug` on Raspberry Pi 5 (arm64-v8a, Android TV).

Repro stream: Source from *Parks and Recreation* S6E20 / S6E16 served by a source whose MKV trips Media3 1.8.0's EBML varint parser during cue-point indexing.

- **Before fix**, player retried forever, no error dialog. Logcat showed `Auto-retry 1/2` and `Auto-retry 2/2` cycling indefinitely as each newly rendered frame reset the budget.
- **After fix**, same two retry attempts logged, then the existing "Playback Error" dialog appears with "Go Back", the expected terminal-error UX.
- **Regression check on a normal stream**, plays end-to-end without interference. The 5-second timer arms once after the first frame, fires, resets the budget, no further intervention.

## Screenshots / Video
Not a UI change.

## Breaking changes
None.

## Linked issues
#2004

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeController.kt
@@ -362,6 +362,7 @@ class PlayerRuntimeController(
     internal var errorRetryCount: Int = 0
     internal var consecutiveAutoPlayCount: Int = 0
     internal var errorRetryJob: Job? = null
+    internal var stableProgressResetJob: Job? = null
     internal var currentScrobbleItem: TraktScrobbleItem? = null
     internal var currentTraktEpisodeMapping: EpisodeMappingEntry? = null
     internal var currentTraktEpisodeMappingKey: String? = null
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerErrorRecovery.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerErrorRecovery.kt
@@ -2,6 +2,7 @@ package com.nuvio.tv.ui.screens.player
 
 import android.util.Log
 import androidx.media3.common.PlaybackException
+import androidx.media3.common.Player
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.datasource.HttpDataSource
 import com.nuvio.tv.R
@@ -12,6 +13,7 @@ import kotlinx.coroutines.launch
 private const val MAX_STARTUP_AUTO_RETRIES = 2
 private const val MAX_AUTO_RETRIES = 2
 private const val RETRY_DELAY_MS = 1_500L
+private const val STABLE_PROGRESS_RESET_DELAY_MS = 5_000L
 
 internal fun PlayerRuntimeController.showRecoveryOverlay() {
     _uiState.update { state ->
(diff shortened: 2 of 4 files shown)


skoruppa:
@halibiram @kernexshadow any thoughts?

halibiram:
> @halibiram @kernexshadow any thoughts?

@LostInTyping explained the issue to me yesterday, and I asked him to open a PR. Your solution looks great, this will completely fix that infinite loop issue with the player. Great work

PR #2101, opened 2026-05-27
Fix: Multi-provider debrid cache checking

Beary-Handsome:
## Summary

When multiple debrid providers are configured (e.g. TorBox + Premiumize), the app only cache-checks torrents against the single "active/preferred" provider. If that provider doesn't have a torrent cached, it's marked NOT_CACHED and hidden — even if another configured provider has it cached and ready to play.

**Symptoms:**
- User has TorBox + Premiumize configured, TorBox is preferred
- A torrent is cached on Premiumize but not TorBox
- App marks it NOT_CACHED (only checked TorBox) → stream is hidden
- User sees fewer playable streams than expected

### LocalDebridAvailabilityService.kt
- `cacheCheckAccount()` (single) → `cacheCheckAccounts()` (all configured providers with LocalTorrentCacheCheck)
- `annotateCachedAvailability()` checks all providers in parallel via `coroutineScope { async {} }`
- Merges results: cached on ANY provider = CACHED, prefers activeResolver when tied
- `isCached()` returns true if ANY provider has it

### DirectDebridResolver.kt
- `shouldResolveToPlayableStream()`: allows resolution for any provider with a configured API key (was restricted to activeResolverProviderId)
- `resolveLocalTorrentStream()`: uses the provider that confirmed the cache hit
- Cache key no longer gated by activeResolverProviderId

### DebridStreamPresentation.kt
- `isInactiveResolverStream()`: hides streams from unconfigured providers only, not non-preferred ones

## Backwards Compatibility
Single-provider users see identical behavior.

## Test Plan
- [ ] Configure TorBox + Premiumize, set TorBox as preferred
- [ ] Open a title cached on Premiumize but not TorBox
- [ ] Verify stream shows as cached and resolves via Premiumize
- [ ] Verify single-provider setup still works

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/debrid/DebridStreamPresentation.kt
+++ b/app/src/main/java/com/nuvio/tv/core/debrid/DebridStreamPresentation.kt
@@ -52,7 +52,9 @@ class DebridStreamPresentation @Inject constructor(
 
     private fun Stream.isInactiveResolverStream(settings: DebridSettings): Boolean {
         val streamProviderId = DebridProviders.byId(clientResolve?.service)?.id ?: return false
-        val activeProviderId = settings.activeResolverProviderId ?: return false
-        return isDirectDebrid() && streamProviderId != activeProviderId
+        if (!isDirectDebrid()) return false
+        // Only hide streams from providers the user hasn't configured,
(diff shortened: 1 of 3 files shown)


coltondick:
Can this be merged?

PR #2160, opened 2026-06-01
Revert "perf(player): use webview for hardware-accelerated subtitle rendering"

skoruppa:
Reverts NuvioMedia/NuvioTV#2154


skoruppa:
@halibiram it needs more work. Although it uses hardware acceleration, subtitles are now not centered anymore.

Issue #2142, opened 2026-05-31
[Bug]: Vision Profile 7 Handling Option Auto selects wrong HDR Format


DeadGazeS:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.7.1

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Fire Tv Cube 2nd Gen

### Android version

Android 9

### Player mode

Internal player

### Area (tag)

Other

### Steps to reproduce

1. Select any movie/show
2. Select a Dolby Vision stream
3. Observe in Settings what Dolby Vision Profile 7 Handling Option Auto did

### Expected behavior

If your display has got no capabilities to play Dolby Vision, HDR should be selected. Otherwise depending on your devices capabilities either DV7, DV8, DV8.1 etc

### Actual behavior

One of my TVs is not able to output Dolby Vision HDR format (it's a Samsung tv). The app correctly sees that, but still chooses the wrong HDR Format (Dolby Vision, even though it can't show Dolby Vision).
I can still watch the movie/ show, but HDR is not turned on. 

<img width="2194" height="1719" alt="Image" src="https://github.com/user-attachments/assets/e4a74304-80de-44f5-b4a8-576cbb6b0f93" />

### Frequency

Always

### Did this work before?

No, it never worked

### Logs (required for crash reports)

```shell

```

ram130:
Looking into it.

DeadGazeS:
> Looking into it.

Amazing. Thanks

Issue #1244, opened 2026-04-05
[Feature]: Switch to disable HDR10+ or Dolby Vision metadata


cowballz:
### Area (tag)

Playback

### Problem statement

I want to be able to switch on or off the HDR10+ or Dolby Vision metadata. This fixes issues with HDR not being displayed properly for some devices/TVs

### Proposed solution

Switch to enable or disable HDR/Dolby metadata 

### Are you planning to implement this yourself?

No, this is only a proposal

guyyagil:
that would be amazing !

ram130:
Will be in new update. Please close once you test.

Issue #687, opened 2026-03-08
[Feature]: Prefer HDR in hybrid DV/HDR files.


sins07:
### Area (tag)

Playback

### Problem statement

Firestick 4k Max 2021 has a known issue where DV/HDR hybrid files show a black screen instead of playing the video.

### Proposed solution

Please add an option to prefer HDR in hybrid files.

### Are you planning to implement this yourself?

No, this is only a proposal

ram130:
Are it's not a DV7 issue?

sins07:
its a specifically firestick 4k max 2021 issue.
https://github.com/androidx/media/issues/1895 

"There is an old bug that won't let you watch mkv video files that have dolby vision/hdr10.

Kodi developers was able to create a workaround for their app on Android TV (by disallowing either HDR10+ or DV). Additionally these devices seem to struggle with Profile 7, forcing 8.1 as a hack seems to work as well."

ram130:
Close if fixed.

sins07:
Fixed by stripping dolby vision rpu layer

PR #2167, opened 2026-06-02
Strip Dolby Vision from all streams option

DeclanSC:
## Summary

This Pull Request builds on the amazing work Ramon has been doing and adds an additional toggle option to Strip DV for all streams not just DV7, this means for non DV devices, conversion to 8.1 isn't required and it can strip the DV layer completely, not only leaving the base layer but allowing HDR10+ too. This replicates the same functionality that other players such as Kodi and Vimu use and from my own testing allows me to play every stream without issues.

As DV5 doesn't have a fallback HDR layer, this is not stripped and allows the device to convert to HDR.

## PR type

- [X] Critical bug fix

## Why

For many user with non DV tvs (especially firestick users) streams with DV will fail to play properly and display a black screen with only audio, or return an error. In these cases, currently the only way to play DV7 files is to use the new conversion to DV8.1 and let the device fallback to HDR (HDR10 layer option doesn't work for most devices and is only attempting to ignore DV data not stripping it). 

On firesticks there is also a device specific limitation meaning files with both HDR10+ and DV will experience the black screen and none of the conversion options will allow these to play properly.

## Issue or approval

#2142, #1244,  #687, 

## Reproduction steps

On non-dv devices play a DV7 file without conversion, the stream will error or display a black screen.
On firestick devices, play a hybrid DV/HDR10+ file, the stream will error or display a black screen.

## UI / behavior impact

- [x] No UI change
- [x] No behavior change

## Policy check

## Scope boundaries

None.

## Testing

On two firestick devices I have played a range of different formats and stream types with the new DV strip enabled and with conversion set to Off

- DV7 - Stripped DV layer and played in HDR
- DV8.1 - Stripped DV layer and played in HDR
- DV/HDR10+ hybrid - Stripped DV layer and played in HDR10+
- HDR and HDR10+ only - Played as normal in correct format
- DV5 - Not stripped and device successfully converted and played in HDR
- Non-HDR - Plays as normal

When using a DV tv this option can just be left off and normal conversion can be used.

## Screenshots / Video

Not a UI change

Using Tron as an example in this video [https://gofile.io/d/v0z64C](https://gofile.io/d/v0z64C), HDR10 Layer option does nothing and all 3 streams show a black screen with audio. 

Turning on my new toggle and turning conversion off completely fixes it, and with tv diagnosis on I can see it actually play in HDR10+ for the first stream. 

Currently the only option which makes a few of the non HDR10+ streams play is the convert to 8.1, but obviously converting to a different profile just for my device to fallback to HDR isn't optimal.

## Breaking changes

None.

## Linked issues

#2142, #1244,  #687


haaihond:
@ram130 what do you think?

ram130:
Make it so the toggle can be active when strip is selected and inactive otherwise. 

Or 

Set it be inactive until off (native) is selected. This would prevent others from having a conflicting experience.

DeclanSC:
> Make it so the toggle can be active when strip is selected and inactive otherwise.
> 
> Or
> 
> Set it be inactive until off (native) is selected. This would prevent others from having a conflicting experience.

Yeah I like the idea of only having the toggle active when Off (native) is selected, will prevent any conflicts between the conversions and this stripping. I'll make the update.

DeclanSC:
Have now made the above change so it works in the same way as the DV7 preserve and DV5 to 8.1 toggles, just in this case only shows as active when Dolby conversion is set to Off.

DeclanSC:
I've also amended the HDR10 base layer option's description as this was misleading, it doesn't actually do any stripping or conversion of DV it just attempts to ignore it's metadata for profile 7 which may help on certain devices, but is completely different to the new Strip DV functionality I've added.

skoruppa:
I'm just thinking if this shouldn't be a new option in the "Dolby Vision Profile 7 Handling" under Native?

DeclanSC:
> I'm just thinking if this shouldn't be a new option in the "Dolby Vision Profile 7 Handling" under Native?

My only concern with that is it isn't just for profile 7 and also applies to profile 8 hybrid streams. Happy for it to be a new option in there if we rename the title to be something like Dolby Vision Handling?

DeclanSC:
> I'm just thinking if this shouldn't be a new option in the "Dolby Vision Profile 7 Handling" under Native?

Have now made this change and does look a bit cleaner now, removed Profile 7 from the title and description just so the dropdown is more accurate as not all options relate to 7.

ram130:
> > I'm just thinking if this shouldn't be a new option in the "Dolby Vision Profile 7 Handling" under Native?
> 
> 
> 
> Have now made this change and does look a bit cleaner now, removed Profile 7 from the title and description just so the dropdown is more accurate as not all options relate to 7.

Thanks. Looks good. Merged.

PR #2208, opened 2026-06-06
Fix grid home stealing focus/scroll back to top on launch

### LINKED ISSUE #2207: [Bug]: Grid home steals focus/scroll back to top for a few seconds on launch
### Pre-flight checks
- Searched existing issues; not a duplicate.
- Reproducible on the latest dev build.
- Single bug.
- Not source/stream-specific (happens regardless of catalogs).

### App version / commit
Built from source, dev @ df6f9538

### Install method
Built from source

### Platform
Android TV

### Device model
NVIDIA Shield TV (2019)

### Android version
Android 11

### Area
Focus / Remote navigation

### Steps to reproduce
1. Set the home layout to **Grid**.
2. Cold-launch the app (force stop first), land on the Home screen.
3. While catalogs are still streaming in, immediately try to scroll down.

### Expected behavior
Scrolling down works and the position holds, the same as the Modern and Classic home layouts already behave on launch.

### Actual behavior
For the first few seconds, focus and the scroll position keep getting pulled back to the top each time another catalog row finishes loading. You scroll down, it snaps back up, repeatedly. Once all catalogs have loaded it behaves normally. The Modern and Classic layouts do not have this problem.

### Frequency
Always (cold launch, Grid layout)

### Did this work before?
Not sure

### Notes
Root cause: in `GridHomeContent.kt` the initial-focus `LaunchedEffect` is keyed on `gridItems.size`, which changes on every catalog that streams in. Each change re-requests focus on the hero (top), and `shouldRequestInitialFocus` stays true for the whole session, so it keeps stealing focus/scroll back to the top until loading settles.

The Modern and Classic layouts already avoid this (Classic keys initial focus on stable boolean signals; Modern guards it behind an "active row established" flag). The fix brings Grid in line with them, so all three layouts behave consistently on launch.

Laskco:
## Summary

Fixes the grid Home screen stealing focus and scroll position back to the top for the first few seconds after launch. While catalogs stream in, the user can't scroll down, each newly loaded row yanks focus/scroll back to the top until loading settles.

This brings the **grid layout in line with the modern and classic layouts**, which already behave correctly on launch, so all three home layouts are now consistent.

## PR type

- [x] Critical bug fix

## Why

In `GridHomeContent.kt`, the initial-focus `LaunchedEffect` is keyed on `gridItems.size`, which changes on every catalog that streams in during launch. Each change re-requests focus on the hero (the top item), and `shouldRequestInitialFocus` stays true for the whole session, so focus (and the scroll position) is pulled back to the top repeatedly until all catalogs finish loading.

The classic and modern layouts don't have this bug because they already guard against it. Classic keys initial focus on stable boolean signals, and modern guards it behind an "active row established" flag. This change applies the same protection to grid, matching their behavior.

## Issue or approval

Fixes #2207

## Reproduction steps

1. Set the home layout to **Grid**.
2. Cold-launch the app (force stop first), land on Home.
3. While catalogs are still loading, try to scroll down.
4. Before fix: focus/scroll snaps back to the top each time a row loads, for several seconds. After fix: scrolling down works and holds (same as modern/classic).

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

- Only the grid layout's initial-focus logic is changed.
- Classic and modern layouts are untouched (they already guard against this); this aligns grid with them.
- No new UI, no new settings, no change to where initial focus lands (still the hero on boot).

## Testing

- Device: NVIDIA Shield TV (2019), Android 11. Debug-signed release APK, sideloaded.
- Grid layout, cold launch: confirmed scrolling down immediately now holds instead of snapping back to the top, matching modern/classic.
- Confirmed initial focus still lands on the hero on boot.
- Verified the "scroll to top" sidebar Home action and saved-focus restoration still work.

## Screenshots / Video

Not a UI change (no visual element changed). _Before/after screen recording of the scroll behavior to be attached._

## Breaking changes

None.

## Linked issues

Fixes #2207

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/GridHomeContent.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/GridHomeContent.kt
@@ -34,6 +34,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.withFrameNanos
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.focus.FocusRequester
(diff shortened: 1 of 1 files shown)


Laskco:
## **Before**

https://github.com/user-attachments/assets/ede05a43-ed63-44df-a97a-4d0627cf67fa




## **After**
https://github.com/user-attachments/assets/3ae4880b-99ec-4377-9f19-5f8db65251d8

PR #2237, opened 2026-06-08
fix(player): select highest resolution in adaptive tracks and bypass underreported HLS capabilities

### LINKED ISSUE #2236: [Bug]: Adaptive streams stay on a first/lower resolution instead of the best one
### Pre-flight checks

### App version / commit

0.7.4

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Firestik 4k

### Android version

Latest

### Player mode

Internal player

### Area (tag)

### Steps to reproduce

1. Install this plugin for Vixsrc sources https://raw.githubusercontent.com/fralk/All-in-One-Nuvio/refs/heads/main/manifest.json
2. Open a movie or episode with an adaptive HLS/DASH stream that has multiple video variants (Ex. The Great)
3. When the 480p variant is listed first in the manifest, ExoPlayer starts at 480p and never goes up, when the first stream in the manifest is at 1080p ExoPlayer selects it correctly from the beginning

### Expected behavior

The player should select the best supported adaptive video quality for the current stream and device, or quickly switch up to it when bandwidth is sufficient.

### Actual behavior

Playback starts on the first video resolution and remains there, even when higher quality variants are available and the connection/device should support them.

Image

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

Even if higher resolution variants are available for the same content exoplayer always pick up the first

Manifest stuck at 480p
#EXT-X-STREAM-INF:BANDWIDTH=1200000,RESOLUTION=854x480
#EXT-X-STREAM-INF:BANDWIDTH=2150000,RESOLUTION=1280x720
#EXT-X-STREAM-INF:BANDWIDTH=4500000,RESOLUTION=1920x1080

Manifest that works at 1080p (first resolution at 1080p):
#EXT-X-STREAM-INF:BANDWIDTH=4500000,RESOLUTION=1920x1080
#EXT-X-STREAM-INF:BANDWIDTH=1800000,RESOLUTION=1280x720
#EXT-X-STREAM-INF:BANDWIDTH=1080000,RESOLUTION=854x480

```shell

```

### Anything else? (optional)

There is a fix implemented recently but didn't work 
https://github.com/NuvioMedia/NuvioTV/issues/2170

maybe setting higher DefaultBandwidthMeter estimate for
C.NETWORK_TYPE_ETHERNET
C.NETWORK_TYPE_WIFI

as well could solve?

halibiram:
## Summary

This PR migrates the HLS resolution false-negative capability override to a context-aware track selection override using `DefaultTrackSelector.selectAllTracks`. In addition, it fixes resolution reporting issues under adaptive playbacks by selecting the highest resolution representation instead of breaking on the first variant in a track group, and adds real-time resolution tracking on format changes.

Key changes:
1. **Safely Gated Resolution Capability Override:** Overrode `selectAllTracks` in `DefaultTrackSelector` setup in [PlayerRuntimeControllerInitialization.kt](app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt). If the current stream is an HLS source, standard 8-bit AVC or HEVC video formats at or below 1080p that report `FORMAT_EXCEEDS_CAPABILITIES` (due to firmware profile/level underreporting) are upgraded to `FORMAT_HANDLED`.
2. **Correct Resolution Reporting for Selected Tracks:** Refactored the selected video format iteration logic in [PlayerRuntimeControllerTracks.kt](app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerTracks.kt). Instead of stopping at the first selected format (which points to the lowest resolution track in an adaptive HLS track group), the loop now evaluates all selected representations to pick and report the highest resolution/bitrate track.
3. **Dynamic Resolution Tracking:** Implemented `onVideoSizeChanged` in the ExoPlayer listener in [PlayerRuntimeControllerInitialization.kt](app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt) to dynamically update the active video dimensions when layout shifts occur during adaptive playbacks.
4. **JVM Unit Tests:** Created [TrackSelectionInvestigationTest.kt](app/src/test/java/com/nuvio/tv/ui/screens/player/TrackSelectionInvestigationTest.kt) to simulate track selection upgrades on underreported decoders.

## PR type

- [x] Critical bug fix

## Why

1. **Firmware Underreporting:** Many devices (e.g. TCL Smart TV Pro) report level 3.1 support in firmware capabilities, but their hardware can decode higher levels. Third-party HLS manifests declare level 4.0 (`avc1.640028`) for all resolutions. This causes ExoPlayer to evaluate 720p/1080p as `FORMAT_EXCEEDS_CAPABILITIES` and drop back to 480p.
2. **ExoPlayer TrackSelector Target:** Applying this capability upgrade inside `DefaultTrackSelector.selectAllTracks` allows us to selectively target HLS AVC/HEVC 8-bit tracks under 1080p for capability upgrades, preserving honest renderer diagnostics and protecting other playback scenarios (like DASH or local MP4 files where real hardware decoder limits might apply).
3. **Underreported Resolution in UI/Diagnostics:** When an adaptive HLS track group is active, ExoPlayer marks all variants in the group as selected. The previous loop stopped at the first entry, mistakenly reporting low-quality resolutions (like 480p) to the UI and scrobbler even if the player was actively decoding a 1080p stream.

## Issue or approval

Fixes #2236.

## Reproduction steps

1. Play a stream containing HLS `avc1.640028` on a device with underreported decoder levels (e.g. level 3.1).
2. Observe that resolution remains stuck at the lowest quality (e.g. 480p) due to capability checks failing.
3. Verify that the UI/Diagnostics report 480p even when a higher representation is actually decoded.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Changes are restricted to the ExoPlayer initialization (`PlayerRuntimeControllerInitialization.kt`), track updates (`PlayerRuntimeControllerTracks.kt`), and unit tests (`TrackSelectionInvestigationTest.kt`). No UI modifications were made.

## Testing

1. **JVM Unit Tests:** Ran the test suite verifying capability upgrades:
   ```powershell
   ./gradlew :app:testFullDebugUnitTest --tests "com.nuvio.tv.ui.screens.player.TrackSelectionInvestigationTest"
   ```
2. **Manual Verification:** Verified that under HLS playbacks, resolutions dynamically switch to 1080p and are correctly reported to the player UI and scrobbler database instead of showing the low-quality fallback.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #2236.


skoruppa:
Not sure I'm comfortable merging this as-is. The fix makes sense for the specific HLS scenario but I think it casts too wide a net.

The fallback condition here:

```
(format.width <= 1920 && format.height <= 1080 && 
 format.sampleMimeType?.startsWith("video/") == true && 
 format.sampleMimeType != "video/dolby-vision")
 ```
...will force FORMAT_HANDLED for basically any video ≤1080p- VP9, AV1, H.264 High 10, HEVC Main 10, whatever. And since the renderer replaces the default one unconditionally in buildVideoRenderers(), this isn't scoped to HLS at all. DASH, progressive MP4 - everything goes through the same override.

If a device's decoder limitation is actually real (not just underreported firmware), we'll force ExoPlayer to pick a track the hardware can't decode. That's decoder errors or black screen instead of a clean fallback to a lower resolution.

I think this logic belongs in the TrackSelector rather than the renderer. A custom DefaultTrackSelector that overrides selectAllTracks() could patch the rendererFormatSupports array before selection - but only when:
- the current source is HLS (set a flag when building the MediaSource)
- codec is video/avc or video/hevc
- standard 8-bit profile only
- resolution ≤ 1080p

That way the renderer reports capabilities honestly, and the TrackSelector makes the conscious decision to override only in the known false-negative scenario. It's also trivial to toggle off if something goes wrong.

I haven't tested this alternative so take it with a grain of salt, but I think it's worth exploring before we ship something that touches all playback globally

halibiram:
Thanks for the review, you're 100% right here. Overriding this at the renderer level was indeed too broad and risky since it replaced the default renderer unconditionally. Doing so would have forced `FORMAT_HANDLED` for all playbacks like DASH and Progressive MP4, as well as codecs like VP9, AV1, and 10-bit HEVC. On devices with actual physical decoder limitations, this would have definitely caused decoder errors or black screens instead of a clean fallback to a lower resolution.

To fix this safely, I've refactored to use your suggested `DefaultTrackSelector` approach. I implemented it as an anonymous subclass of `DefaultTrackSelector` that overrides `selectAllTracks`. Now, the capability override only kicks in if we are dealing with an HLS source, the codec is standard H.264 or H.265, the resolution is 1080p or lower, and it's a standard 8-bit stream (filtering out 10-bit and HDR profiles since capability limits there are usually physical). 

This allows the renderer to report capabilities honestly for logging and diagnostics, while keeping the bypass safely gated. I also deleted the `NuvioMediaCodecVideoRenderer` class and cleaned up the reflection-based factory setup. The unit tests have been updated and run successfully too. Let me know what you think!

skoruppa:
Much better- the TrackSelector approach addresses my original concerns. The HLS guard, codec restriction to avc/hevc only, and the 10-bit/HDR exclusion make this significantly safer than the renderer override.

Two notes:

1. The `updateAvailableTracks` change (picking highest resolution across all selected tracks instead of breaking on the first one) and the `onVideoSizeChanged` listener are unrelated to the level-underreporting fix. Can you split those out or explain why they're needed here? Keeps the PR focused on the one issue.

2. `rendererFormatSupports` is typed as `Array<out Array<out IntArray>>` - the write to `IntArray` elements works since primitive arrays are mutable regardless of `out`, but worth a quick check that Kotlin doesn't emit a warning there.

Otherwise LGTM.

halibiram:
> 2\. `rendererFormatSupports` is typed as `Array<out Array<out IntArray>>` - the write to `IntArray` elements works since primitive arrays are mutable regardless of `out`, but worth a quick check that Kotlin doesn't emit a warning there.

Checked – since IntArray is a primitive array, the out projection doesn't apply to element writes. Kotlin compiler doesn't emit any warnings here and the build is clean.

halibiram:
> 1. The `updateAvailableTracks` change (picking highest resolution across all selected tracks instead of breaking on the first one) and the `onVideoSizeChanged` listener are unrelated to the level-underreporting fix. Can you split those out or explain why they're needed here? Keeps the PR focused on the one issue.

In adaptive HLS streams, multiple resolution tracks reside within the same track group.

- `updateAvailableTracks` change: The previous track-selection loop logic returned early on the first matched track (which often default-resolved to a lower resolution representation). The update ensures we iterate through the track group to correctly identify and report the actual highest resolution being decoded under the new bypass policy.
- `onVideoSizeChanged` listener: In adaptive scenarios, the decoder layout dimensions shift dynamically. Capturing this listener update keeps the UI and scrobbler resolution stats accurate with the actual rendered frame dimensions instead of stale initial stream metadata.

Both of these changes are directly tied to the level-bypass fix, ensuring that when the player successfully decodes the higher-level (e.g. 1080p) stream, the system correctly identifies and reports it to the diagnostics and player UI instead of underreporting the resolution.

skoruppa:
Can you provide a build for the affected user to let him confirm if it helps?

halibiram:
> Can you provide a build for the affected user to let him confirm if it helps?

Yes, I've done that. I forwarded the build containing our recent changes to the person who reported the issue, and they confirmed that it has been fixed

[#2236](https://github.com/NuvioMedia/NuvioTV/issues/2236#issuecomment-4654930238)

PR #2245, opened 2026-06-08
fix: prevent autoplay from skipping post-credits scenes

### LINKED ISSUE #2243: [Bug]: Autoplay skips post-credits scenes
### Pre-flight checks

### App version / commit

0.7.4

### Install method

GitHub Release APK

### Platform

Google TV

### Device model

Chromecast with Google TV (4K)

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

Next Episode / Auto-play

### Steps to reproduce

1. Play an episode that has a post-credits scene.
2. Let playback run into the outro/credits segment.
3. Observe the "Next Episode" prompt.

### Expected behavior

For an episode with a post-credits scene, the autoplay prompt should hold until the episode truly ends, so the post-credits scene isn't skipped. The "fire as soon as the outro starts" behavior should only apply when the episode has no post-credits content after the outro.

### Actual behavior

When outro timecodes exist for the episode, the "Next Episode" prompt fires as soon as the outro/credits segment starts, regardless of whether there's a post-credits scene after it, so autoplay advances to the next episode and skips the post-credits scene.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

siriusvoid:
## Summary

Fixes autoplay skipping post-credits scenes when outro timecodes are available. With outro data, the next-episode prompt used to fire the moment the outro started, jumping past anything after it. Now, when there's a gap between the outro ending and the actual end of the video, the prompt waits until it finishes.

## PR type

- [x] Critical bug fix

## Why

When a show has outro timecodes and a post-credits scene, autoplay triggered the moment the outro began and skipped right past the scene. This holds the next-episode prompt until the video actually ends whenever it detects a gap after the outro. Playback with no outro data is unchanged and still uses the normal threshold.

## Issue or approval

Fixes https://github.com/NuvioMedia/NuvioTV/issues/2243

## Reproduction steps

1. Play an episode/title that has outro timecodes and a post-credits scene after the outro.
2. With autoplay enabled, let playback reach the outro.
3. Observe that the next-episode prompt fires at the outro start and autoplay advances, skipping the post-credits scene.

Expected: the prompt should hold until the video nearly ends so the post-credits scene is not skipped.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

- Only changes when the next-episode prompt fires; nothing else in autoplay or playback is touched.
- When an outro is detected with a gap over 5s before the video ends, the prompt holds until the video nearly ends (within 1s), treating it as a post-credits scene.
- With no such gap, the prompt still fires when the earliest outro starts, as before.
- With no outro data, it falls back to the existing percentage or minutes-before-end threshold.

## Testing

- Episode with a post-credits scene -> next-episode prompt waits until the video ends.
- Episode with an outro and no gap -> prompt fires when the outro starts, as before.
- Episode with no outro data -> prompt falls back to the threshold setting.
- Tested on Chromecast with Google TV (4K) with a debug build.

## Screenshots / Video

Not a UI change

## Breaking changes

None

## Linked issues

Fixes https://github.com/NuvioMedia/NuvioTV/issues/2243

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerNextEpisodeRules.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerNextEpisodeRules.kt
@@ -36,23 +36,33 @@ object PlayerNextEpisodeRules {
         thresholdPercent: Float,
         thresholdMinutesBeforeEnd: Float
     ): Boolean {
-        val outroInterval = skipIntervals.firstOrNull {
-            it.type == "outro" || it.type == "ed" || it.type == "mixed-ed"
-        }
-        return if (outroInterval != null) {
-            positionMs / 1000.0 >= outroInterval.startTime
-        } else {
+        val outroSegments = skipIntervals.filter { it.type in OUTRO_SEGMENT_TYPES }
+
+        if (outroSegments.isNotEmpty()) {
             if (durationMs <= 0L) return false
-            when (thresholdMode) {
-                NextEpisodeThresholdMode.PERCENTAGE -> {
-                    val clampedPercent = thresholdPercent.coerceIn(97f, 100f)
-                    (positionMs.toDouble() / durationMs.toDouble()) >= (clampedPercent / 100.0)
-                }
-                NextEpisodeThresholdMode.MINUTES_BEFORE_END -> {
-                    val clampedMinutes = thresholdMinutesBeforeEnd.coerceIn(0f, 3.5f)
-                    val remainingMs = durationMs - positionMs
-                    remainingMs <= (clampedMinutes * 60_000f).toLong()
-                }
+            val latestOutroEndMs = (outroSegments.maxOf { it.endTime } * 1_000.0).toLong()
+            val postOutroGapMs = durationMs - latestOutroEndMs
+
+            return if (postOutroGapMs > POST_OUTRO_AUTOPLAY_GAP_MS) {
+                // Post-credits scene: hold prompt until video truly ends.
+                positionMs >= durationMs - END_OF_VIDEO_EPSILON_MS
+            } else {
+                // Fire at earliest outro start if outro ends near file end.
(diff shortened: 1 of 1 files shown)


skoruppa:
After using this for some time, I'm not happy with how the post-credits detection works.

The current POST_OUTRO_AUTOPLAY_GAP_MS of 5s is too small - it incorrectly treats normal endings as having post-credits scenes. And when it does detect one, waiting until the last 1s before showing the next episode card means the countdown never has time to trigger properly.

I'm changing the logic: if the gap between the outro end and the file end is larger than the user's configured next episode threshold, the ending timestamp will not be used as "end of episode" - instead, the regular threshold setting (percentage or minutes before end) will apply. This way power users can set threshold to 99.9% if they really want to catch everything, but for regular users the experience won't be broken by false post-credits detection

siriusvoid:
Fair points, I see how the 5s gap misfires on files where the outro just ends a bit before the actual file end. I originally added this because in anime there is usually a short next episode preview or post-credits scene right after the ED, and autoplay would either skip it entirely or the next episode card would end up covering the subtitles for the whole scene. Your approach is cleaner and still covers that case with a high threshold, so no objections.

Just to confirm with an example: on a 25 minute episode with the threshold set to 99.9%, the next episode prompt would appear in the last ~1.5 seconds, right?

I'll close the mobile PR as well, since it used the same approach. Will the new logic be ported to the mobile version too?

skoruppa:
> Just to confirm with an example: on a 25 minute episode with the threshold set to 99.9%, the next episode prompt would appear in the last ~1.5 seconds, right? 

Yes

I have created #2272. You can port this logic into your mobile PR

siriusvoid:
Thanks for sorting this out, I'll port it into the mobile PR then

PR #2381, opened 2026-06-19
fix(player): harden external auto-next loader dismiss, fast fallback, and season-less anime

Laskco:
For context on what this fixes: external player auto next had a few problems. It would intermittently fail to advance to the next episode, the Loading Next Episode screen could get stuck with no way to back out, a next episode with no instant source left you on that loader instead of the source list, and absolute numbered anime with no season never advanced. This addresses all four. I am about 90% confident in external player auto next now.

## Summary

Hardens the external-player auto-next flow. Several related fixes:

**1. Stop a routine screen settle from suppressing auto-next.** The Stream-screen settle path (`isLoading` going false) routed into `dismissAutoNextOverlay`, which set the durable `autoNextChainAborted` flag meant only for a user Back press. `startTracking` cleared that flag only when the next launch was >12s after the last emit, so a fast source resolve left it set and the next completion silently skipped auto-next, which looked intermittent. Adds `releaseAutoNextOverlay` (hide-only, no abort) for the settle path and a separate `autoNextOverlaySuppressed` flag so a routine release still blocks `onStart` from re-raising a loader with no job behind it.

**2. Make the "Loading Next Episode" loader reliably dismissable.** Compose back-dispatch ordering kept putting the freshly-navigated Stream screen's `BackHandler` above the loader's, so Back never reached it. Back is now intercepted in `MainActivity.dispatchKeyEvent` (Activity level, before any Compose handler), so it always wins while the loader is up. The loader's safety net now clears unconditionally and is shortened to 10s (the identity-guarded clear could leave a stale overlay stuck).

**3. Fail fast to the source picker.** When the direct-auto-play flow gives up with no instant stream (all addons finished, or the 60s hard timeout), it now clears `isLoading` and releases the auto-next loader, so a non-cached next episode drops the user onto the source list immediately instead of holding the loader.

**4. Absolute-numbered (season-less) anime.** `resolveNextEpisode` now accepts a null season and orders by episode number; the advance and loader guards no longer require a season. `ExternalAutoNextEpisode.nextSeason` is nullable.

**5. Manual relaunch after an abort.** A manual launch now always clears the chain abort (it previously kept a stale abort if it landed inside the 12s continuation window).

Flag/timing logic is extracted into a pure, `Log`-free `ExternalAutoNextPolicy`, covered by unit tests.

## PR type

- [x] Critical bug fix

## Why

External-player auto-next was intermittently skipping episodes, the "Loading Next Episode" loader could get stuck with no way to back out, and non-cached next episodes left the user waiting on the loader. Anime with absolute episode numbering never auto-advanced.

## Issue or approval

Hardens the external-player auto-next introduced in #2338 (intermittent skipped episodes, an un-dismissable stuck loader, and no fallback for non-cached next episodes). No separate issue was filed.

## Reproduction steps

With an external player + debrid: auto-next through several episodes. Intermittently the next episode would not auto-advance; on a non-cached episode the "Loading Next Episode" loader could not be dismissed with Back.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

External auto-next now advances reliably, the loader is always dismissable, and a non-cached next episode falls back to the source list. No layout changes.

## Policy check

## Scope boundaries

Limited to the external-player auto-next path (`ExternalPlaybackTracker`, `ExternalAutoNextPolicy`, `PlayerNextEpisodeRules`, the auto-next loader in `MainActivity`, and the direct-auto-play give-up points in `StreamScreenViewModel`). The 60s direct-auto-play hard timeout and the configured auto-play timeout are unchanged. Internal-player playback is untouched.

## Testing

- `./gradlew :app:assembleFullDebug` builds; `:app:compileFullDebugKotlin` clean.
- New unit tests pass: `ExternalAutoNextPolicyTest` (13) and `PlayerNextEpisodeRulesTest` (7).
- Installed on an NVIDIA Shield and manually verified: auto-next chains across episodes; episode-card watched-marking; loader dismiss. On-device testing is ongoing (this is a draft).

## Screenshots / Video

Not a UI change. (Behavior fix.)

## Breaking changes

None.

## Linked issues

No linked issue.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/MainActivity.kt
+++ b/app/src/main/java/com/nuvio/tv/MainActivity.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.content.res.Configuration
 import android.os.Bundle
 import android.util.Log
+import android.view.KeyEvent
 import android.widget.Toast
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.BackHandler
@@ -805,9 +806,8 @@ class MainActivity : ComponentActivity() {
                     // of the NavHost) to hide the app cold-starting while the next source resolves.
                     val autoNextOverlay by externalPlaybackTracker.autoNextOverlay.collectAsState()
                     autoNextOverlay?.let { ov ->
-                        BackHandler(enabled = true) {
-                            externalPlaybackTracker.dismissAutoNextOverlay()
-                        }
+                        // Back is intercepted at the Activity level (dispatchKeyEvent) so it reliably
+                        // beats the destination screen's BackHandler.
                         com.nuvio.tv.ui.screens.player.LoadingOverlay(
                             visible = true,
                             backdropUrl = ov.backdrop,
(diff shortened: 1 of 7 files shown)


Laskco:
@skoruppa for some reason my previous fixes for this ended up breaking on the latest release, but everything should be working again now.

PR #2457, opened 2026-06-26
fix(player): merge skip-intro providers instead of first-match-wins

### LINKED ISSUE #2456: [Bug]: Skip Intro/Outro drops valid segments when the first provider only has a partial result
### Pre-flight checks
- [x] I can reproduce this on the latest main build.
- [x] This issue is limited to a single bug.
- [x] This is not a source/stream-specific error (it depends only on the skip-data providers, not the stream).

### App version / commit
dev, built from source

### Install method
Built from source

### Platform
Android TV

### Device model
NVIDIA Shield TV

### Android version
Android 11

### Player mode
Internal player (also affects the external-player skip-segment forwarding; the logic lives in the shared SkipIntroRepository)

### Area (tag)

### Bug description
`SkipIntroRepository` resolves intro/outro skip times from several providers (IntroDB, Anime-Skip, AniSkip), but it returns the result of the **first provider that has any data** and stops. If that first provider only has a partial result (e.g. just the intro), the remaining segments are silently dropped even when a lower-priority provider has them.

Concrete case: **Angel Beats (`kitsu:4604`) episode 1**. AniSkip lists only the intro; IntroDB has **both** the intro and the outro. Because the mal/kitsu path queries AniSkip first and returns immediately on a non-empty result, the user gets only the intro and the outro is lost, even though the app could have supplied it from IntroDB. The same shadowing can occur in the other direction depending on which provider is queried first for a given ID type.

### Steps to reproduce
1. Enable Skip Intro.
2. Play an episode where one provider has only a partial result and another has the complete set (e.g. Angel Beats S1E1: AniSkip has only the intro, IntroDB has intro + outro).
3. Observe that only the intro is skipped; the outro is never offered/skipped.

### Expected behavior
The most complete set of skip segments available across all providers should be used. If IntroDB has the outro, it should be skipped even though AniSkip (queried first) only had the intro.

### Actual behavior
Only the first non-empty provider's (partial) result is used; segments other providers have are dropped. For Angel Beats S1E1 only the intro is skipped and the outro is missed.

### Frequency
Always (for any episode where the first-matched provider has a partial result).

### Did this work before?
No, it never worked - the cascade has always returned the first non-empty provider.

Laskco:
## Summary

Fixes #2456. `SkipIntroRepository` returned the result of the **first provider that had any data** and stopped, so a provider with only a partial result shadowed a more complete one (e.g. AniSkip's intro-only result hid IntroDB's intro + outro, dropping the outro entirely). This changes the resolver to query the applicable providers and **merge their results by segment category**, so the most complete set is always used.

## How it works

For a title/episode the resolver now:
1. Resolves the cross-IDs it needs (IMDb for IntroDB, MAL for AniSkip, AniList for Anime-Skip) via ARM - exactly as before.
2. Queries each applicable provider and **collects** all their segments, instead of returning at the first non-empty one.
3. Merges by category - **opening** (`intro`/`op`/`mixed-op`), **ending** (`outro`/`ed`/`mixed-ed`/`credits`/`ending`) and **recap** - filling each category from the highest-priority provider that actually has it: **IntroDB, then Anime-Skip, then AniSkip**.

Because the merge is per-category, a provider that only has an intro can no longer hide another provider's outro (or vice versa). Each segment type comes from whoever actually has it, in priority order.

## Why this is much better than the previous method

- **Before - first-non-empty-wins:** whichever provider was queried first and returned anything "won", and everything else was discarded. A provider that listed only the intro silently dropped the outro the app could have gotten elsewhere. The outcome depended on provider ordering rather than data completeness.
- **After - best-of-all-providers:** no provider can shadow another in either direction. The user gets the complete intro + outro (+ recap) whenever *any* provider has each piece.
- Verified on-device (NVIDIA Shield):
  - **Angel Beats S1E1:** was intro-only (AniSkip shadowing IntroDB); now **intro + outro**, both from IntroDB.
  - **Absolute Duo S1E1:** IntroDB has nothing; now **intro + ending**, both from Anime-Skip.
  - **Re:Zero S1E1:** IntroDB has the outro, Anime-Skip has the intro; the merge **combines** them.
- Same caching as before - the extra provider lookups only run once per episode on a cache miss.

## PR type

- [x] Critical bug fix

## Why

The skip feature silently drops intros/outros the app actually has access to, purely because of provider query ordering. Users miss outros (or intros) that a lower-priority provider has, with no way to tell why. This makes intro/outro skipping behave correctly and completely.

## Issue or approval

Fixes #2456.

## Reproduction steps

1. Enable Skip Intro.
2. Play an episode where the first-matched provider has only a partial result (e.g. **Angel Beats S1E1**: AniSkip has only the intro, IntroDB has intro + outro).
3. Before this change only the intro is skipped and the outro is dropped; after, both are skipped.

## UI / behavior impact

- [x] No UI change

## Policy check

## Scope boundaries

Only the three resolve entry points (`getSkipIntervals`, `getSkipIntervalsForMal`, `getSkipIntervalsForKitsu`) change from return-first to collect-and-merge, plus two small private helpers (`mergeByPriority`, `segmentCategory`) and one extraction (`fetchAnimeSkipForEntries`). The per-provider fetch logic, ID resolution, provider gating, and caching are unchanged. No UI, settings, or unrelated code touched. One file: `SkipIntroRepository.kt`.

## Testing

- `:app:compileFullDebugKotlin` clean.
- On an NVIDIA Shield (Android 11), internal + external player:
  - Angel Beats S1E1: intro + outro (both IntroDB) - was intro-only before.
  - Absolute Duo S1E1: intro + ending (both Anime-Skip) - IntroDB has nothing.
  - Re:Zero S1E1: mixed (Anime-Skip intro + IntroDB outro).

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #2456.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/repository/SkipIntroRepository.kt
+++ b/app/src/main/java/com/nuvio/tv/data/repository/SkipIntroRepository.kt
@@ -40,72 +40,41 @@ class SkipIntroRepository @Inject constructor(
         val cacheKey = "$imdbId:$season:$episode"
         cache[cacheKey]?.let { return it }
 
-        if (introDbConfigured) {
-            val result = fetchFromIntroDb(imdbId, season, episode)
-            if (result.isNotEmpty()) return result.also { cache[cacheKey] = it }
-        }
-
+        val introDb = if (introDbConfigured) fetchFromIntroDb(imdbId, season, episode) else emptyList()
         val entries = resolveImdbEntries(imdbId)
+        val animeSkip = fetchAnimeSkipForEntries(entries, season, episode)
         val malId = entries.getOrNull(season - 1)?.myanimelist?.toString()
             ?: entries.firstOrNull()?.myanimelist?.toString()
-        if (malId != null) {
-            val result = fetchFromAniSkip(malId, episode)
-            if (result.isNotEmpty()) return result.also { cache[cacheKey] = it }
-        }
+        val aniSkip = if (malId != null) fetchFromAniSkip(malId, episode) else emptyList()
 
-        // AnimeSkip: try season-specific AniList ID first, then season-1 as fallback with season filter
-        val seasonAnilistId = entries.getOrNull(season - 1)?.anilist?.toString()
-        val fallbackAnilistId = entries.firstOrNull()?.anilist?.toString()
-        for ((anilistId, seasonFilter) in listOfNotNull(
-            seasonAnilistId?.let { it to null },
-            if (fallbackAnilistId != null && fallbackAnilistId != seasonAnilistId) fallbackAnilistId to season else null
-        )) {
-            val result = fetchFromAnimeSkip(anilistId, episode, season = seasonFilter)
-            if (result.isNotEmpty()) return result.also { cache[cacheKey] = it }
-        }
-
-        return emptyList<SkipInterval>().also { cache[cacheKey] = it }
+        return mergeByPriority(introDb, animeSkip, aniSkip).also { cache[cacheKey] = it }
     }
 
     suspend fun getSkipIntervalsForMal(malId: String, episode: Int): List<SkipInterval> {
         val cacheKey = "mal:$malId:$episode"
         cache[cacheKey]?.let { return it }
 
-        val aniSkipResult = fetchFromAniSkip(malId, episode)
-        if (aniSkipResult.isNotEmpty()) return aniSkipResult.also { cache[cacheKey] = it }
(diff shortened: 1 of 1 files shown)


skoruppa:
I think the same change is needed for mobile ;) The intro logic might be different there so we need to match them between versions

Laskco:
> I think the same change is needed for mobile ;) The intro logic might be different there so we need to match them between versions

I knew you were going to say that so I already made a PR for mobile as well when I made this PR haha

PR #2460, opened 2026-06-26
Hebrew translation update

haveAnIssue:
Translate the interface into Hebrew for version 0.7.11. Also, you need to change the folder name to "values-he".


haveAnIssue:
@skoruppa I just translated the texts. Do I need to submit using the PR Template?

PR #2463, opened 2026-06-26
fix(player): prevent external-player launch stalls after skip merge

### LINKED ISSUE #2462: External player launch can stall after skip segment provider merge
## Pre-flight checks

## Version & device
- App version / commit: dev after #2457 / local 0.7.11-beta test build
- Install method: Built from source
- Platform: Android TV
- Device model: SHIELD Android TV
- Android version: Android 11
- Player mode: External player
- Area: Playback (start/stop/buffering)

## Bug description
After the skip intro/outro provider merge work, launching an external player with "send intro/outro timestamps" enabled can pause on the NuvioTV loading screen before the external player appears. The app is resolving resume position and skip segments before the external player intent is sent, so the loading UI can disappear or sit silently while network provider lookups are still happening.

There is also a rare startup race where the internal player can open once even though external player is selected. The stream screen initially treats the player preference as internal while settings are still loading.

## Steps to reproduce
1. Enable external player playback.
2. Enable sending intro/outro timestamps to the external player.
3. Pick a stream for content that resolves skip segments through the merged providers.
4. Observe the launch handoff to the external player.
5. In rare cases during fast/direct autoplay, observe the player choice before settings finish loading.

## Expected behavior
The external player should open as soon as the launch intent is ready, with any resume lookup and skip-segment resolution work performed concurrently. NuvioTV should not route to the internal player while the real external-player preference is still loading.

## Actual behavior
External-player launch can be delayed by sequential skip-provider/resume lookups, and the loading state does not clearly report that intro/outro segments are being loaded. Rarely, direct/autoplay routing can see the temporary internal default and open the internal player despite external being selected.

## Frequency
Often for the launch delay when skip timestamp sending is enabled; rarely for the internal-player race.

## Did this work before?
Yes, it used to work before the skip intro/outro provider merge.

## Media details
- Content type: series/anime affected by skip-segment provider lookups
- Source: not source-specific

## Logs
Not a crash. Runtime timestamps showed the external launch path beginning, then the external activity starting noticeably later while skip segment resolution was pending.

## Anything else?
Regression follow-up to #2457.

Laskco:
## Summary

Fixes #2462. Follow-up to #2457.

This fixes three regressions around the external-player + skip-segment merge path:

1. **External player launch could stall for 1-2s after picking a stream.** When **External Player -> Send intro and outro timestamps** was enabled, Nuvio waited for resume-position lookup and skip-segment provider resolution before actually sending the external-player intent. With the merged skip resolver, that could mean multiple IntroDB / ARM / Anime-Skip / AniSkip round trips before the player opened.
2. **The loading screen could look like it was dropped right before the external player opened.** `StreamScreen` treated `launchPlayer()` like fire-and-forget, so it could begin popping the stream/loading UI before the external-player intent was actually sent. That left a silent gap where Nuvio looked finished but the external player had not opened yet.
3. **The internal player could rarely open even though External was selected.** Direct/autoplay initially collected a placeholder `PlayerPreference.INTERNAL` while DataStore was still loading the saved external-player preference, so a fast route could use the wrong player once.

This keeps the same skip data and the same merge priority from #2457. It only changes when independent work is started and when the stream screen considers the external launch complete.

## How it works

For an external-player launch the app now:

1. Starts resume-position lookup and skip-segment resolution at the same time, because those two pieces of work are independent.
2. Inside `SkipIntroRepository`, starts independent provider/ARM calls concurrently where possible instead of making every network round trip wait for the previous one.
3. Still merges by category in the same priority order as #2457: **IntroDB, then Anime-Skip, then AniSkip**. A lower-priority provider only fills a missing opening/ending/recap category.
4. Sends the external-player intent on the main thread, and only then lets `StreamScreen` run its post-launch pop/delay path.
5. Shows the existing launch/loading overlay message as **Loading intro/outro segments...** while skip timestamps are being prepared for the external player, so the wait is no longer silent.
6. Treats the player preference as unavailable until DataStore emits the real saved value, so autoplay cannot fall through to the internal player during startup.

## Why this fixes the regression

- **Before - serial pre-launch work:** resume lookup finished first, then skip resolution ran, and skip resolution could call IntroDB + ARM + Anime-Skip + AniSkip mostly back-to-back. The external-player intent was gated behind the sum of those waits.
- **After - concurrent pre-launch work:** independent work overlaps, so the external-player handoff waits on roughly the slowest required path instead of the sum of every path.
- **Before - wrong launch completion point:** the stream screen could dismiss/popup-transition before the external-player intent had actually been sent.
- **After - launch completion is real:** `launchPlayer()` suspends until the intent is sent, so the existing post-launch delay starts at the actual handoff point.
- **Before - placeholder player preference:** direct/autoplay could see `INTERNAL` before the saved setting loaded.
- **After - saved preference only:** routing waits for the real preference and respects External.

## What did not change

The #2457 merge behavior is still intact. It does **not** go back to first-provider-wins, and it does **not** pick the largest provider blob wholesale. It still fills each segment category from the highest-priority provider that has that category:

1. IntroDB
2. Anime-Skip
3. AniSkip

So a provider with only an intro cannot hide another provider's outro, and a lower-priority provider still only fills categories that higher-priority providers did not provide.

## PR type

- [x] Critical bug fix

## Why

This is a regression from the merged skip intro/outro provider work. With skip timestamps enabled for external players, launching a stream could feel delayed or silently stuck because the app resolved multiple network-backed skip sources before opening the external player. It also made the launch screen message incomplete: users could see preparation happening without any indication that intro/outro timestamps were being loaded.

The rare internal-player launch is the same external-player path seeing a temporary default before the saved setting arrives. External-player users should not randomly land in the internal player when their saved preference is external.

## Issue or approval

Fixes #2462. Follow-up to #2457.

## Reproduction steps

1. Set player mode to External.
2. Enable **Send intro and outro timestamps**.
3. Pick a stream for an episode that resolves skip data through the merged providers.
4. Before this change, the app can pause on/around the loading screen before the external player appears because resume lookup and skip resolution are serialized.
5. In rare direct/autoplay timing, the internal player can open once even though the saved player preference is external.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

The only UI-visible change is the existing loading status text now says when intro/outro segments are being loaded for the external-player launch. No layout or visual styling changes are included.

## Policy check

## Scope boundaries

Limited to the external-player skip timestamp launch path and the direct/autoplay player-preference race:

- `ExternalPlaybackTracker` now suspends until the external-player intent is sent, while resolving resume position and skip segments concurrently.
- `SkipIntroRepository` fans out independent provider/ARM calls, but keeps the #2457 merge behavior unchanged: per-category priority remains IntroDB -> Anime-Skip -> AniSkip.
- `StreamScreen` waits for the real player preference instead of using an internal-player placeholder while settings load.
- `StreamScreenViewModel` updates the existing launch/loading status text while external skip segments are being prepared.

No changes to provider priority, skip segment schema, settings, internal-player skip behavior, external result tracking, watched marking, auto-next policy, or loading-overlay layout. The loading-logo/text spacing experiment was reverted and is not part of this PR.

## Testing

- `git diff --cached --check` clean.
- `:app:compileFullDebugKotlin` clean.
- `:app:assembleFullDebug` builds.
- Installed on an NVIDIA Shield / SHIELD Android TV (Android 11): `com.nuvio.tv.test`, versionCode 1028, versionName 0.7.11-beta.
- Manually exercised the external-player launch path while investigating the regression: the delay was traced to pre-launch resume + skip resolution work, and the fix moves independent work into concurrent paths while preserving merge priority.
- Verified the PR Template Check workflow passes on #2463.

## Screenshots / Video

Not a layout change. The only UI-visible difference is transient status text on the existing loading overlay: **Loading intro/outro segments...**

## Breaking changes

None.

## Linked issues

Fixes #2462.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
@@ -21,6 +21,8 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -31,6 +33,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import javax.inject.Inject
 import javax.inject.Singleton
@@ -244,7 +247,7 @@ class ExternalPlaybackTracker @Inject constructor(
      * @param resumePositionMs Position to resume from (ms), 0 to auto-fetch
      * @param context Fallback context for fire-and-forget launch
      */
-    fun launchPlayer(
+    suspend fun launchPlayer(
         metadata: ExternalPlaybackMetadata,
         url: String,
         title: String?,
@@ -259,10 +262,18 @@ class ExternalPlaybackTracker @Inject constructor(
         // Resolve resume position (if not given) and intro/outro skip segments off the main
         // thread, then launch. Skip resolution is bounded so it never stalls the launch for long
         // and is cached, so an auto-next chain pays it only once.
-        scope.launch {
-            val position = if (resumePositionMs > 0L) resumePositionMs else getResumePosition(metadata)
-            val skipSegmentsJson = resolveSkipSegmentsJson(metadata)
-            doLaunch(url, title, headers, position, subtitles, skipSegmentsJson, context)
+        coroutineScope {
+            val positionDeferred = async {
+                if (resumePositionMs > 0L) resumePositionMs else getResumePosition(metadata)
+            }
+            val skipSegmentsDeferred = async {
+                resolveSkipSegmentsJson(metadata)
+            }
+            val position = positionDeferred.await()
+            val skipSegmentsJson = skipSegmentsDeferred.await()
+            withContext(Dispatchers.Main.immediate) {
+                doLaunch(url, title, headers, position, subtitles, skipSegmentsJson, context)
+            }
         }
     }
 
--- a/app/src/main/java/com/nuvio/tv/data/repository/SkipIntroRepository.kt
+++ b/app/src/main/java/com/nuvio/tv/data/repository/SkipIntroRepository.kt
@@ -13,6 +13,8 @@ import com.nuvio.tv.data.remote.api.IntroDbSegment
 import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.firstOrNull
 
 data class SkipInterval(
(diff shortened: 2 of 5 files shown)


Laskco:
@skoruppa This will need to be merged, after some more testing I found these few issues after the merge intro/outro providers PR. Sorry I forgot to make the previous PR as a draft!

PR #2492, opened 2026-06-30
Fix ExoPlayer playback startup, seek, and buffering defaults

tapframe:
## Summary

Fixes several playback reliability regressions around startup, seeking, MIME detection, and buffering recovery.

Changes included:
- Add a first-frame startup fallback so non-tunneled Exo playback does not remain paused forever on devices that never render the first frame while paused.
- Use sync-frame seek parameters for ExoPlayer seeks to reduce seek stalls and fragile exact-seek behavior.
- Remove active MIME preflight probing before playback; MIME is now inferred passively from filename, URL, and known response headers.
- Harden the buffering stall watchdog so it only self-seeks when duration, buffer position, and target position are valid.

## PR type

- [ ] Translation/localization only
- [x] Critical bug fix

## Why

Recent playback reports show stalls, freezing after seek, startup playback hangs, and streams that work in other Exo-based players but fail or buffer in Nuvio.

These fixes reduce risky default behavior in the playback path:
- Exact seeks can stall on some formats/devices, especially MKV/progressive/HEVC/high-bitrate streams.
- Active MIME probing can cause extra network requests before playback and can break fragile tokenized/redirected streams.
- First-frame-gated startup can remain stuck if a device does not render a frame while paused.
- Stall watchdog self-seek needed stronger guards to avoid invalid or non-forward recovery seeks.

## Issue or approval

No linked issue: critical playback stability fixes based on recent user playback reports and internal investigation.

## Reproduction steps

Reported/relevant flows:
1. Start playback on a non-tunneled Exo stream where the player reaches `STATE_READY` but video does not render while paused.
2. Seek forward/backward on progressive, MKV, HEVC, or high-bitrate streams.
3. Open streams with tokenized or redirected URLs where MIME type is unclear before playback.
4. Enter `STATE_BUFFERING` where `bufferedPosition` stops advancing and the stall watchdog attempts recovery.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change
- [ ] UI changed only to fix a documented glitch/bug
- [x] Behavior changed only to fix a documented bug/regression
- [ ] UI change has explicit maintainer approval
- [ ] Behavior change has explicit maintainer approval

## Policy check

- [x] I have read and understood `CONTRIBUTING.md`.
- [x] This PR fits the current PR policy: localization/translation only or a critical bug fix.
- [x] This PR does not add features, UI changes, refactors, or other non-critical changes.
- [x] This PR is small, focused, and limited to one issue.
[...]

PR #2493, opened 2026-06-30
perf(player): optimize playback performance, native RPU parsing and stream speed test

halibiram:
## Summary

This PR introduces a comprehensive performance optimization, diagnostic tooling, and critical bug-fix suite for video playback. It resolves severe CPU/memory bottlenecks, playback freezes on seeking, startup deadlocks, and compatibility issues with mixed-case HTTP headers.

By offloading heavy frame parsing to a native C++ layer, rewriting the parallel buffering system to support zero-copy off-heap memory, and adding host throughput diagnostic tools in Settings, this change eliminates garbage collection (GC) halts, prevents player freezes, and enables precise network tuning for end users.

---

### 1. Dolby Vision & HDR10+ Sample Processing (CPU & Memory Starvation)
*   **The Problem:** During playback of 4K Blu-ray REMUX streams containing Dolby Vision (Profile 7/8.1) or HDR10+ metadata, the player intercepted every video frame on the JVM (Kotlin). Scanning raw bytes for start codes, un-escaping emulation prevention bytes, and stripping or converting Dolby Vision RPUs and Samsung HDR10+ SEI messages frame-by-frame created extreme memory churn. This led to continuous Garbage Collection pauses (500ms - 900ms) and high CPU overhead, which starved ExoPlayer's audio renderer (causing `DefaultAudioSink: Spurious audio timestamp` errors and loop rebuffering).
*   **The Solution:** Moved all byte-level searches, HEVC start code scans, RBSP un-escaping/escaping, and SEI message filtering to the native C++ layer (`dovi_bridge.cpp`). Kotlin's track output now uses a unified `NativeOptimizedVideoTrackOutput` that delegates processing to the JNI. In addition:
    *   **Zero-Allocation:** C++ uses reusable `thread_local` buffers, preventing native and JNI heap allocation overhead.
    *   **Zero-Copy Bypass:** If a video sample is unmodified (i.e. no Dolby Vision conversion or HDR10+ stripping is required for that frame), the native layer returns `0` and the Kotlin wrapper forwards the original sample to the decoder without copying it.

### 2. Matroska (MKV) Extraction & Profile Prioritization
*   **The Problem:** 
    *   Previously, the player intercepted and buffered all HEVC video frames in Matroska containers, executing a length-delimited parser even if the stream was NOT Dolby Vision or didn't need conversion. This caused high memory allocations and playback freezes on normal HEVC tracks.
    *   The Dolby Vision profile was resolved from the Matroska codec string first. However, container codec strings are often generic, leading to incorrect profile matching (e.g. failing to distinguish Profile 7 vs 8).
    *   In AUTO mode, Dolby Vision base layer mapping did not respect display specifications, resulting in Dolby Vision data being sent to non-DV TVs (causing a purple/green tint).
    *   Matroska track format was not initialized early enough, causing a `NullPointerException` in `SampleQueue` when the extractor tried to commit samples before the output format was set.
*   **The Solution:**
    *   **Transformer Gating (`MatroskaExtractor.java`):** Introduced a `shouldTransform` check so only Dolby Vision tracks needing conversion are intercepted. Normal HEVC tracks bypass the parser entirely, preventing overhead.
    *   **Hot Path String Optimization:** Cached track codec flags (`track.isHevc`) at track initialization, avoiding expensive string comparisons on every frame.
    *   **Aligned Buffer Doubling:** Replaced immediate array allocations with aligned buffer doubling (using a 256KB boundary) in hot loops, eliminating garbage collection pressure.
    *   **Profile Prioritization:** Modified `DolbyVisionMatroskaTransformer.kt` to resolve the Dolby Vision profile from binary configuration bytes (`configBytes` containing the `dvcC` configuration block) first, falling back to the codec string only when private headers are absent.
    *   **Display-Driven Base Layer Mapping:** Gated MKV Dolby Vision base layer mode based on screen specs in AUTO mode, activating HDR10 base layer mapping if the screen does not support Dolby Vision.
    *   **Early Format Initialization:** Ensured Matroska track format is committed to the `SampleQueue` before any sample data is written, preventing `NullPointerException` crashes.

### 3. Parallel Network Buffering (`ParallelRangeDataSource` Freezes & Leaks)
*   **The Problem:**
    *   When a user performed rapid seeking, the parallel data source tried to reuse connection streams from closed or out-of-date scopes. This corrupted byte offsets, leading to indefinite playback freezes or duration metadata corruption.
    *   Large JVM `ByteArray` allocations for chunks caused heap fragmentation and Out-Of-Memory (OOM) crashes on low-end devices.
    *   CDNs/Debrid providers returning mixed-case HTTP headers (like `content-length` instead of `Content-Length`) caused chunk size detection to fail, falling back to slow sequential single-connection streaming.
    *   For non-faststart MP4 streams where the header is at the end, the bootstrap range probe connection remained open in sequential mode, blocking the player from instantly seeking back and parsing the `moov` atom.
*   **The Solution:**
    *   **Direct ByteBuffer Support:** Implemented `ByteBufferDataReader` in `ParallelRangeDataSource`. ExoPlayer can now read chunks directly from the data source into native decoder buffers, bypassing JVM-to-native array copying.
    *   **Off-Heap Allocations:** Migrated the buffer pool from Java ByteArrays to `PooledBuffer` (which wraps direct native memory allocations). Built a clean reflection-based memory collector (`freeDirectBuffer`) to explicitly release direct buffers at the OS level upon closing.
    *   **State Reset & Reuse Gate:** Enforced strict position matching for connection reuse. Closed data sources are explicitly banned from reuse.
    *   **Case-Insensitive Headers:** Header key lookups are now case-insensitive, ensuring reliable multi-connection split downloads regardless of server configurations.
    *   **Non-Faststart MP4 Seek Optimization:** The bootstrap range probe connection is closed immediately after initial analysis, allowing the player to instantly seek and parse the `moov` atom without being blocked by sequential background reads.

### 4. ExoPlayer Performance, Connection Dispatcher & Buffer Limit Tuning
*   **The Problem:**
    *   The parallel network pipeline was limited by OkHttp's default dispatcher limit (max 64 concurrent requests and max 5 requests per host), causing connection queue starvation during multi-connection chunk fetching.
    *   Excessive buffer targets (400 MB) and long buffer durations (280s) caused memory bloat, leading to crashes on low-RAM streaming devices.
*   **The Solution:**
    *   **Dispatcher limits lifted:** Configured OkHttpClient's connection pool limits to `128` concurrent requests and `128` requests per host, unlocking the full throughput of parallel chunk loading.
    *   **Settings Discrete Slider:** Replaced the continuous MB slider with an index-based chunk size selection slider (in KB) in the Settings UI for micro-tuning download sizes.
    *   **Buffer Limit Adjustments:** Updated `NuvioExoPlayerPerformanceHelper.kt` default targets from 400 MB to a safer 250 MB limit, and max buffer duration from 280s to a more stable 120s. Increased startup pre-playback buffer from 1.5s to 3s to prevent immediate rebuffering loops.

### 5. Stream Throughput Speed & Chunk Diagnostic Test
*   **The Problem:** Users had no way of knowing whether their ISP / CDN connection to a specific Debrid host was fast enough to stream high-bitrate REMUX files (e.g. 80-120 Mbps), or which chunk size configuration yielded optimal parallel download speeds.
*   **The Solution:** Added a dedicated **Stream Throughput Diagnostics** tool under Network Settings (`StreamSpeedTester.kt` & `LastPlaybackDiagnostics.kt`):
    *   **Host Diagnostics:** Automatically records the host, authentication headers, and average video bitrate of the last played stream.
    *   **Multi-Chunk Benchmarking:** Runs sequential tests benchmarking baseline single-connection speed against parallel connections with 1 MB, 4 MB, 8 MB, and 16 MB chunk sizes directly against the active Debrid/CDN server.
    *   **Real-Time Result Display:** Shows average video bitrate alongside test results so users can immediately verify if their connection meets the stream's bitrate demands.

### 6. AFR (Auto Frame Rate) Pre-flight Latency & NextLib Bypass Logic
*   **The Problem:**
    *   AFR speed probes running on headers-only streams (which are metadata calls, not actual media) introduced unnecessary network requests, delaying stream startup by 1-2 seconds.
    *   The NextLib bypass logic incorrectly treated every non-Range header as a custom header. Since `runAfrPreflightIfEnabled()` always injected `Connection: close` into probe headers, NextLib was unintentionally disabled for all non-MKV streams. Conversely, the MKV early-return (`isMkvSource`) was checked before the custom-header check, so authenticated MKV streams still used NextLib even though NextLib cannot forward those headers.
*   **The Solution:**
    *   Bypassed NextLib AFR probes on metadata/headers streams.
    *   Introduced a global frame rate cache to reuse resolved stream frame rates and prevent redundant network round-trips.
    *   Separated original stream headers from synthetic probe headers: `shouldUseNextLibProbe` now receives the original stream headers (before `Connection: close` is injected), so it can accurately determine whether the stream requires authentication headers that NextLib cannot forward. `Connection: close` is only added for the extractor fallback path.
    *   Moved the custom-header check before the MKV early-return, ensuring authenticated MKV streams also bypass NextLib correctly.

### 7. TV Recommendations IPC Optimization
*   **The Problem:** The TV recommendation manager was updating the home screen recommendation row during active playback, triggering unnecessary IPC calls to the system's content provider on every playback state change. This caused micro-stutters on low-end devices.
*   **The Solution:** Gated TV recommendation updates to only fire when playback is paused or stopped, preventing IPC thrashing during active video rendering.

### 8. Subtitle Loading & Stream Screen Optimization
*   **The Problem:**
    *   Subtitle download calls created fresh HTTP clients per request, incurring redundant SSL handshake overhead.
    *   The stream/debrid resolution screen showed a loading overlay flicker when transitioning to the player.
*   **The Solution:**
    *   Rewrote subtitle downloading and mime probes to share the global playback `OkHttpClient` instance, avoiding redundant connection overhead.
    *   Prevented loading overlay flicker on the stream screen during debrid resolution and player launch transitions.

---

## PR type

- [x] Critical bug fix

## Issue or approval

No linked issue - fixes critical playback freezes, CPU-induced audio desync loops, NextLib AFR bypass regression, and Matroska NullPointerException crashes. Approved for local optimization.

## Reproduction steps

1. Play a 4K DV7 REMUX MKV stream and observe GC pauses and audio desync in previous builds.
2. Seek rapidly on a parallel-buffered stream and observe freeze/duration corruption.
3. Play a stream from a CDN returning lowercase `content-length` headers and observe single-connection fallback.
4. Play any non-MKV HTTP stream and observe NextLib AFR probe being incorrectly bypassed (no frame rate detection).
5. Open Network Settings after playback and run the Stream Speed Test to verify host throughput across baseline and parallel chunk sizes.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

This PR does not alter ExoPlayer's core rendering pipelines, the underlying FFmpeg library binaries, or the UI layout outside of updating the buffer size settings slider representation and adding the stream throughput diagnostic card in Network Settings.

## Testing

- **Compilation:** Successfully built using `:app:compileFullDebugSources` task.
- **Dolby Vision & HDR10+:** Tested playing hybrid Profile 7 + HDR10+ BD-Remux streams. Verified that RPU conversion to 8.1 and SEI stripping compile and run natively with zero JNI heap allocations, preventing spurious audio sync messages.
- **Parallel Data Source:** Tested seek responsiveness and prefetching under high parallel connection counts.
- **Stream Throughput Test:** Tested running diagnostic speed test on host servers with baseline, 1 MB, 4 MB, 8 MB, and 16 MB chunks; verified accuracy against average video bitrate readout.
- **NextLib AFR:** Verified NextLib probe is correctly used for public streams (no custom headers) and correctly bypassed for authenticated streams (RealDebrid/Premiumize), regardless of MKV or non-MKV format.
- **Matroska Format Init:** Verified early format initialization prevents `SampleQueue NullPointerException` on track preparation.
- **TV Recommendations:** Verified recommendation updates only trigger on pause/stop, not during active playback.

## Screenshots / Video

image

## Breaking changes

None.

## Linked issues

None.


ram130:
Great updates.  Will review as well. 

Can you confirm watch dog changes are in sync with this change? https://github.com/NuvioMedia/NuvioTV/pull/2492

halibiram:
> Can you confirm watch dog changes are in sync with this change? #2492

There was a conflict, but I resolved it by keeping the changes from the dev branch

ram130:
Looks great. Ok to merge.

PR #2447, opened 2026-06-25
fix(player): backport upstream Media3 DTS-HD MA extraction logic to custom MKV parser

iMakeSoftware:
#### Summary ####
Backports the official upstream AndroidX Media3 `MatroskaExtractor` frame-parsing fix to the project's decoupled container extraction layer. This change ensures that high-bitrate multi-channel audio tracks are correctly identified and processed by the player.

#### Why ####
The built-in player uses a decoupled, custom package for parsing MKV containers. Because these files were separated from standard framework updates, they lacked the definitions required to parse DTS-HD headers. This caused the engine to drop the lossless extension frames entirely.

#### Issue ####
On capable Android 11+ hardware, DTS-HD Master Audio and DTS:X tracks consistently fail to bitstream. They drop down to lossy 5.1 DTS Core instead. 

Previously, the only workaround was forcing **Tunneled Playback**. However, tunneling bypasses the Android OS video pipeline. This automatically disables critical system features on high-end hardware.

#### Reproduction steps ####
1. Connect an Android TV device to a compatible AV receiver or soundbar over HDMI.
2. Launch a 1080p Blu-ray remux file containing a native DTS-HD MA audio track.
3. Ensure **Tunneled Playback** is disabled in the app configuration settings.
4. Observe the AV receiver display showing a fallback to legacy lossy **DTS 5.1**.
5. Check system overlays to confirm that AI Upscaling is active but audio is downgraded.

#### UI / behavior impact ####
Adds support for DTS-HD passthrough to the built-in ExoPlayer without requiring tunneled playback. This allows devices like the NVIDIA Shield TV to retain features like AI Upscaling and frame-rate matching.

Verified real-world system behavior using an **NVIDIA Shield TV**:
* **Audio:** AV receiver successfully detects and lights up the **DTS-HD MA / DTS:X** bitstream.
* **Video:** The Shield's hardware **AI Upscaling** engine remains fully **Active**.
* **Display:** OS-level **Refresh Rate Matching** successfully triggers and switches display modes.

PR #2542, opened 2026-07-04
fix(player) repair DTS-HD MA detection logic implented in PR #2447 

iMakeSoftware:
Summary
Backports the official upstream AndroidX Media3 MatroskaExtractor frame-parsing fix to the project's decoupled container extraction layer. This change ensures that high-bitrate multi-channel audio tracks are correctly identified and processed by the player.

Why
The built-in player uses a decoupled, custom package for parsing MKV containers. Because these files were separated from standard framework updates, they lacked the definitions required to parse DTS-HD headers. This caused the engine to drop the lossless extension frames entirely.

Issue
On capable Android 11+ hardware, DTS-HD Master Audio and DTS:X tracks consistently fail to bitstream. They drop down to lossy 5.1 DTS Core instead.

Previously, the only workaround was forcing Tunneled Playback. However, tunneling bypasses the Android OS video pipeline. This automatically disables critical system features on high-end hardware.

Reproduction steps
Connect an Android TV device to a compatible AV receiver or soundbar over HDMI.
Launch a 1080p Blu-ray remux file containing a native DTS-HD MA audio track.
Ensure Tunneled Playback is disabled in the app configuration settings.
Observe the AV receiver display showing a fallback to legacy lossy DTS 5.1.
Check system overlays to confirm that AI Upscaling is active but audio is downgraded.
UI / behavior impact
Adds support for DTS-HD passthrough to the built-in ExoPlayer without requiring tunneled playback. This allows devices like the NVIDIA Shield TV to retain features like AI Upscaling and frame-rate matching.

Verified real-world system behavior using an NVIDIA Shield TV:

Audio: AV receiver successfully detects and lights up the DTS-HD MA / DTS:X bitstream.
Video: The Shield's hardware AI Upscaling engine remains fully Active.
Display: OS-level Refresh Rate Matching successfully triggers and switches display modes.


harryposter:
@iMakeSoftware I just tested the latest debug version and dts hd/hd-ma is still not passing through, it's playing core standard dts 5.1 only. I double checked the same file with emby again which passes it through correctly.
(Shield pro 2019)
One thing to note is that dts hd/hd-ma passthrough was working properly in versions prior to 0.7.11, can't we reverse to that previous version of exoplayer?

iMakeSoftware:
My experience has been the inverse, dts hd/hd-ma passthrough had never worked for me despite the fact im using a direct hdmi connection to a denon avr with all codec passthrough options enabled and all other audio processing in the shield settings disabled yet it never worked pre 0.7.11 no matter what I did (it wasn't a handshake issue because I powered off all devices completely, and tried with both arc disabled and enabled as well as cec, and no luck) after the changes I've tested against multiple samples of 4k DV and 1080p remuxes with dts-hd/hd-ma (and confirmed ai upscaling and shield os frame rate matching toggle are was functional simultaneously on 1080p dts hd/hd-ma) without issue.

ill try to reintegrate the old logic pre 0.7.11 logic as a seamless fallback check for all DTS core only detected tracks but not sure when I'll be able to get it done, probably within a week or two

Note: I'm referring only to the built-in exoplayer, not other any other players

ysosrs123:
I think I have this fixed, will open a PR today/tomorrow hopefully.

ysosrs123:
PR https://github.com/NuvioMedia/NuvioTV/pull/2558

PR #2544, opened 2026-07-04
fix(player): survive seek reopens in parallel range data source

### LINKED ISSUE #2543: Parallel Connections: constant rebuffering on non-faststart MP4s with non-interleaved track layout — chunk downloads restart on every seek reopen
### Pre-flight checks

### App version / commit

0.7.14-beta (self-built from the release tag; ParallelRangeDataSource identical to current dev)

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Homatics Box R 4K Plus (Amlogic S905X4-K)

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

### Bug description

On MP4s muxed with the moov atom at the tail and track data in large non-interleaved lumps, Mp4Extractor maintains several distant read cursors and ExoPlayer's loader closes/reopens the data source every few seconds. ParallelRangeDataSource treats each reopen as a cold start: all in-flight chunk downloads are cancelled and restarted. Result: the same chunks download repeatedly, bandwidth is exhausted on discarded transfer, and playback rebuffers constantly on files whose bitrate the connection could easily sustain.

### Steps to reproduce

1. Official 0.7.14-beta with Parallel Connections ON (tested at 4 connections × 32 MB).
2. Play a large non-faststart MP4 with non-interleaved track layout over HTTP (common among web-DL releases; sample details available on request).
3. Watch for 2+ minutes.
4. Observe rebuffering beginning around the 1-minute mark, recurring thereafter.
5. Capture logcat and filter tag ParallelRangeDS to see the chunk-restart signature.

### Expected behavior

Smooth playback. Connection bandwidth comfortably exceeds the file bitrate, and well-interleaved files of similar size/bitrate from the same source play without any rebuffering on the same device and settings.

### Actual behavior

Playback starts normally, then rebuffers within ~1 minute and keeps rebuffering. No crash, freeze, or error text — the UI stays responsive. Logcat (tag ParallelRangeDS) shows chunk downloads being cancelled and restarted on every seek reopen: in an ~84 s window, 55 "Starting chunk download" lines vs 21 "Successfully downloaded chunk", with 13 "Cleared global buffer pool" teardowns and the same indices restarting repeatedly (idx 3 ×11, idx 2 ×11, idx 4 ×9, idx 1 ×9). Roughly 60% of downloaded data is discarded mid-flight, which starves playback of bandwidth.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

- Content type: movie
- Stream/source: debrid (TorBox) via Torrentio addon
- Video format: MP4 container, HEVC / Dolby Vision Profile 8.1 (dvhe.08.06), 3840×1920, 24 fps, 18.5 GB, moov atom at end of file, track data non-interleaved

```shell
07-04 11:42:09.415 D/ParallelRangeDS: Scheduling chunk 0
07-04 11:42:09.416 D/ParallelRangeDS: Scheduling chunk 1
07-04 11:42:09.417 D/ParallelRangeDS: Starting chunk download: idx=1, range=33554432-67108864
07-04 11:42:09.417 D/ParallelRangeDS: Starting chunk download: idx=0, range=0-33554432
07-04 11:42:09.418 D/ParallelRangeDS: Starting chunk download: idx=2, range=67108864-100663296
07-04 11:42:09.420 D/ParallelRangeDS: Starting chunk download: idx=3, range=100663296-134217728
07-04 11:42:09.423 D/ParallelRangeDS: Cleared global buffer pool as all ParallelRangeDataSource instances are closed
07-04 11:42:10.940 D/ParallelRangeDS: Parallel mode: 4 connections, 32MB chunks, file=17716MB
07-04 11:42:12.037 D/ParallelRangeDS: Starting chunk download: idx=552, range=18522046464-18555600896
(pattern repeats — full session shows 55 starts / 21 completions / 13 pool teardowns in ~84 s)
```

### Anything else? (optional)

Root cause identified: ExoPlayer creates a new data source instance per seek, and ParallelRangeDataSource cancels all in-flight chunk downloads on close and re-probes on open. With Parallel Connections OFF the same access pattern shows as raw per-request latency instead (49 reopens averaging ~1.2 s in one session). I have a tested fix ready to submit as a PR: chunk downloads survive reopens via a shared session, prefetch is gated to sequential readers, and eviction is LRU under the existing MemoryBudget tiers.

ysosrs123:
## Summary

Fixes constant rebuffering on scatter-read MP4s (non-faststart, non-interleaved track layout) when Parallel Connections is enabled. ExoPlayer creates a new data source instance on every seek; ParallelRangeDataSource cancelled all in-flight chunk downloads on close and re-probed on open, so the multi-cursor read pattern these files force made the same 32 MB chunks restart dozens of times — measured 55 chunk starts vs 21 completions in ~84 s, (~60% of transfer discarded). 

Chunk futures (completed and in-flight) now belong to a shared session that survives the close/open boundary; prefetch lookahead is earned (granted after 1 MB served sequentially) so side cursors stop triggering prefetch fan-outs;eviction is touch-LRU under a memory-tiered cap (connections+2 chunks on low-RAM, connections+4 on high-RAM, via the existing MemoryBudget.isLowRamTier); a 2 s touch guard protects chunks mid-read, so occupancy is bounded above by cap + 2 rather than exact; 

MemoryBudget.BUFFER_OVERHEAD corrected 1 → 2 so settings enforcement matches the data source's actual ceiling. 

Result on the pathological sample: zero rebuffering, each chunk downloaded exactly once.

Update: fix decoupled from the parallel-connections opt-in

The session-owned chunk mechanism is the actual fix for the scatter-read MP4 pathology, and the parallel prefetch fan-out is a separate throughput feature. The latest commit decouples them: progressive MP4 (resolved mime video/mp4) now engages the chunk session even when Parallel Connections is off, pinned to a single connection and a fixed 8 MB chunk.

Bounds with the setting off: earned prefetch caps lookahead at two chunks, side cursors fetch only the chunk they touch, and the session cap keeps the existing memory tiers — so request concurrency and retained memory hold at plain-path levels while the re-downloaded transfer measured above disappears. HLS/DASH, non-MP4 progressive, and parallel-on behaviour are unchanged. Engagement is logged as MP4_SESSION for easy verification.

This means users no longer need to enable a multi-connection mode (with its memory and rate-limit trade-offs on some debrid providers) just to make non-faststart MP4s watchable.

## PR type

- [x] Critical bug fix

## Why

Affected files are effectively unwatchable — rebuffering begins within a minute and worsens — despite connection bandwidth comfortably exceeding the file bitrate. The defect wastes roughly two-thirds of downloaded data and applies to a common class of web-DL MP4 releases, so users hit it through no fault of their configuration.

## Issue or approval

Fixes #2543 

## Reproduction steps

Official 0.7.14-beta, Parallel Connections ON (reproduced at 4 connections × 32 MB).
Play a large non-faststart MP4 with non-interleaved track layout over HTTP.
Playback starts, then rebuffers within ~1 minute, worsening over time.
Logcat (tag ParallelRangeDS): the same Starting chunk download: idx=N lines repeat dozens of times per index, with completions far below starts. Full counts and analysis in the linked issue.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Single-connection (Parallel OFF), HLS/DASH, and subtitle-bypass paths untouched. No new settings, UI, or user-facing options. The BUFFER_OVERHEAD correction is included because it makes the settings-screen memory enforcement consistent with the component this PR fixes; it is one constant. A related frame-rate-matching fix for the same file class was deliberately excluded and will be proposed separately if wanted.

## Testing

Device: Homatics Box R 4K Plus (Amlogic S905X4-K, Android 14, armeabi-v7a), TorBox debrid CDN. Fix developed and soak-tested on a fork based on 0.7.14-beta; this branch (upstream dev + this commit) build-verified with CI_USE_DEBUG_SIGNING=true ./gradlew assembleFullRelease (note: the debug build type currently requires the CI release keystore, hence the release-variant verification). Manual flows: pathological 18.5 GB non-faststart MP4 — before: rebuffering within a minute; after: zero rebuffering across full sessions, logcat confirms each chunk index downloads once and reopens attach to the warm session. Regression: well-interleaved MKV remuxes (debrid + direct HTTP) unchanged; aggressive seeking (±10 min jumps) recovers cleanly; sideloaded subtitles unaffected; Parallel OFF unchanged; memory stable via dumpsys meminfo across a soak session on both tiers' maths (4 GB device).

## Screenshots / Video

Not a UI change.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/ParallelRangeDataSource.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/ParallelRangeDataSource.kt
@@ -123,6 +123,7 @@ internal class ParallelRangeDataSource(
 
         private class ChunkSession(
             val requestUri: Uri,
+            val requestHeaders: Map<String, String>,
             val chunkSize: Long,
             val chunkCap: Int
         ) {
@@ -201,6 +202,7 @@ internal class ParallelRangeDataSource(
          */
         private fun obtainSession(
             requestUri: Uri,
+            requestHeaders: Map<String, String>,
             chunkSz: Long,
             chunkCap: Int,
             poolCap: Int
@@ -210,14 +212,15 @@ internal class ParallelRangeDataSource(
                 if (existing != null) {
                     val fresh = SystemClock.uptimeMillis() - existing.lastUsedAtMs <= RETAINED_SESSION_TTL_MS
                     if (fresh && !existing.abandoned.get() &&
-                        existing.requestUri == requestUri && existing.chunkSize == chunkSz
+                        existing.requestUri == requestUri && existing.chunkSize == chunkSz &&
+                        existing.requestHeaders == requestHeaders
                     ) {
                         existing.lastUsedAtMs = SystemClock.uptimeMillis()
                         return existing
                     }
                     teardownSessionLocked(existing, poolCap)
                 }
-                val created = ChunkSession(requestUri, chunkSz, chunkCap)
+                val created = ChunkSession(requestUri, requestHeaders, chunkSz, chunkCap)
                 currentChunkSession = created
                 return created
             }
@@ -417,7 +420,7 @@ internal class ParallelRangeDataSource(
         // re-opens, the failed futures are gone, and downloads retry against
         // the session's URI — with the full probe as the eventual fallback via
         // session teardown on TTL.
-        val attachedSession = obtainSession(dataSpec.uri, chunkSize, sessionChunkCap, maxPoolSize)
+        val attachedSession = obtainSession(dataSpec.uri, dataSpec.httpRequestHeaders, chunkSize, sessionChunkCap, maxPoolSize)
         session = attachedSession
         val warmLength = attachedSession.totalLength
         if (warmLength > 0L && dataSpec.position in 0 until warmLength) {
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/ParallelRangeDataSource.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/ParallelRangeDataSource.kt
@@ -6,6 +6,7 @@ import androidx.media3.common.C
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.datasource.DataSource
 import androidx.media3.datasource.DataSpec
+import androidx.media3.datasource.HttpDataSource
 import androidx.media3.datasource.TransferListener
 import androidx.media3.datasource.okhttp.OkHttpDataSource
 import java.io.InterruptedIOException
@@ -21,6 +22,7 @@ import java.util.concurrent.ThreadFactory
 import java.util.concurrent.TimeUnit
 import com.nuvio.tv.data.local.PlayerSettings
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
 import android.os.SystemClock
 
 import java.nio.ByteBuffer
@@ -121,6 +123,18 @@ internal class ParallelRangeDataSource(
         // the chunk instead of spinning forever.
         private const val MAX_CONSECUTIVE_ZERO_READS = 3
 
+        // nt-tier2: HTTP 429/503 is server-side rate-limiting, not a stalled
+        // socket. Back off before retrying (immediate retry just re-hits the
+        // limit), and after repeated hits set a one-way session latch so
+        // prefetch drops to a single connection. Only reachable when parallel
+        // connections are enabled (opt-in; off by default).
+        private const val RATE_LIMIT_MAX_BACKOFF_RETRIES = 3
+        private const val RATE_LIMIT_CLAMP_THRESHOLD = 3
+        private const val RATE_LIMIT_BACKOFF_BASE_MS = 500L
+        private const val RATE_LIMIT_BACKOFF_MAX_MS = 3_000L
+        private const val RATE_LIMIT_BACKOFF_JITTER_MS = 250L
+        private const val RATE_LIMIT_SLEEP_SLICE_MS = 100L
+
         private class ChunkSession(
             val requestUri: Uri,
             val requestHeaders: Map<String, String>,
@@ -132,6 +146,11 @@ internal class ParallelRangeDataSource(
             val futures = ConcurrentHashMap<Long, CompletableFuture<DownloadedChunk>>()
             val lastTouch = ConcurrentHashMap<Long, Long>()
             val abandoned = AtomicBoolean(false)
+            // nt-tier2: set once when the server rate-limits us repeatedly; the
+            // read path then holds prefetch to a single connection for this
+            // session. One-way (never cleared) — a new stream gets a new session.
+            val rateLimited = AtomicBoolean(false)
+            val rateLimit429s = AtomicInteger(0)
             val activeSources: MutableSet<DataSource> = java.util.concurrent.ConcurrentHashMap.newKeySet()
             @Volatile var lastUsedAtMs: Long = SystemClock.uptimeMillis()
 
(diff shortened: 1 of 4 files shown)


halibiram:
makes sense for fixing seek buffering issues, but reading the shared `ByteBuffer` without `duplicate()` will cause a concurrency bug, and incomplete chunk downloads can trigger infinite recursion

ysosrs123:
> makes sense for fixing seek buffering issues, but reading the shared `ByteBuffer` without `duplicate()` will cause a concurrency bug, and incomplete chunk downloads can trigger infinite recursion

Both confirmed and fixed, thanks. The byte-array read path was mutating the shared chunk buffer's position directly — it now reads through a duplicate(), as the ByteBuffer path already did. On incomplete downloads: a chunk that hit EOF short of its requested range was cached as complete, and a short non-final chunk then loops both read paths at the phantom chunk boundary — short chunks now fail and retry, with a zero-progress guard for a source that keeps returning 0. Pushed as two commits: the review fixes, plus a follow-up that cancels dropped chunk futures (an orphaned download otherwise leaks its pooled buffer) and makes the cap ceiling hard at cap + 2.

Still can't improve/fix the buffering issues affecting mp4's when parallel connections are off. But I believe this issue existed before parallel connections were implemented, and is simply a result of the way these mp4 files are designed. Will keep working on a solution.

halibiram:
looks good overall, the session-based chunk reuse makes sense for the seek/rebuffering issue.

 the normal fallback path doesn’t set `totalFileLength`, `bytesRemaining`, or `position`, unlike the subtitle fallback path. That can leave the datasource in a weird state on non range/unknown length responses.

session reuse only checks URI + chunk size, so headers/auth changes could be risky.

ysosrs123:
> looks good overall, the session-based chunk reuse makes sense for the seek/rebuffering issue.
> 
> the normal fallback path doesn’t set `totalFileLength`, `bytesRemaining`, or `position`, unlike the subtitle fallback path. That can leave the datasource in a weird state on non range/unknown length responses.
> 
> session reuse only checks URI + chunk size, so headers/auth changes could be risky.

Both addressed, thanks. The fallback branch now sets the length fields the same way the subtitle path does (position was already being set earlier in open(), but the two length fields weren't) — and the deeper version of the problem is fixed too: the fields were never reset between opens, so a fallback open could inherit a stale totalFileLength from an earlier parallel open on the same instance, which was enough to satisfy the reopen fast path's length check. A fresh open now resets both and closes any prior fallback source, and the reopen fast path is explicitly gated to non-fallback mode.

On session identity: the DataSpec's httpRequestHeaders are now part of the session match alongside URI and chunk size, so a reopen with changed headers/auth tears the old session down and probes fresh rather than adopting its chunks and resolved URL.

While re-running the suite I also found MemoryBudgetTest.testTotalUsageMb still asserted the pre-correction BUFFER_OVERHEAD arithmetic — updated in a third, test-only commit.

Pushed as three commits.

ysosrs123:
Also, just a reminder that this fixes the buffering issues affecting mp4 files, but only if a user has parallel connections turned on.  I haven't been able to fix it with parallel connections turned off yet.  Will leave you guys to decide whether to hold off on merging until the second part is fixed. Or do that separately later.

ysosrs123:
Have fixed so will update the PR

ysosrs123:
@halibiram I've pushed one more commit: it decouples the MP4 fix from the parallel-connections opt-in. The session-owned chunks are what actually fix the scatter-read pathology, so progressive MP4 now engages the session source even with the setting off, pinned to one connection and a fixed 8 MB chunk — request concurrency and retained memory stay at plain-path levels (earned prefetch caps lookahead at two chunks; the session cap keeps the existing memory tiers). Verified on-device: MP4_SESSION engagement logged and a previously unwatchable non-faststart MP4 plays cleanly with parallel connections off.

ysosrs123:

PR #2557, opened 2026-07-05
fix(player): correct Dolby Vision Profile 5/7 conversion output

### LINKED ISSUE #3: 4 things to work on no rush the app is amazing
1. Adding Trakt
2. The "continue watching" cannot be edited to be removed or deleted
3. For TV Series the ability to consecutively play the next proceeding episode by default
4. Time stamp being able to see the video player rewinding and fast forwarding where exactly the content is on screen progression

ysosrs123:
**Summary**

Four related correctness fixes to the libdovi-based Dolby Vision → Profile 8.1 conversion path. They share the same code paths and rationale but touch distinct cases (P5 conversion, P7 in-band enhancement layer, conversion-mode labels, and BlockAdditional conversion accounting). Grouped as a single PR; each fix is independently small and could be split if needed.

1. **P7 single-track in-band enhancement-layer strip (primary fix).** FraMeSToR-class UHD remuxes carry the DV7 enhancement layer in unspec63 (type-63) wrapper NALs at nuh_layer_id 0, in-band in the HEVC samples — not in Matroska BlockAdditional. The existing EL filter only drops NALs at layer_id > 0, so these pass through; a converted 8.1 stream still carrying an enhancement layer makes the DV decoder fall back to HDR10. Now stripped in both the MKV rewriter and the native MP4/TS loops.

2.  **P5 conversion mode.** conversionMode() sent libdovi mode 3 for Profile 5, but in the bundled C API mode 3 is "convert to static 8.4 (HLG)", not "P5 to 8.1". Corrected to mode 2, which explicitly handles source profile 5.

3. **Advanced conversion-mode labels.** The Mode 3 / Mode 4 setting strings described the old dovi_tool CLI numbering; corrected to match the C API (Mode 3 = static 8.4/HLG, Mode 4 = 8.1 preserve-mapping).

4. **Single conversion + clean drop on the MKV BlockAdditional path**. The BlockAdditional RPU was converted once in the extractor hook and again in transformHevcSample; now converted once and appended verbatim. On conversion failure the RPU is dropped (base layer continues as HDR10) instead of forwarding raw P7 bytes under 8.1 signalling, which produced a stream whose codec string and per-frame metadata disagreed.

**PR type**

[ ] Translation/localization only
[x] Critical bug fix

**Why**

DV7→8.1 conversion silently produced HDR10 instead of Dolby Vision on the most common Profile 7 content (single-track UHD remuxes), because the enhancement layer is carried in a NAL form the existing filter never removed. DV5 conversion produced the wrong transfer flavour (static 8.4/HLG instead of 8.1) due to a CLI-vs-C-API mode-number mismatch. The BlockAdditional path forwarded unconvertible RPUs under mismatched 8.1 signalling. Each degrades Dolby Vision output on affected content.

**Issue** 
Not yet filed

**Reproduction steps**

1. On a device using the DV convert path (no native P7 decoder, or manual "Convert to DV8.1" selected).
2. Play a Profile 7 single-track UHD remux (FraMeSToR-class, e.g. a P7 FEL or MEL Blu-ray remux).
3. Observe the display's HDR badge: it shows HDR10, not Dolby Vision, despite DV8.1 conversion being active.
4. (P5 case) Enable "Convert DV5 to DV8.1" and play a Profile 5 stream; the output uses a static 8.4/HLG transfer rather than 8.1.

**UI / behavior impact**

 [ ] No UI change
 [ ] No behavior change
 [x] UI changed only to fix a documented glitch/bug
 [x] Behavior changed only to fix a documented bug/regression

_(UI item refers only to fix #3: two advanced conversion-mode setting labels that described the wrong libdovi modes. No layout or new UI.)_

Policy check

 [x]  I have read and understood CONTRIBUTING.md.
  [x] This PR fits the current PR policy: localization/translation only or a critical bug fix.
 [x]  This PR does not add features, UI changes, refactors, or other non-critical changes.
_(The only UI-facing change is correcting two incorrect setting labels — a documented-bug fix, not new UI or polish.)_
  [x] This PR is small, focused, and limited to one issue.
Combined scope (four related DV-conversion correctness fixes) was approved by the maintainer for this PR; noted here for transparency.
  [x] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
No refactors, cleanups, or formatting changes. The four fixes are related DV-conversion corrections; combined scope approved by the maintainer.
 [x]  This PR includes a linked issue, reproduction steps, and testing notes if it is a critical bug fix.
 [x]  I listed the testing performed below.

**Scope boundaries**
Intentionally not included, to keep this to conversion correctness:

- FEL/MEL enhancement-layer detection/diagnostics (developed downstream, deliberately excluded as non-critical).
- Any diagnostics-panel rows or reporting.
- Unrelated audio (DTS-HD/DTS:X) extractor changes from the same downstream fork.
- No refactors, formatting, or cleanups.

**Testing**
Verified on-device: Homatics Box R 4K Plus (Amlogic S905X4, armeabi-v7a, Android 14), LG C9 OLED, manual "Convert to DV8.1".

- Profile 7 FEL remux (Braveheart 2160p) and Profile 7 MEL remux (The Last Jedi 2160p): both now engage Dolby Vision through conversion; previously both fell back to HDR10. The unspec63 NAL carriage was confirmed by probing the actual per-sample NAL composition of both files.
- No conversion-failure drops occurred on either file (anchored logcat, DV7_MKV: tag).

**Caveat — I have not built this against a clean dev checkout**. Changes were developed and tested in a downstream fork build; the native side (dovi_bridge.cpp) in particular warrants a maintainer compile-check before merge. Verified by symbol-resolution and inspection against current dev, not by a from-dev build.

**Screenshots / Video**
Fix #3 changes two advanced setting label strings (Mode 3 / Mode 4 under DV7 conversion mode). No layout or visual change beyond the corrected text; can provide a before/after screenshot of the setting rows on request.

**Breaking changes**
None.

**Linked issues**
No separate bug issue filed yet — can file one for the P7→HDR10 fallback if preferred.

### DIFF
--- a/app/src/main/cpp/dovi_bridge.cpp
+++ b/app/src/main/cpp/dovi_bridge.cpp
@@ -1,5 +1,6 @@
 #include <jni.h>
 #include <android/log.h>
+#include <atomic>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
@@ -93,6 +94,19 @@ static inline uint8_t map_conversion_mode(jint mode) {
             return 4U;
         default:
             return 2U;
+    }
+}
+
+// DV7 review F5: counts RPUs dropped because per-frame conversion failed on
+// the MP4/TS sample path, with a throttled log (first drop + every 100th).
+// Process-wide by design: it is diagnostics-only and monotonically counts;
+// worst-case cross-stream interleave slightly skews the throttle cadence.
+static std::atomic<uint64_t> g_rpuDropCount{0};
+static inline void noteRpuDropOnFailure() {
+    const uint64_t n = g_rpuDropCount.fetch_add(1, std::memory_order_relaxed) + 1;
+    if (n == 1 || n % 100 == 0) {
+        LOGW("DV7_NATIVE: RPU conversion failed; dropping RPU (dropped=%llu)",
+             static_cast<unsigned long long>(n));
     }
 }
 #endif
@@ -511,7 +525,7 @@ Java_com_nuvio_tv_core_player_DoviBridge_nativeProcessVideoSample(
 
             if (convertDovi || stripDoviRpu) {
                 bool isRpu = (nalType == 62);
-                bool isEl = (layerId > 0);
+                bool isEl = (layerId > 0) || (nalType == 63); // DV7 F3: unspec63 = single-track EL carriage at layer 0 (probe-verified)
                 if (isRpu) {
                     if (stripDoviRpu) {
                         shouldDrop = true;
@@ -545,6 +559,16 @@ Java_com_nuvio_tv_core_player_DoviBridge_nativeProcessVideoSample(
                         }
 #endif
                         if (!processed) {
+#if DOVI_REAL_LINKED
+                            // DV7 review F5: conversion failed — drop the RPU
+                            // rather than forwarding the raw P7 RPU under 8.1
+                            // signalling (mixed streams, frame after frame).
+                            // The base layer continues as HDR10.
+                            shouldDrop = true;
+                            noteRpuDropOnFailure();
+#else
+                            // Stub build (no libdovi linked): conversion is
+                            // never truly active; keep raw passthrough.
                             processed = true;
                             processedNal.resize(nalSize);
                             std::memcpy(processedNal.data(), sampleBuffer.data() + nalStart, nalSize);
@@ -552,6 +576,7 @@ Java_com_nuvio_tv_core_player_DoviBridge_nativeProcessVideoSample(
                                 processedNal[0] &= 0xFE;
                                 processedNal[1] &= 0x07;
                             }
+#endif
                         }
                     }
                 } else if (isEl) {
@@ -621,7 +646,7 @@ Java_com_nuvio_tv_core_player_DoviBridge_nativeProcessVideoSample(
 
                 if (convertDovi || stripDoviRpu) {
                     bool isRpu = (nalType == 62);
-                    bool isEl = (layerId > 0);
+                    bool isEl = (layerId > 0) || (nalType == 63); // DV7 F3: unspec63 = single-track EL carriage at layer 0 (probe-verified)
                     if (isRpu) {
                         if (stripDoviRpu) {
                             shouldDrop = true;
@@ -655,6 +680,13 @@ Java_com_nuvio_tv_core_player_DoviBridge_nativeProcessVideoSample(
                             }
 #endif
                             if (!processed) {
+#if DOVI_REAL_LINKED
+                                // DV7 review F5: drop, don't forward raw P7
+                                // under 8.1 signalling (see length-delimited
+                                // loop above).
+                                shouldDrop = true;
+                                noteRpuDropOnFailure();
+#else
                                 processed = true;
                                 processedNal.resize(nalSize);
                                 std::memcpy(processedNal.data(), sampleBuffer.data() + nalBegin, nalSize);
@@ -662,6 +694,7 @@ Java_com_nuvio_tv_core_player_DoviBridge_nativeProcessVideoSample(
                                     processedNal[0] &= 0xFE;
                                     processedNal[1] &= 0x07;
                                 }
+#endif
                             }
                         }
                     } else if (isEl) {
--- a/app/src/main/java/com/nuvio/tv/core/player/DolbyVisionExtractorsFactory.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/DolbyVisionExtractorsFactory.kt
@@ -99,13 +99,20 @@ internal enum class NalFormat { ANNEX_B, LENGTH_DELIMITED }
  */
 internal object DolbyVisionConversionStats {
     private val codecStringRewriteCount = AtomicLong(0)
+    // DV7 review F5: RPUs dropped because conversion failed (never forwarded raw).
+    private val rpuDropCount = AtomicLong(0)
     @Volatile private var lastSourceProfile: Int? = null
     @Volatile private var lastConversionMode: Int? = null
 
     fun reset() {
         codecStringRewriteCount.set(0)
+        rpuDropCount.set(0)
         lastSourceProfile = null
         lastConversionMode = null
+    }
+
+    fun recordRpuDrop() {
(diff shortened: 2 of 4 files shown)


halibiram:
Nice work on this. I exercised both DV5 and DV7 conversion at runtime and both behaved as expected the P5 mode mapping and conversion path look correct. Spotting that the BlockAdditional RPU was being converted twice and collapsing that to a single conversion site was especially sharp.
it cuts unnecessary native work.

ram130:
Pause merging on this. I'll need to review this week in more detail. 

Mode 2 does not accept profile 5 input per lib docs and per testing. This would create a null situation.

ysosrs123:
> Pause merging on this. I'll need to review this week in more detail.
> 
> Mode 2 does not accept profile 5 input per lib docs and per testing. This would create a null situation.

Thanks for pausing this — genuinely glad you did, because digging into it turned up more than I expected. Short version: mode 2 is safe for Profile 5 in the library we actually ship, but your instinct that something around the mode numbering was off turned out to be right — just not where either of us thought.

First, making sure we're looking at the same library. The vendored header (DV7/libdovi/*/include/libdovi/rpu_parser.h) declares libdovi 3.3.2, and it's byte-identical across all four ABIs and between dev and my branch (md5 8c1ab432029e9c15944cbc71cec4714d). So this is the contract we both ship.

Mode 2 and Profile 5. In that version, dovi_convert_rpu_with_mode dispatches through ConversionMode::from(u8) (dolby_vision/src/rpu/mod.rs), which maps 0 → Lossless, 1 → ToMel, 2 | 3 → To81, 4 → To84. And To81 explicitly accepts profiles 7, 8 and 5 (p5_to_p81() in dovi_rpu.rs). The bundled header even says it in plain words: mode 2 "handles source profiles 5, 7 and 8". So no null situation with the shipped lib.

To rule out the header being newer than the actual .a: the static lib demonstrably contains the ConversionMode enum (its Display symbol and variant strings are right there in the archive), and I went through every libdovi release that has that enum — 1.5.3 through 3.3.2 plus current main. Every one of them behaves identically here. The only versions where mode 2 rejects P5 are 1.5.2 and older, which predate the enum, so they can't be this binary. I suspect that's where the docs/testing you saw came from — in the 1.x days, mode 3 really was the dedicated P5 path and mode 2 really did bail on P5 with exactly the kind of error you described.

Where I got it wrong. My commit message claimed mode 3 meant "static 8.4" in the bundled C API. It doesn't — mode 3 is just a legacy alias of mode 2, so my 3 → 2 change was a behavioural no-op. I'd taken that claim from the header's own per-mode doc comment ("3 = static 8.4, 4 = 8.1 preserve-mapping"), which turns out to be stale and contradicts the shipped code — an upstream libdovi documentation bug. That also means reproduction step 4 in my description (P5 coming out as 8.4/HLG) was inferred from that doc rather than observed on a device. My mistake, and I'm correcting the record here.

I'd still keep P5 on mode 2, though: it's the one value where the header's documented contract and the code agree, and it stays correct even if libdovi ever brings its code in line with its docs (which would quietly turn mode 3 into 8.4).

Fixup pushed. It reverts the Mode 3 / Mode 4 label changes — the original strings match what the library actually does (and the current dovi_tool CLI docs), while my edited ones didn't: someone picking "Mode 4 — 8.1 (preserve mapping)" would actually have got static 8.4. The revert also fixes a side effect I'd missed: the 16 locale copies of those strings were never edited, so English had drifted out of sync with every other language. The fixup also rewrites the conversionMode() KDoc to the verified mapping.

One real bug this shook loose. map_conversion_mode() in dovi_bridge.cpp translates Kotlin mode 5 (the "Preserve DV mapping" toggle) to C-API 4, based on that same stale header comment. The shipped code maps 4 to To84 — so the preserve-mapping path on dev today actually outputs static 8.4, not 8.1 with mapping kept. And no u8 value reaches To81MappingPreserved in any 3.x release, so the C API simply can't do preserve-mapping at all right now. I've kept that out of this PR to avoid scope creep, but I'll open a small separate PR for it (fall back 5 → 2 with a one-time warning). The toggle's wording across the locale files is probably your call after that — reword, grey it out, or drop it.


Small separate PR - https://github.com/NuvioMedia/NuvioTV/pull/2598

ram130:
Review the lib dovi info on their GitHub so you can understand how it works and what's it for. https://github.com/quietvoid/dovi_tool

ysosrs123:
> Review the lib dovi info on their GitHub so you can understand how it works and what's it for. https://github.com/quietvoid/dovi_tool

I did go through the dovi_tool repo — that's actually where the discrepancy comes from.
 
The README's mode table documents the **CLI**. This app doesn't invoke the CLI; it links **libdovi**, and the contract for that is the C header vendored in this repo. From `DV7/libdovi/android-arm64/include/libdovi/rpu_parser.h` (identical across all four ABIs, on `dev` today), the doc for `dovi_convert_rpu_with_mode`:
 
> 2: Converts the RPU to be profile 8.1 compatible. Both luma and chroma mapping curves are set to no-op. **This mode handles source profiles 5, 7 and 8.**
 
And in the library source at the shipped version (`dolby_vision/src/rpu/mod.rs`, tag `libdovi-3.3.2`), the C entry point dispatches through `ConversionMode::from(u8)`, which maps `2 | 3 => ConversionMode::To81` — modes 2 and 3 are aliases for the same conversion — and `To81` accepts profile 5 (`p5_to_p81()` in `dovi_rpu.rs`). That mapping is unchanged in every release from 1.5.3 through current main; I checked each tag. So mode 2 cannot produce a null situation for profile 5 in the library we ship — and if it somehow could, mode 3 would fail identically, because it is the same code path.
 
This is easy to confirm independently: run a current dovi_tool build with `--mode 2` on profile 5 content (it converts to 8.1), or just open the vendored header above in this repo.
 
I'm leaving the line on mode 2, since that's the value the shipped header explicitly documents for profile 5. If you'd prefer 3 for consistency with the CLI README's numbering, that's a one-character, runtime-identical change — no objection from me if you want to make it on merge.

Also happy for you to close this PR and figure it out yourself.

ysosrs123:

PR #2569, opened 2026-07-06
fix(player): resolve AFR preflight bypass and restoration regressions

### LINKED ISSUE #2559: [Bug]: Match content frame rate broken with last update
### Pre-flight checks

### App version / commit

0.7.15-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Amazon Fire TV Stick 4K Max

### Android version

Android 12

### Player mode

Internal player

### Area (tag)

### Bug description

The auto frame rate and resolution has been broken on the last few Nuvio updates. It used to work flawlessly when set to On start/stop.

Now it doesn’t switch the frame rate at all and plays everything in 60fps

### Steps to reproduce

1. Navigate to Settings > Playback > Audio and Video > Video > Auto Frame Rate & Resolution > Set to On start/stop
2. Test with any movie, test at least 2-3 different streams

### Expected behavior

When clicking on a movie stream, the screen should switch to the movies frame rate, typically 24fps.

### Actual behavior

Yet, it doesn’t do this and plays it in 60fps which causes bad playback and sometimes lag in high action scenes. Tested on 2 different devices.

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

Temporary fix looks to be to disable and then enable frame rate switching and it seems to work.

halibiram:
## Summary

This PR resolves a critical regression in the Auto Frame Rate (AFR) matching implementation introduced in recent updates.
Specifically, it addresses:
1. AFR failing to trigger on video startup for network/debrid streams due to a bug in the NextLib bypass check.
2. The Android TV display refresh rate getting stuck at 24Hz after exiting the player due to a Compose key race condition.
3. Stabilizes and improves the frame rate cache hits by utilizing the filename or query-stripped URL as cache keys rather than dynamic temporary debrid links.

## PR type

- [x] Critical bug fix

## Why

1. NextLib is bypassed on authenticated streams because its `MediaInfoBuilder` runs native C/C++ code that cannot forward custom headers (such as authentication tokens or cookies), which causes 403 Forbidden errors and player startup delays. 
   However, the check in `shouldUseNextLibProbe` was set to bypass NextLib if *any* non-blank header is present. Since all extension streams and debrid playbacks include standard headers (`User-Agent` and `Connection`), NextLib was 100% bypassed on standard playbacks, forcing a fallback to the slow `MediaExtractor` probe which consistently timed out, breaking AFR.
2. Keying `DisposableEffect` with `frameRateMatchingMode` in `PlayerScreen.kt` caused the effect to dispose when settings were loaded asynchronously. This ran the cleanup block, clearing the original 60Hz refresh rate cache (`originalModeId`) and leaving the TV stuck at 24Hz on player exit.

## Issue or approval

Fixes #2559

## Reproduction steps

1. Enable Auto Frame Rate (AFR) in NuvioTV settings (set to On start/stop).
2. Play a movie or TV show from a network/debrid source (e.g. RealDebrid).
3. The video will start playing at 60fps instead of 24fps.
4. Exit the player back to the main menu.
5. The TV UI refresh rate remains stuck at 24Hz, causing heavy lag/stuttering in navigation.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

This PR only fixes the bug in `shouldUseNextLibProbe` and the race condition in `PlayerScreen.kt`'s lifecycle. It does not introduce any styling changes, new player controls, or other unrelated behavior tweaks.

## Testing

1. Verified compile check via `./gradlew.bat :app:compileFullDebugSources` successfully.
2. Verified that excluding `User-Agent` and `Connection` headers prevents the NextLib bypass from triggering on standard debrid playbacks.
3. Verified the cache keys are correctly constructed using `currentFilename` and query parameters are successfully stripped.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #2559


skoruppa:
What is the state of it?

halibiram:
> What is the state of it?

I confirmed it with tests as well, the part I regressed should be fixed

PR #2598, opened 2026-07-11
fix(player): preserve-mapping path silently produced static 8.4 output

ysosrs123:
**Summary**

One-line native fix: `map_conversion_mode()` in `dovi_bridge.cpp` translated Kotlin conversion mode 5 ("Preserve DV mapping") to libdovi C-API mode 4. In the shipped library, C-API mode 4 is `To84` (static profile 8.4), not "8.1 preserving mapping" — so the preserve-mapping path silently produced the wrong profile family. Mode 5 now falls back to C-API mode 2 (standard 8.1, mapping not preserved) with a one-time warning log.

**PR type**

- [x] Critical bug fix

**Why**

The 5 -> 4 translation was written against the vendored header's doc comment ("4: Converts to profile 8.1 preserving luma and chroma mapping. Old mode 2 behaviour."). That doc comment is stale relative to the shipped code: in the bundled libdovi (3.3.2 per the header macros `RPU_PARSER_MAJOR/MINOR/PATCH`, byte-identical across all four vendored ABIs), `dovi_convert_rpu_with_mode()` dispatches through `ConversionMode::from(u8)`, which maps `4 -> To84` and anything `>= 5 -> Lossless`; `To81MappingPreserved` is not reachable from any u8 value. I checked every libdovi release containing the `ConversionMode` enum (1.5.3 -> 3.3.2 and current main in quietvoid/dovi_tool): the mapping is identical in all of them, and the vendored `libdovi.a` provably contains that enum (its `Display` symbol and the concatenated variant string block `To MELTo 8.1To 8.4To 8.1, preserving the mapping metadata` are present in the archive).

Net effect before this fix: enabling "Preserve DV mapping" (only reachable with manual "Convert to DV8.1") converted streams to **static profile 8.4** (HLG-flavoured transfer) instead of profile 8.1 with mapping preserved — the wrong profile family and the opposite of what the toggle promises.

Why mode 2 as the fallback: passing 5 through unchanged is not an option (`>= 5` is Lossless — the RPU would be forwarded unconverted under 8.1 signalling, recreating the signalling/metadata mismatch class of bug), and 4 produces the wrong profile. Mode 2 is the standard 8.1 conversion; mapping preservation is simply unavailable until libdovi exposes it through its C API (I can file that upstream against dovi_tool if useful).

**Issue**

Not yet filed — can file one if preferred.

**Reproduction steps**

1. Device on the DV convert path; enable manual "Convert to DV8.1" and the "Preserve DV mapping" toggle.
2. Play a Profile 7 title.
3. Inspect the converted output RPU (e.g. `dovi_tool info` on a captured segment): it reports profile 8.4 / HLG-compatible rather than 8.1. (The on-screen DV badge alone may not distinguish the flavour on all displays; the RPU inspection is the ground truth.)

**UI / behavior impact**

- [x] No UI change
- [ ] No behavior change

**Policy check**

- [x] This PR is small, focused, and limited to one issue. (One switch case in one file.)

**Scope boundaries — deliberately not included**

- The "Preserve DV mapping" toggle's UI strings (17 locale files) still describe mapping preservation. After this fix the toggle is behaviourally inert (falls back to the standard 8.1 conversion). Whether to reword the subtitle, grey the toggle, or remove it is left as a maintainer decision — happy to do whichever in a follow-up.
- No changes to `conversionMode()` in Kotlin — that code is under review in #2557 and this fix is independent of it (mode 5 flows identically before and after that PR).

**Testing**

Verified at source level against the vendored libdovi as described above (header macros, per-tag source audit of quietvoid/dovi_tool, and string/symbol inspection of the vendored `libdovi.a`).

On-device: verified in a downstream fork build (Homatics Box R 4K Plus, Amlogic S905X4, Android 14, LG C9): with manual Convert-to-DV8.1 and the toggle enabled, the one-time `DV7_NATIVE` preserve-mapping warning appears in logcat and playback proceeds on the standard 8.1 conversion path.

Caveat as with #2557: developed in a downstream fork; a maintainer compile-check against clean dev is advised (the change is a single switch case using the file's existing LOGW macro).

**Breaking changes**

None. (Behaviour change is confined to the preserve-mapping path, which previously produced the wrong profile.)

**Linked issues / related**

Related discussion: #2557 review thread (where this was discovered while verifying the mode-numbering question). Independent of that PR — applies cleanly to dev with or without it.

### DIFF
--- a/app/src/main/cpp/dovi_bridge.cpp
+++ b/app/src/main/cpp/dovi_bridge.cpp
@@ -89,8 +89,23 @@ static inline uint8_t map_conversion_mode(jint mode) {
         case 3:
         case 4:
             return static_cast<uint8_t>(mode);
-        case 5:
-            return 4U;
+        case 5: {
+            // Kotlin mode 5 = "preserve DV mapping". The bundled libdovi
+            // 3.3.2 C API has no reachable To81MappingPreserved variant:
+            // ConversionMode::from(u8) maps 4 -> To84 (static 8.4) and
+            // >= 5 -> Lossless, despite the vendored header's doc comment
+            // claiming "4 = 8.1 preserving mapping". The previous 5 -> 4
+            // translation therefore silently produced static 8.4 output on
+            // the preserve-mapping path. Fall back to standard 8.1 (mode 2,
+            // mapping NOT preserved) until libdovi exposes preserve-mapping
+            // through its C API.
+            static bool warned = false;
+            if (!warned) {
+                warned = true;
+                LOGW("DV7_NATIVE: preserve-mapping unsupported by bundled libdovi C API; using standard 8.1 (mode 2)");
+            }
+            return 2U;
+        }
         default:
             return 2U;
     }


ysosrs123:

PR #2632, opened 2026-07-14
Fix watched badge on still-airing series and new episodes disappearing from Continue Watching

### LINKED ISSUE #2631: [Bug]: Completed checkmark shown on series that are still airing, and newly aired episodes vanish from Continue Watching
### Pre-flight checks

### App version / commit

dev @ fbca72af

### Install method

Built from source

### Platform

Android TV

### Device model

NVIDIA Shield TV Pro

### Android version

Android 11

### Player mode

External player

### Area (tag)

Watch Progress (resume, watched state, history)

### Bug description

A series is treated as fully watched as soon as every **aired** episode has been watched, without checking whether the series still has episodes to come. Being caught up on a currently airing show is not the same as having completed it, but the app treats the two identically.

This shows up as two symptoms that share one cause:

1. The completed checkmark appears on the poster of a show that is still airing. Example: a currently airing series with 2 of 12 episodes released, both watched, shows the checkmark as if the whole series were finished.
2. When the next episode airs, it appears in Continue Watching on launch and then disappears a moment later when the row refreshes. The series is flagged fully watched, and the Continue Watching pipeline drops the next-up entry for any series carrying that flag.

Both are driven by the same fully-watched series flag, so they appear and disappear together.

### Steps to reproduce

1. Pick a series that is currently airing and has unaired episodes remaining (for example a weekly anime, 2 of 12 episodes released).
2. Watch every episode that has aired so far, so you are fully caught up.
3. Open the Library and look at that series' poster: the completed checkmark is shown, even though the series is not finished.
4. Wait until the next episode airs, then relaunch the app.
5. Watch Continue Watching on the home screen: the newly aired episode shows up briefly, then disappears when the row refreshes.

### Expected behavior

The completed checkmark should only appear once a series is actually finished, i.e. every episode has aired and every episode has been watched. A series that is still releasing episodes should never be marked completed just because you are caught up.

When a new episode airs for a series you are caught up on, it should appear in Continue Watching and stay there.

### Actual behavior

Being caught up on the aired episodes of an ongoing series marks it fully watched. The poster gets the completed checkmark, and the next episode to air is dropped from Continue Watching after the row refreshes, so it appears on launch and then vanishes.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

- Content type: series
- Example: currently airing weekly anime, 2 of 12 episodes released, both watched
- Not specific to any one addon or source

### Anything else? (optional)

Completion is computed independently in three places, and all three check only whether the aired episodes have been watched:

- `publishBadgeUpdate` in `HomeViewModelContinueWatching.kt`
- `reconcileFullyWatchedFromLocalItems` in `HomeViewModelCatalogPipeline.kt`
- `updateFullyWatchedState` in `MetaDetailsViewModel.kt`

Because the catalog reconcile runs on library/catalog load, it re-adds the flag even if another pass removes it, so fixing only one of the three has no visible effect.

Laskco:
## Summary

A series was marked fully watched as soon as every aired episode had been watched, with no check for whether more episodes were still to come. Being caught up on an ongoing show was treated as having completed it.

This fixes both symptoms of that single cause: the completed checkmark showing on a series that is still airing, and a newly aired episode appearing in Continue Watching on launch and then vanishing when the row refreshes.

## PR type

- [x] Critical bug fix

## Why

Watched state is wrong in two user-visible ways for anyone following a currently airing series, which is the normal case for weekly shows.

Old behavior: a series was flagged fully watched when every aired episode had been watched.

Broken behavior: catching up on an airing series (for example 2 of 12 episodes released, both watched) marked it completed. The poster got the completed checkmark, and because the Continue Watching pipeline drops the next-up entry for any series carrying that flag, the next episode to air was removed from Continue Watching. It appeared on launch from the cached snapshot and then disappeared once the pipeline re-evaluated, so new episodes looked like they were flashing in and out.

New behavior: a series is only completed when it is positively known that nothing is upcoming, so an airing series keeps neither the checkmark nor the Continue Watching drop, and its new episodes stay in Continue Watching.

Completion was computed independently in three places, and all three had the same gap. The catalog reconcile runs on library and catalog load and re-added the flag that the Continue Watching pass had just removed, so fixing any one of them alone had no visible effect.

Manual marking: marking a series watched only marks its aired episodes, which leaves the watch data identical to simply being caught up. Derived evaluation cannot tell the two apart, so tightening it also stripped the badge straight back off after the user set it by hand on an airing series. For series the poster menu derives its mark/unmark state from that same badge, so losing it also removed the option to unmark, leaving the episodes watched with no way to undo it from the library. The series a user marks by hand are now recorded and left out of derived evaluation. The override is dropped again when the series is unmarked, or when the user unwatches an episode so the claim no longer holds.

## Issue or approval

Fixes #2631

## Reproduction steps

1. Pick a series that is currently airing and still has unaired episodes remaining (for example a weekly anime with 2 of 12 episodes released).
2. Watch every episode that has aired, so you are fully caught up.
3. Open the Library and look at that series' poster. The completed checkmark is shown even though the series is not finished.
4. Wait until the next episode airs, then relaunch the app.
5. Watch the Continue Watching row on the home screen. The newly aired episode appears, then disappears when the row refreshes.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Intentionally not changed:

- No UI code, layout, or styling is touched. The completed badge component is unchanged and simply receives correct data.
- Marking a series watched by hand still applies the badge immediately, including while the series is airing. Keeping that working needed a change rather than no change, described under "Manual marking" below.
- The two Continue Watching call sites that only stamp a revalidation deadline are unchanged, since they do not add the flag.
- The existing `fullyWatchedByCount` heuristic for Trakt anime remapping is preserved rather than replaced.
- The Play button on the detail screen restarts an airing series at S1E1 once you are caught up, because no unwatched available episode remains. That is pre-existing and out of scope here.
- `app/src/test/java/com/nuvio/tv/core/trakt/TraktPublicListSourceResolverTest.kt` currently fails to compile on `dev` with an ambiguous `TraktCredentialSyncService` import. It reproduces on a clean checkout without these changes, so it is left alone rather than bundled into this fix.

The only structural change is a local `expandTmdb` helper in `publishBadgeUpdate`, which replaces two copies of an identical ID expansion block because the fix needs a third. It is required by the change rather than a drive-by cleanup.

## Testing

Tested on an NVIDIA Shield TV Pro (Android 11) against a real library of 29 watched series, built from source as `fullDebug` and installed over ADB.

- `./gradlew :app:assembleFullDebug` passes, which is what CI builds.
- Verified with instrumented logging (removed before this commit) that all 29 series resolved and every currently airing series was classified as still airing, while a genuinely finished series was not.
- Airing series with every aired episode watched (for example 2 of 12 released, both watched): the completed checkmark is gone from the Library poster.
- A genuinely finished and fully watched series: the completed checkmark is still shown, confirming the fix does not suppress legitimate badges.
- Opening the detail page of a caught-up airing series and returning to the Library: the checkmark stays gone. Before the detail screen was fixed, this round trip re-added it.
- Continue Watching: a series with a newly aired episode keeps that episode in the row instead of dropping it on refresh.
- Marking an airing series watched by hand from the library poster menu: the badge appears and stays, and "Unmark as watched" remains available on the poster so it can be undone without opening the series. Unmarking clears it again.
- Unwatching an episode of a hand-marked series from the detail screen drops the override, and the series returns to the derived rules.

The unit test task `:app:testFullDebugUnitTest` does not compile on `dev` for the pre-existing Trakt reason noted above, unrelated to these files.

## Screenshots / Video

No UI code, layout, or styling is modified. The only visible difference is the one being fixed: the existing completed checkmark no longer appears on series that are still airing, and a newly aired episode stays in Continue Watching instead of vanishing on refresh.

Recordings from an NVIDIA Shield TV Pro, each showing the behavior before and after the fix.

**New episode staying in Continue Watching**

https://github.com/user-attachments/assets/00c8e3ba-f6d1-4bec-ba6e-48c34425e0bc

**Watched badge no longer shown on a series that is still airing**

https://github.com/user-attachments/assets/01d1cf52-61ae-4c88-9a6d-0a83e40743de

https://github.com/user-attachments/assets/d549a67a-5f68-455c-9526-aec1fc5efc3d

## Breaking changes

None.

`VALIDATION_RESET_VERSION` is bumped from 1 to 2 so that badges already stored on disk are re-evaluated once after upgrade. Without it, existing incorrect badges would persist behind their revalidation deadline. This only forces one re-check of series that already have watch history and does not clear watch progress or change any stored schema.

## Linked issues

Fixes #2631

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/WatchedSeriesStateHolder.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/WatchedSeriesStateHolder.kt
@@ -37,7 +37,9 @@ class WatchedSeriesStateHolder @Inject constructor(
         private val KEY = stringSetPreferencesKey("fully_watched_ids")
         private val REVALIDATE_KEY = stringPreferencesKey("revalidate_after")
         private val VALIDATION_RESET_KEY = intPreferencesKey("validation_reset_version")
-        private const val VALIDATION_RESET_VERSION = 1
+        // v2: re-validate every series once so shows that are still airing drop the
+        // premature "completed" badge they were given while caught up on aired episodes.
+        private const val VALIDATION_RESET_VERSION = 2
         private const val DEFAULT_TTL_MS = 7L * 24 * 60 * 60 * 1000 // 7 days
     }
 
--- a/app/src/main/java/com/nuvio/tv/data/local/WatchedSeriesStateHolder.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/WatchedSeriesStateHolder.kt
@@ -35,6 +35,7 @@ class WatchedSeriesStateHolder @Inject constructor(
     companion object {
         private const val FEATURE = "watched_series_cache"
         private val KEY = stringSetPreferencesKey("fully_watched_ids")
+        private val MANUAL_KEY = stringSetPreferencesKey("manually_marked_ids")
         private val REVALIDATE_KEY = stringPreferencesKey("revalidate_after")
         private val VALIDATION_RESET_KEY = intPreferencesKey("validation_reset_version")
         // v2: re-validate every series once so shows that are still airing drop the
@@ -48,6 +49,16 @@ class WatchedSeriesStateHolder @Inject constructor(
     private val _fullyWatchedSeriesIds = MutableStateFlow<Set<String>>(emptySet())
     val fullyWatchedSeriesIds: StateFlow<Set<String>> = _fullyWatchedSeriesIds.asStateFlow()
 
+    /**
+     * Series the user marked watched by hand. Derived badge evaluation must leave these alone:
+     * a still-airing series never earns the badge on its own, but saying so explicitly is a
+     * deliberate choice and outranks what the episode counts imply.
+     */
+    @Volatile
+    private var manuallyMarkedIds: Set<String> = emptySet()
+
+    fun isManuallyMarked(contentId: String): Boolean = contentId in manuallyMarkedIds
+
     /** Per-series revalidation deadline (contentId → epochMs when re-check is needed). */
     @Volatile
     private var revalidateAfterMap: Map<String, Long> = emptyMap()
@@ -59,6 +70,7 @@ class WatchedSeriesStateHolder @Inject constructor(
         if (loaded) return
         val prefs = store(profileId).data.first()
         val persisted = prefs[KEY] ?: emptySet()
+        manuallyMarkedIds = prefs[MANUAL_KEY] ?: emptySet()
         val resetVersion = prefs[VALIDATION_RESET_KEY] ?: 0
         if (resetVersion < VALIDATION_RESET_VERSION) {
             revalidateAfterMap = emptyMap()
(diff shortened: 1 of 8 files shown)


skoruppa:
Although I appreciate the fix for "new episodes" disappearing from CW the change to mark series as watched when you watched everything that is available was intentional and made by @tapframe to align behavior with the mobile version

PR #2717, opened 2026-07-22
feat(sync): add multi-ID mapping (IMDB, TMDB, TVDB, MAL, Simkl, Kitsu) support to watched items sync

ryan-winkler:
### Overview

This PR adds multi-ID mapping support to `WatchedItem`, `SupabaseWatchedItem`, and `WatchedItemsSyncService` for watched items synchronization.

### Motivation & Background

Currently, watched items sync accepts a single `content_id` string. When syncing watched history across external platforms and bridges (such as Trakt, Simkl, MyAnimeList, and external importers), items may have different IDs across TVDB, TMDB, IMDB, MAL, Kitsu, AniDB, and Simkl.

By passing pre-resolved `ids` maps directly in `sync_push_watched_items` RPC payloads and `WatchedItem`:
1. **Eliminates Runtime Resolution Bottlenecks**: Avoids unnecessary external API calls (e.g. MDBList/TMDB) during sync.
2. **Instant Cross-Platform Matching**: Ensures anime and TV shows with differing TMDB vs MAL IDs match accurately across Nuvio, MAL, and Stremio without rate-limit issues.
3. **Non-Breaking Backwards Compatibility**: Optional `ids` map field; falls back to `content_id` when omitted.

### Changes Made

- **`WatchedItem.kt`**: Added optional `ids: Map<String, String>? = null` field.
- **`SupabaseModels.kt`**: Added `ids` map field to `SupabaseWatchedItem` and `SupabaseWatchedItemEvent`.
- **`WatchedItemsSyncService.kt`**: Updated `pushItemsToRemote` to serialize `ids` into the `p_items` JSON RPC array, and `pullFromRemote` to decode and populate `ids`.

### Server-side companion required

Heads-up that this is the **client half** of a two-part change. For the `ids` map to actually persist and round-trip, the Supabase side needs a matching change:

- an `ids` (jsonb) column on the `watched_items` table,
- `sync_push_watched_items` updated to persist it,
- the pull query/view updated to return it.

Without that, the client will send `ids` in `p_items` but the server will drop it and always return `null` on pull. Happy to align the payload shape to whatever you prefer server-side. Canonical keys I'm sending: `imdb, tmdb, tvdb, mal, kitsu, anidb, simkl`.

### Considerations

- Something to enrich the anime mappings: https://github.com/Goldenfreddy0703/Otaku-Mappings/blob/main/README.md
- Additionally, providing a public endpoint or csv/json upload would simplify the use of watched items sync as well.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/sync/WatchedItemsSyncService.kt
+++ b/app/src/main/java/com/nuvio/tv/core/sync/WatchedItemsSyncService.kt
@@ -150,6 +150,11 @@ class WatchedItemsSyncService @Inject constructor(
                             if (item.episode != null) put("episode", item.episode)
                             else put("episode", JsonPrimitive(null as Int?))
                             put("watched_at", item.watchedAt)
+                            if (item.ids != null && item.ids.isNotEmpty()) {
+                                put("ids", buildJsonObject {
+                                    item.ids.forEach { (k, v) -> put(k, v) }
+                                })
+                            }
                         }
                     }
                 })
@@ -203,9 +208,11 @@ class WatchedItemsSyncService @Inject constructor(
                         title = entry.title,
                         season = entry.season,
                         episode = entry.episode,
-                        watchedAt = entry.watchedAt
+                        watchedAt = entry.watchedAt,
+                        ids = entry.ids
                     )
                 })
+
 
                 if (remote.size < WATCHED_ITEMS_PAGE_SIZE) break
                 page++
--- a/app/src/main/java/com/nuvio/tv/data/remote/supabase/SupabaseModels.kt
+++ b/app/src/main/java/com/nuvio/tv/data/remote/supabase/SupabaseModels.kt
@@ -140,7 +140,8 @@ data class SupabaseWatchedItem(
     val season: Int? = null,
     val episode: Int? = null,
     @SerialName("watched_at") val watchedAt: Long,
-    @SerialName("profile_id") val profileId: Int = 1
+    @SerialName("profile_id") val profileId: Int = 1,
+    val ids: Map<String, String>? = null
 )
 
 @Serializable
@@ -152,8 +153,10 @@ data class SupabaseWatchedItemEvent(
     val title: String = "",
     val season: Int? = null,
     val episode: Int? = null,
-    @SerialName("watched_at") val watchedAt: Long
+    @SerialName("watched_at") val watchedAt: Long,
+    val ids: Map<String, String>? = null
 )
+
 
 @Serializable
 data class SupabaseProfile(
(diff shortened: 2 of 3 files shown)


ryan-winkler:
Follow-up on classification @tapframe / @skoruppa / @CrissZollo / @guuilp — worth deciding now so the schema only changes once.

`contentType` already gives us the movie/series split, so that's covered. What's missing is an **anime flag** (`is_anime: Boolean`, or folding `"anime"` into a richer content type). The reason it matters: an anime film and a live-action film are both `contentType = "movie"`, but they resolve against different ID spaces (MAL/Kitsu/AniDB vs TMDB/IMDB). A flag lets the sync route each to the right mapping source instead of guessing.

Not blocking for this PR — just flagging it alongside the `ids` work so it's a single schema decision rather than two.

skoruppa:
Although it is a nice idea it creates a very problematic issue of "what id to use for meta resolving"

I know that with addons like AIOMetadata returned meta is almost always the same (except for video id for anime when using the "retain" option) but there exist real cases when you want to resolve something by for example MAL ID and sometimes by IMDB ID

And what if you start watching something using MAL id? How to we resolve all other ids? We would need to always rely in the SIMKL api for this. Not to mention that something with one IMDB id can have multiple MAL/Kitsu ids (as each season there is a separate entry). And don't get me started on seasons mismatch between meta providers (look for example how are episode split between seasons for One Piece in imdb/tvdb/tmdb)

Goldenfreddy0703:
Hey there, I'm not sure if I will be updating my mappings anytime soon, ever sense the Jikan api went down, some github projects that uses anime databases have also been archived unfortunately. The one database that was used to make most of my mappings was the anime offline database. 

https://github.com/manami-project/anime-offline-database

Unfortunately it got archived sadly. 
Now if you guys are looking for multiple id's of a move, show, or anime that are mapped. Simkl or Mdblist is where its at. if you guys want to convert anidb to tvdb, use Anime List

https://github.com/Anime-Lists/anime-lists/

Anyway hope this helps and I'm sorry 😅

thisiswinks:
Closed until I have better answers

Goldenfreddy0703:
Hey! Absolutely, I’d be happy to take a look at this and give you some feedback.

I’ve gone through https://github.com/dannyvfilms/Floppy/issues/652 and I can definitely see a lot of overlap with problems I’ve run into while developing Otaku, especially around provider identity, anime season/part structures, episode numbering, cross-provider mappings, watch history, and syncing multiple trackers.

The overall direction of having a provider-neutral canonical identity/mapping layer makes a lot of sense to me. I also really like the idea of separating the canonical entity from how individual providers represent that entity, because anime in particular can get pretty messy when MAL, AniList, TMDB, TVDB, SIMKL, etc. structure the same series differently.

There are quite a few things I’d like to discuss in more depth, though, particularly around episode-level mappings, split/merge relationships, specials/OVAs, season vs. part/cour handling, mapping confidence/provenance, and some of the things I’ve learned from building Otaku and Otaku Mappings.

Rather than dumping a massive wall of text into the issue, I’d be happy to discuss it with you directly on Discord or Keybase and then post any useful architectural feedback back to the issue afterward.

**Discord:** The Steampunk Owl#3126
**Keybase:** Goldenfreddy0703

I appreciate you reaching out and specifically asking for my input. I’d definitely like to help where I can.

PR #2723, opened 2026-07-23
feat(search): Live search while typing, parity from Mobile/Desktop

### LINKED ISSUE #2720: [Feature]: Live search results while typing, matching mobile
### Area (tag)

Other (Search)

### Problem statement

Search on TV only runs when you explicitly submit the query (press Done). While typing, the app fetches suggestions but actively cancels any in-flight search, so no results appear until you commit. On mobile the results narrow as you type, so the two platforms behave differently for the same action, and on a remote the extra "type, then confirm" step is the slowest part of searching.

### Proposed solution

Run the search as you type, debounced, matching mobile:

- Fire the search automatically once the query is at least 2 characters.
- Debounce it so each keystroke doesn't fan out to every enabled addon catalog. A longer debounce than the suggestion debounce is appropriate because a search hits all installed catalogs.
- Keep the explicit submit working: pressing Done still searches immediately and supersedes any pending debounced run.
- Only write to recent searches on an explicit submit, so typing "harry" doesn't leave "h", "ha", "har" in search history.

### Additional context

The existing suggestion path already debounces, so the same pattern applies. The main consideration is request volume: a live search fans out to every enabled addon catalog per debounce tick, unlike a local list filter.

Laskco:
> **Platform parity.** Mobile and desktop both already search as you type. TV is the only client that requires an explicit submit, so this brings it in line rather than introducing new behaviour.

## Summary

Search now runs **as you type**, matching mobile and desktop, instead of only running when the query is explicitly submitted.

Previously `onQueryChanged` fetched suggestions and then actively cancelled any in-flight search ("Search is explicit on submit only"), so no results appeared until you pressed Done. On a remote that confirm step is the slowest part of searching, and it made TV behave differently from mobile/desktop for the same action.

Now the query kicks off a debounced search once it reaches 2 characters. Pressing Done still searches immediately and supersedes any pending run.

**Debounce.** A search fans out to **every enabled addon catalog**, so this is deliberately debounced at 350ms, matching mobile, and longer than the existing 150ms suggestion debounce, to avoid a request storm per keystroke. Each keystroke also cancels the previous in-flight run, so only the latest query is ever resolving.

**Loading state.** The window between a keystroke and results arriving shows skeleton rows rather than the empty/no-results state, the same way mobile does while a typed query is waiting to run. `SearchSkeletonRow` is a dedicated composable picked by screen state, mirroring mobile's `HomeSkeletonRow`, rather than a fabricated `CatalogRow`: a row built from blank names still draws its header decorations, which showed up as a stray " - Movie" and "from " with nothing after them. Skeletons appear when there is nothing real on screen, and a trailing one is appended while further catalogs are still answering. Unlike mobile, results already on screen are kept while a new query settles instead of being swapped out per keystroke: on a remote every letter outlasts the debounce, so that swap reads as flicker.

**Clear button.** The field can now be emptied with a D-pad reachable clear button beside the voice button, instead of holding down delete. With search running live, clearing is a normal part of the flow rather than a rare action.

**Fetching.** A search is identified by the query, the released-content filter and the exact set of catalogs it would hit, and an identical request that already finished or is still arriving is not refetched. Previously, pressing Done after live search had already run that query tore the rows down and refetched everything, and deleting a letter then retyping it did the same. Mobile has had this guard in `SearchRepository`; TV had none. A run cancelled part way is deliberately not counted as satisfied, so it re-runs rather than staying half filled, and Retry always refetches.

**Searchable catalogs.** A catalog that needs a search plus something else that cannot be supplied, a mandatory genre for instance, is no longer queried. It answered with an error or nothing, so it cost one request per keystroke per addon and left a row that never filled in. This matches mobile's `supportsSearch` rule, which requires a search extra and no other required extras.

**Search history.** Queries are recorded once they actually return results, rather than only on an explicit submit, so typing something and backing out still remembers it while typos that match nothing never get saved. Saving also drops any existing entry that is a prefix of the new one, so "f", "fr", "fri" collapse into "frieren" as you type. That prefix rule is TV specific rather than mobile parity: mobile only de-duplicates exact matches, and gets away with it because phone typing outruns the debounce, whereas on a remote every prefix settles long enough to be recorded.

Clearing the field also retires the submitted query. Leaving it set kept the screen in its results state with nothing to show, instead of falling back to recent searches, until the screen was rebuilt by navigating away and back.

**Removed hint.** The "Press Done on the keyboard to search" hint is dropped. It no longer describes how the screen works, and neither mobile nor desktop has an equivalent.

## PR type

- [ ] Critical bug fix

## Why

Mobile and desktop already narrow results while typing, so TV requiring an explicit submit makes the same interaction behave differently depending on which client you pick up. Searching on a remote is also slow enough without an extra confirm step. This is a parity change, not a new capability.

## Issue or approval

Fixes #2720

## Reproduction steps

N/A, not a bug fix. To see the change:

1. Open Search.
2. Start typing a title (at least 2 characters).
3. Shimmer rows appear while the query settles, then results populate on their own, without pressing Done.
4. Keep typing to refine: the previous results stay up until the new ones land, rather than flashing between states.
5. Press Done at any point: it searches immediately.
6. Back out without pressing Done, then reopen Search: the query you typed is in recent searches, listed once rather than as every prefix you passed through.
7. Press the clear button beside the voice button to empty the field.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

UI changes are the shimmer loading rows, the clear button, and the removal of the now-inaccurate keyboard hint. Behaviour changes are when the search runs and when history is written.

## Policy check

- [x] This PR fits the current PR policy: maintainer approved change.
- [x] This PR includes a linked issue and testing notes.

This is platform parity rather than a critical bug fix, so the box asserting no UI or feature change is left unchecked rather than overclaimed.

## Scope boundaries

- `SearchViewModel`: the debounce constant, the live search job, request de-duplication, the searchable-catalog rule, and when history is written.
- `SearchScreen`: the clear button, skeleton rows chosen by screen state, and removal of the keyboard hint block.
- `SearchSkeletonRow`: new, and used only by this screen.
- `CatalogRowSection` is untouched. The loading rows no longer pass through it as fabricated data, so it did not need to learn about them.
- `SearchHistoryDataStore`: prefix collapse on save. Normalization, deduplication and the cap on stored entries are unchanged.
- The suggestion path, the search request/fan-out logic, result rendering, catalog ordering and the Discover path are all unchanged.
- No change to what a search returns, only to when it is triggered and how the wait is presented.

## Testing

- `./gradlew :app:compileFullDebugKotlin`, `./gradlew :app:assembleFullDebug` and `./gradlew :app:lintFullDebug` pass.
- `./gradlew :app:testFullDebugUnitTest`: failures are confined to the same pre-existing, environment-dependent classes (media3, player, tmdb, trakt, datastore, poster options) with and without this change. Nothing in search or search history fails.
- Installed and manually verified on an NVIDIA Shield:
  - Typing at least 2 characters populates results without pressing Done.
  - Pausing mid-word does not fire a request per keystroke; only the settled query resolves.
  - Shimmer appears for the first query of a session; refining an existing search swaps straight from the old results to the new ones without flashing.
  - Pressing Done still searches immediately.
  - Typing a title and backing out without submitting leaves that query in recent searches.
  - Recent searches show the settled query, not the prefixes typed along the way, and a typo that returned nothing is not recorded.
  - The clear button is reachable with the D-pad and empties the field.
  - Clearing back under 2 characters stops results without leaving a stale in-flight search.
  - Pressing the clear button returns to recent searches rather than an empty results view.
  - Pressing Done after live search has already run the query does not refetch or rebuild the rows.
  - Deleting a character and retyping it does not refetch.
  - Retry still refetches after a failed search.

## Screenshots / Video

**Demo (results narrowing while typing):**

https://github.com/user-attachments/assets/24f59191-10ad-4146-be1b-50f12f13739f

## Breaking changes

None.

## Linked issues

Fixes #2720

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/SearchHistoryDataStore.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/SearchHistoryDataStore.kt
@@ -41,7 +41,15 @@ class SearchHistoryDataStore @Inject constructor(
         val current = recentSearches.first()
         val updated = buildList {
             add(normalized)
-            addAll(current.filterNot { it.equals(normalized, ignoreCase = true) })
+            addAll(
+                current.filterNot { existing ->
+                    existing.equals(normalized, ignoreCase = true) ||
+                        // Live search saves each query the user pauses on, and on a remote every
+                        // prefix of a word is one of those. Collapse them into the query actually
+                        // landed on instead of listing "f", "fr", "fri" alongside "frieren".
+                        normalized.startsWith(existing, ignoreCase = true)
+                }
+            )
         }.take(maxItems.coerceAtLeast(1))
 
         store().edit { prefs ->
--- a/app/src/main/java/com/nuvio/tv/ui/components/SearchSkeletonRow.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/components/SearchSkeletonRow.kt
@@ -1,0 +1,100 @@
+package com.nuvio.tv.ui.components
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import com.nuvio.tv.ui.theme.NuvioTheme
+
+private val SKELETON_TITLE_WIDTH = 180.dp
+private val SKELETON_TITLE_HEIGHT = 24.dp
+private val SKELETON_SUBTITLE_WIDTH = 90.dp
+private val SKELETON_SUBTITLE_HEIGHT = 12.dp
+private val SKELETON_BAR_CORNER_RADIUS = 6.dp
+private const val SKELETON_POSTER_COUNT = 8
+
+/**
+ * Loading stand-in for a search result row, drawn from scratch rather than by handing a fabricated
+ * [com.nuvio.tv.domain.model.CatalogRow] to the real row: a row built from blank names still renders
+ * its header decorations, which shows up as a stray " - Movie" and "from " with nothing after them.
+ *
+ * Mirrors the two header lines this screen actually has, a catalog title and the addon name, so the
+ * layout does not shift when the real rows arrive.
+ */
+@Composable
+fun SearchSkeletonRow(
+    posterCardStyle: PosterCardStyle = PosterCardDefaults.Style,
+    showAddonName: Boolean = true,
+    modifier: Modifier = Modifier
+) {
+    val shimmerOffsetState = rememberPlaceholderShimmerOffsetState(label = "searchSkeletonShimmer")
+
+    Column(modifier = modifier) {
+        Column(
+            modifier = Modifier.padding(
+                start = NuvioTheme.spacing.xxxl,
+                end = NuvioTheme.spacing.xxxl,
+                bottom = NuvioTheme.spacing.md
+            ),
+            verticalArrangement = Arrangement.spacedBy(NuvioTheme.spacing.xs)
+        ) {
+            SkeletonBar(
+                width = SKELETON_TITLE_WIDTH,
+                height = SKELETON_TITLE_HEIGHT,
+                shimmerOffsetState = shimmerOffsetState
+            )
+            if (showAddonName) {
+                SkeletonBar(
+                    width = SKELETON_SUBTITLE_WIDTH,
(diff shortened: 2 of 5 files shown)


skoruppa:
I like the idea. But if we are going to do it, we need to see shimmer rows while search is just like on mobile so the whole screen does not flash with each letter (especially with the "Press Done on the keyboard to search" message). A button to clear the input would be also appreciated 

And this needs @tapframe approval before you invest more time here ;)

Laskco:
> I like the idea. But if we are going to do it, we need to see shimmer rows while search is just like on mobile so the whole screen does not flash with each letter (especially with the "Press Done on the keyboard to search" message). A button to clear the input would be also appreciated 
> 
> 
> 
> And this needs @tapframe approval before you invest more time here ;) 

Sounds good, I’ll wait to see what Tap says.

skoruppa:
> > I like the idea. But if we are going to do it, we need to see shimmer rows while search is just like on mobile so the whole screen does not flash with each letter (especially with the "Press Done on the keyboard to search" message). A button to clear the input would be also appreciated
> > And this needs @tapframe approval before you invest more time here ;)
> 
> Sounds good, I’ll wait to see what Tap says.

Approved ;)

PR #2724, opened 2026-07-23
feat(library): Search the cloud library and move Refresh in line with the Saved/Cloud tabs

### LINKED ISSUE #2721: [Feature]: Search the cloud library, and move Refresh in line with the Saved/Cloud tabs
### Area (tag)

UI / Layout

### Problem statement

Two issues in Library → Cloud:

1. **No way to search the cloud library.** The only filters are the provider and type dropdowns. Once an account has a lot of cached items, finding a specific release means scrolling the whole list, even though the item names and file names are already loaded on the client.
2. **The "Refresh cloud library" button is hard to reach.** It sits in its own row between the filter dropdowns and the item grid, and D-pad navigation doesn't route to it reliably, so it can't be pressed. It is also rendered as part of the cloud layout flow rather than sitting with the view controls.

### Proposed solution

1. Add a search field to the cloud view that filters the already-loaded cloud library only. It should be purely local (never trigger a provider request), match against the item name **or** any of its file names, stack on top of the existing provider/type filters, and narrow live as you type. Include a way to clear it.
2. Move "Refresh cloud library" to the top right, in line with the Saved/Cloud tabs, and only render it while the Cloud view is selected. Sitting next to the focusable tabs also makes it reachable with the D-pad.

### Additional context

The cloud items and their file lists are already in memory, so the search is a local filter over `visibleCloudItems` with no extra network cost. Searching file names matters because the useful identifier is often in the filename rather than the item title.

Laskco:
## Summary

Two changes to Library → Cloud:

**1. Search the cloud library.** The cloud view only had provider and type dropdowns, so finding a specific release in a large account meant scrolling the entire list, even though the item names and file names are already loaded on the client. Adds a search field that filters the loaded cloud library only:

- Purely local. It never triggers a provider request, it just narrows `visibleCloudItems`.
- Matches the item name **or any of its file names**, since the useful identifier is often in the filename rather than the item title.
- Stacks on top of the existing provider/type filters.
- Narrows on every keystroke.
- A Clear button appears once there is text.
- Text entry follows the same D-pad pattern as the list editor dialog: the field stays read-only until SELECT opens the keyboard, so arrow keys keep navigating the grid instead of moving a caret.

**2. Refresh cloud library moved in line with the tabs.** It previously sat in its own row between the filter dropdowns and the item grid, where D-pad navigation did not reliably route to it, so it could not be pressed. It now sits at the top right in line with the Saved/Cloud tabs and only renders while the Cloud view is selected. Being adjacent to the focusable tabs also makes it reachable again.

## PR type

- [ ] Critical bug fix

## Why

Once a debrid account has a large cached library, the provider/type filters are not enough to find a specific release, and everything needed to filter it is already in memory. The Refresh button relocation fixes a control that could not be reached with a remote, and puts it with the other view-level controls rather than in the middle of the content flow.

## Issue or approval

Fixes #2721

## Reproduction steps

For the Refresh button (the bug half):

1. Open Library and switch to the Cloud tab.
2. Try to reach "Refresh cloud library" with the D-pad.
3. Before this change focus does not land on it, so it cannot be pressed. After, it sits beside the Saved/Cloud tabs and is reachable.

For the search (the new filter):

1. Open Library → Cloud with a connected provider that has several cached items.
2. Press SELECT on the search field to open the keyboard and type part of a title, or part of a filename.
3. The list narrows as you type, combined with whatever provider/type filters are set.
4. Clear the field to restore the full list.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

Maintainer approved. The Refresh relocation additionally fixes a control that could not be reached with a remote.

## Policy check

- [x] This PR fits the current PR policy: maintainer approved change.

This is a maintainer approved change rather than a pure bug fix, so the box asserting no UI/feature change is left unchecked rather than overclaimed.

## Scope boundaries

- Cloud view only. The Saved view, its filters, sorting, list tabs and genre/year pickers are untouched.
- The search is a client-side filter over already-loaded items. No provider API, request, refresh or caching behaviour is changed.
- Filtering runs through the existing `withVisibleCloudItems()` path, so provider/type counts and validation behave exactly as before.
- The Refresh button's action is unchanged; only its placement and visibility condition changed. The now-unused `CloudLibraryActionsRow` was removed rather than left dead.
- No changes to playback, file resolution or the cloud item cards themselves.

## Testing

- `./gradlew :app:compileFullDebugKotlin` and `./gradlew :app:assembleFullDebug` pass.
- `./gradlew :app:testFullDebugUnitTest` compared against a stashed baseline: identical failures with and without this change (pre-existing/environment-dependent tests), so nothing new is broken by it.
- Installed and manually verified on an NVIDIA Shield with a Torbox account:
  - Refresh is reachable with the D-pad from the tabs and triggers a refresh; it is absent in the Saved view.
  - Typing narrows the grid live; matches hit both release titles and filenames.
  - Search combines correctly with the provider and type dropdowns.
  - Clear restores the full list; leaving and re-entering Cloud behaves normally.
  - Arrow keys still navigate the grid while the search field is focused but not in edit mode.

## Screenshots / Video

**Demo (searching the cloud library, with Refresh in line with the Saved/Cloud tabs):**

https://github.com/user-attachments/assets/9d190275-8345-4618-a046-c46be43150ca

## Breaking changes

None.

## Linked issues

Fixes #2721

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/library/LibraryScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/library/LibraryScreen.kt
@@ -285,6 +285,29 @@ fun LibraryScreen(
                 onSelected = { mode ->
                     viewMode = mode
                     expandedPicker = null
+                },
+                // Refresh belongs to the cloud view only, pinned right in line with the tabs.
+                trailing = if (viewMode == LibraryViewMode.Cloud) {
+                    {
+                        Button(
+                            onClick = viewModel::refreshCloudLibrary,
+                            enabled = !uiState.cloudLibrary.isRefreshing,
+                            colors = ButtonDefaults.colors(
+                                containerColor = NuvioTheme.colors.BackgroundCard,
+                                contentColor = NuvioTheme.colors.TextPrimary
+                            )
+                        ) {
+                            Text(
+                                if (uiState.cloudLibrary.isRefreshing) {
+                                    stringResource(R.string.library_syncing_btn)
+                                } else {
+                                    stringResource(R.string.cloud_library_refresh)
+                                }
+                            )
+                        }
+                    }
+                } else {
+                    null
                 }
             )
         }
@@ -411,9 +434,9 @@ fun LibraryScreen(
             }
 
             item(span = { GridItemSpan(maxLineSpan) }) {
-                CloudLibraryActionsRow(
-                    isRefreshing = uiState.cloudLibrary.isRefreshing,
-                    onRefresh = viewModel::refreshCloudLibrary
+                CloudLibrarySearchRow(
+                    query = uiState.cloudSearchQuery,
+                    onQueryChange = viewModel::onCloudSearchQueryChange
                 )
             }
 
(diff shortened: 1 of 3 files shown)


skoruppa:
Good QOL improvement ;) But the same needs to be prepared for mobile version

Laskco:
> Good QOL improvement ;) But the same needs to be prepared for mobile version 

Will do! I'll get one prepared for mobile.

Laskco:
> Good QOL improvement ;) But the same needs to be prepared for mobile version

Mobile done https://github.com/NuvioMedia/NuvioMobile/pull/1610
I have a Library of 1400 items so scrolling that would actually take years lmao

PR #2730, opened 2026-07-23
feat(simkl): use native anime[] envelope for anime history and list writes

thisiswinks:
## What

Route anime **history** and **list** writes through Simkl's native `anime[]` envelope, making them consistent with the scrobble path and the [Simkl anime guide](https://api.simkl.org/guides/anime).

## Why

`buildSimklScrobbleBody` already does the right thing: flat/absolute anime → the native `anime` wrapper (Path B), seasonal anime → `show` with TVDB cross-mapping (Path A). But `buildHistoryRequest` and `buildSimklListMutationBody` sent **all** anime through `shows[]`. So a native-id anime (`mal`/`anilist`/`anidb`/`kitsu` with a flat absolute episode, and frequently no TMDB/TVDB) was written as a show — the exact case the guide routes through `anime[]`.

## Change

- `SimklMutationBodies.kt`:
  - History: series groups with no seasonal coordinates → `anime[]` (flat episodes); seasonal anime stays in `shows[]` with `use_tvdb_anime_seasons` (unchanged). The split mirrors the scrobble predicate exactly.
  - List: anime → `anime[]`; movies/shows unchanged.
  - `buildShowHistoryItem` is reused untouched — it already emits the flat, flagless DTO for flat episodes.
- No read-path change: the response parser already sums `not_found` across all keys and maps `"anime" -> ANIME`.
- Updated the three affected assertions in `SimklMutationServiceTest`.
- Added `SimklAnimeContractVectorsTest` — golden vectors pinning the Path A/B envelope + payload shapes across history, list, and scrobble.

## Verification note (please read)

I authored this in an environment where I **could not run the Android/Gradle build locally**, so I have not executed the unit tests against this branch. The change is small, reuses existing helpers, and mirrors the already-tested scrobble logic, but please let CI (or a local `:app:testDebugUnitTest --tests "*Simkl*"`) confirm before merging. One item worth validating against a live Simkl account: that a native-only anime (e.g. `mal` id + flat episode) is accepted under `anime[]` for `/sync/history` as the guide describes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/simkl/SimklMutationBodies.kt
+++ b/app/src/main/java/com/nuvio/tv/data/simkl/SimklMutationBodies.kt
@@ -33,8 +33,13 @@ fun buildSimklListMutationBody(
         SimklListMutationRequestDto(
             movies = requestItems.filter { (item, _) -> item.kind == TrackingMediaKind.MOVIE }
                 .map { it.second },
-            shows = requestItems.filter { (item, _) -> item.kind != TrackingMediaKind.MOVIE }
-                .map { it.second }
+            // Anime uses its own native envelope, mirroring the scrobble path
+            // (buildSimklScrobbleBody). See the Simkl anime guide, Path B.
+            anime = requestItems.filter { (item, _) -> item.kind == TrackingMediaKind.ANIME }
+                .map { it.second },
+            shows = requestItems.filter { (item, _) ->
+                item.kind != TrackingMediaKind.MOVIE && item.kind != TrackingMediaKind.ANIME
+            }.map { it.second }
         )
     )
 }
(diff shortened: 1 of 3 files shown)


thisiswinks:
@tapframe  i was working on a plugin before i saw you working on native, and then did some of this work, did some testing and built it out.

I'll keep this as draft just so you can see it, but may be useful for what you're doing.

Pretty much my goals were:
1) have a local cache so if nuvio sync went down , the writes wouldn't disappear
2) Ensure "Anime" is handled fully and can fully expand to SIMKLs full mapping
https://api.simkl.org/guides/anime#how-the-cross-mapping-works

(As a side note: https://github.com/manami-project/anime-offline-database and https://github.com/Anime-Lists/anime-lists/ also exist)

You have really mature integration (would love to have the calendar pull directly!), so I'm stopping my work, but lemme know if you need help testing.

Issue #2599, opened 2026-07-11
[Bug Subtitles font size and position not working


drazenfirst-ui:
I### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.7.16 

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Samsung TV

### Android version

-

### Player mode

Internal player

### Area (tag)

Subtitles (download, styling, sync)

### Bug description

The settings for font size and position changes, but are not applied. The font size stays the same and is almost unreadable, because of the very small default font size.

### Steps to reproduce

Open subtitles options from menu. Click + or - on font size. The value changes, but are not applied. Same for subtitle position. 

### Expected behavior

Change of subtitle font size and position 

### Actual behavior

Subtitles font size and position not applied 

### Frequency

Always

### Did this work before?

No, it never worked

### Logs (required for crash reports)

```shell

```

PR #2770, opened 2026-07-27
fix(player): add FFmpeg software audio decode to recovery ladder

### LINKED ISSUE #2608: [Bug]: Video plays with no audio when the current engine can't play the stream's audio track — no notification, engine switch fixes it
### Pre-flight checks

### App version / commit

0.7.16-beta (latest GitHub release)

### Install method

GitHub Release APK

### Platform

Fire TV

### Device model

Fire TV (streaming stick)

### Android version

Fire OS

### Player mode

Internal player

### Area (tag)

Audio tracks

### Bug description

Playing a series episode from a Blu-ray remux-type source with the internal player on the default engine (ExoPlayer): video plays normally but there is NO audio at all — no error, no toast, nothing. Pressing the engine-switch (⇄) button in the OSD makes audio work immediately on the other engine (libmpv decodes it in software).

Full code analysis is in #2605 (closed by the labels bot, re-filed here as instructed). Summary: two silent paths — (1) unsupported audio codec (e.g. TrueHD/DTS-HD MA on Fire TV): ExoPlayer never selects the audio track and no onPlayerError fires; updateAvailableTracks has a 'present but unselected' check for VC-1 video but nothing equivalent for audio. (2) the error-recovery ladder's audio-disabled rung (after safe-audio mode fails) silently drops audio and keeps playing video. Neither path informs the user or considers the other engine.

### Steps to reproduce

1. Use a device with no decoder/passthrough for a lossless audio codec (e.g. Fire TV + TrueHD/DTS-HD MA)
2. Open a series episode and pick a Blu-ray remux source with such an audio track
3. Play with the internal player on the default engine (ExoPlayer)
4. Observe: video plays, no audio, no error or notice
5. Press the engine-switch (⇄) button in the player OSD
6. Observe: audio works on the other engine

### Expected behavior

If the current engine can't play the stream's audio, the player should not just play silently: either auto-switch to the other internal engine, or tell the user (switch engine / continue without audio), ideally rememberable as a preference.

### Actual behavior

Video plays with no audio, indefinitely, with no notification. Audio only works after manually pressing the engine-switch (⇄) button, which most users won't associate with fixing sound.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

- Content type: series episode
- Stream/source: Blu-ray remux-type source
- Video played fine; only audio was missing until the engine switch.

```shell

```

### Anything else? (optional)

Original report with full code analysis and a proposed fix: #2605 (closed by the labels bot because API-created issues can't carry labels without triage permission). A patch is prepared — engine-switch prompt / persisted auto-switch preference reusing the existing switch machinery, with a per-stream one-shot guard — and a PR against dev can follow if this is accepted.

supergera13:
## Summary

Adds FFmpeg software audio decoding as a recovery step between SafeAudio and AudioDisabled in the ExoPlayer recovery ladder. When hardware audio decode fails and the safe-audio retry doesn't help, the player now tries FFmpeg-based software audio decoding before falling back to audio-disabled mode.

The fix sets `hasTriedAudioPcmFallback = true` in all three recovery paths (DECODING_FAILED, AudioTrack failure, StuckPlayingNoProgress), triggering the existing `effectiveDecoderPriority` path that switches to `EXTENSION_RENDERER_MODE_PREFER` — which is audio-only FFmpeg decode, matching the behavior of the "Prefer hardware" decoder setting.

## Why

The recovery ladder previously went straight from safe-audio to audio-disabled. The FFmpeg fallback existed in `effectiveDecoderPriority` but was never triggered from the recovery retry logic — `hasTriedAudioPcmFallback` was only set via the PCM audio detection path, not from recovery retries.

## Scope

- FFmpeg audio decode step only — no engine switch to libmpv
- No new settings or UI
- Recovery ladder only: DECODING_FAILED, AudioTrack failure, StuckPlaying paths
- 3 lines added per path

## Testing

- [x] Code compiles
- [x] Existing unit tests pass
- Device testing: play TrueHD/DTS-HD MA stream, verify audio recovers via FFmpeg after decode failure

Fixes #2608

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
@@ -1321,6 +1321,11 @@ internal fun PlayerRuntimeController.initializePlayer(
                                 retryCurrentStreamWithSafeAudioFallback(currentPosition)
                                 return
                             }
+                            if (!hasTriedAudioPcmFallback) {
+                                hasTriedAudioPcmFallback = true
+                                retryCurrentStreamWithSafeAudioFallback(currentPosition)
+                                return
+                            }
                             if (!isAudioDisabledForCurrentPlayback) {
                                 audioDisabledForcedStreamUrls.add(currentStreamUrl)
                                 retryCurrentStreamWithAudioDisabled(currentPosition)
@@ -1339,6 +1344,11 @@ internal fun PlayerRuntimeController.initializePlayer(
                                 retryCurrentStreamWithSafeAudioFallback(currentPosition)
                                 return
                             }
+                            if (!hasTriedAudioPcmFallback) {
+                                hasTriedAudioPcmFallback = true
+                                retryCurrentStreamWithSafeAudioFallback(currentPosition)
+                                return
+                            }
                             if (!isAudioDisabledForCurrentPlayback) {
                                 audioDisabledForcedStreamUrls.add(currentStreamUrl)
                                 retryCurrentStreamWithAudioDisabled(currentPosition)
@@ -1349,6 +1359,11 @@ internal fun PlayerRuntimeController.initializePlayer(
                         if (error.isStuckPlayingNoProgress()) {
                             if (!isSafeAudioModeActiveForCurrentPlayback) {
                                 safeAudioForcedStreamUrls.add(currentStreamUrl)
+                                retryCurrentStreamWithSafeAudioFallback(currentPosition)
+                                return
+                            }
+                            if (!hasTriedAudioPcmFallback) {
+                                hasTriedAudioPcmFallback = true
                                 retryCurrentStreamWithSafeAudioFallback(currentPosition)
                                 return
                             }


skoruppa:
Shouldn't that be controlled by the "switch engine on failure" setting?

supergera13:
@skoruppa That's a fair question. The reason I didn't gate it behind the "switch engine on failure" setting is that this is a targeted audio-recovery switch, not a general startup-error switch.

The existing `autoSwitchInternalPlayerOnErrorEnabled` setting controls `maybeAutoSwitchInternalPlayerOnStartupError()` — which fires on _startup_ errors (stream won't load, decoder init failure). That's a broader, more disruptive switch that users might want to disable (e.g. if they prefer ExoPlayer and would rather see an error than auto-switch to libmpv).

The audio failover is different: the user is already watching video, audio is about to be silently disabled, and the only alternative is silent video with no notification. Silent video is never an acceptable user experience — there's no reason a user would want to stay on a codec they can't hear rather than switching to libmpv which decodes all audio in software.

That said, if you prefer it to be gated, I can add a check for the setting. But I think the user setting is meant for startup errors, not for the audio recovery ladder (which already has its own internal escalation: safe-audio → libmpv → audio-disabled).

skoruppa:
I understand. But I'm still not sure. MPV engine does not work too well for most device on resolutions higher than 1080p (and even that is usually a problem). There is a dedicated button to switch the engine on such cases and if that switch is not forwarded to next episode - this is maybe something worth a fix. 

As to the main issue - have you tried forcing software decoders for exo in settings? Because if that works but does not work with "prefer hardware decoder" maybe our fallback doesn't work correctly for your case, and it is something that we should look into

supergera13:
@skoruppa Good points, let me address both:

**1. MPV above 1080p:**
You're right, MPV struggles above 1080p on most devices. I'll add a resolution guard — `switchEngineOnAudioFailure()` should check the current video resolution and skip the auto-switch if it exceeds 1080p. In that case, the user can still manually switch via the dedicated button (which is the right UX since they can see the tradeoff). Would that address your concern?

**2. Software decoders:**
The issue we're targeting is specifically when ExoPlayer's audio decoder fails entirely (codec not supported, tunneling fails, etc.) — the safe-audio fallback in `PlayerRuntimeControllerErrorRecovery` tries constrained channels + no passthrough, but if the codec itself isn't supported by the hardware decoder, software decoding for audio would help. However, forcing software decoders globally affects video too, which has performance implications. Our PR targets the case where audio codec support is the bottleneck — libmpv handles all audio codecs in software natively, so it's the right tool for this specific failure mode.

That said, if there's a way to force software decoding for audio only while keeping hardware video decode, that would be a better first step before switching engines. Is that something NuvioTV currently supports?

**3. Forwarding engine switch to next episode:**
Agreed — that's a separate issue worth fixing. I'll open a new issue/PR for propagating the manual engine switch to the next episode in a binge group.

skoruppa:
1. No, we will for sure not add this >< if the setting to auto switch engine is not enabled I don't ever want to mpv be auto switched
2. There is no "software video decoder path" in the app. Selecting "software decoder" in settings is only for audio. Such option already exist and you didn't answer my question. Did you try it?

supergera13:
@skoruppa Thanks for the clarification — I didn't realize the "software decoder" setting was audio-only. That changes things significantly.

1. **Understood** — no auto-switch to MPV without the user setting enabled. I'll gate it behind `autoSwitchInternalPlayerOnErrorEnabled`.

2. **I haven't tested with the software decoder setting enabled** — that's the key question. If enabling "software decoder" (audio-only) in settings resolves the audio failure case, then the real fix might be different: instead of switching engines, the recovery ladder should automatically try software audio decoding before escalating to engine switch or audio-disabled mode.

So the question becomes: **does the existing safe-audio fallback in `PlayerRuntimeControllerErrorRecovery` already try the software decoder path, or does it only constrain channels/passthrough while still using hardware audio decode?** If it doesn't try software audio decode, that's the missing step — and adding it to the recovery ladder would be a better fix than switching to MPV.

I'll investigate the recovery code to check if `safeAudioForcedStreamUrls` enables software audio decoding or just constrains the audio parameters. If it doesn't, I'll rework this PR to add software audio decode as a step before the engine switch (which stays gated behind the setting).

supergera13:
@skoruppa I investigated the recovery code and here's what I found:

**The recovery ladder currently does NOT try software audio decoding automatically.** Here's the flow:

1. `safeAudioForcedStreamUrls` → retry with tunneling off + constrained channels — but `decoderPriority` stays at `playerSettings.decoderPriority` (hardware by default)
2. `audioDisabledForcedStreamUrls` → disables audio entirely
3. (our PR) → switch to MPV

The "software decoder" setting (`playerSettings.decoderPriority`) maps to `EXTENSION_RENDERER_MODE_PREFER` which makes FFmpeg audio decode first, but **this is only applied if the user manually enables it in settings** or if `hasTriedAudioPcmFallback` is true (which only happens through a different PCM fallback path, not through the safe-audio recovery).

So there's a gap: when audio hardware decoding fails, the ladder tries safe-audio (no tunneling + channel constraints) but never escalates to software FFmpeg decode before jumping straight to "disable audio."

**Proposed rework for this PR:**

Instead of switching to MPV, add a new recovery step between safe-audio and audio-disabled:
- After safe-audio fails → set a flag like `softwareAudioDecoderForcedStreamUrls` → reinit with `EXTENSION_RENDERER_MODE_PREFER` (which is audio-only FFmpeg)
- Only if that also fails → fall through to audio-disabled
- MPV switch stays as the absolute last resort, gated behind `autoSwitchInternalPlayerOnErrorEnabled`

This way: safe-audio → FFmpeg audio decode → audio-disabled → (if setting enabled) MPV switch.

Would you accept this approach? It fills the gap without auto-switching engines, and matches what the user gets by manually enabling "software decoder" — but automated in the recovery ladder.

skoruppa:
But there are 3 settings. "Use hardware only" , "use software only" and "prefer hardware". The third option should behave as you described, and auto switch to software path where audio codec is not present or audio failed.

The additional fallback to mpv if even the ffmleg software decoder fails can be added but must be gated behind the "auto engine switch" setting

supergera13:
@skoruppa Perfect, that's clear. So the fix should be:

1. **"Prefer hardware" mode** should already auto-fallback to FFmpeg software audio decode when hardware fails — if it currently doesn't do this in the recovery ladder, that's the bug to fix. I'll rework the PR to make the safe-audio recovery path actually honor the "prefer hardware" setting by trying `EXTENSION_RENDERER_MODE_PREFER` before escalating to audio-disabled.

2. **MPV fallback** as the absolute last resort (FFmpeg software audio also failed), gated behind `autoSwitchInternalPlayerOnErrorEnabled`.

I'll rework the PR with this approach. Thanks for the direction.

supergera13:
Updated the recovery ladder as discussed: added FFmpeg software audio decode fallback (hasTriedAudioPcmFallback) between SafeAudio and AudioDisabled in all three recovery paths (DECODING_FAILED, AudioTrack failure, StuckPlayingNoProgress).

The fix adds 3 lines in each path — when SafeAudio fails and PCM/FFmpeg fallback hasn't been tried yet, sets hasTriedAudioPcmFallback = true and reinitializes. The existing player initialization code at line 756 already picks up this flag to set EXTENSION_RENDERER_MODE_PREFER, so ExoPlayer will prefer the FFmpeg extension for audio decoding.

This matches the expected behavior of the 'Prefer hardware' decoder setting: auto-fallback to FFmpeg software decode when the hardware decoder can't handle the audio codec, before giving up entirely (AudioDisabled) or switching to MPV.

app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerScreen.kt
```diff
@@ -1376,6 +1377,7 @@ private fun ExoPlayerSurface(
 ) {
     val context = LocalContext.current
     val latestAspectMode by rememberUpdatedState(aspectMode)
+    val latestSubtitleStyle by rememberUpdatedState(subtitleStyle)
```
skoruppa, line 1381: again, things from your other PRs are leaking

supergera13:
@skoruppa You're right — the subtitle commits from #2599 had leaked into this branch. I've rebased it so only the audio failover recovery ladder commit remains. Should be clean now.

skoruppa:
Looks good. Just update PR description and title

PR #2749, opened 2026-07-25
fix(stream): keep scroll position per provider

hxreborn:
## Summary

Each provider tab keeps its own scroll position now. `RightStreamSection` passes `StreamsList` a `LazyListState` keyed on `selectedAddonFilter`, instead of the list quietly making one for itself.

## PR type

- [ ] Translation/localization only
- [x] Critical bug fix

## Why

All the provider tabs on the Sources screen share one scroll position, so switching tabs drops you at the last tab's offset instead of that provider's own. Full write-up and videos in #2538.

## Issue or approval

Fixes #2538

## Reproduction steps

1. Open a title with two providers installed and go to the Sources screen.
2. On the first provider's tab, scroll the list down to the bottom.
3. Switch to the second provider's tab.
4. The second provider's list opens scrolled to the position from the first tab.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change
- [x] UI changed only to fix a documented glitch/bug
- [x] Behavior changed only to fix a documented bug/regression
- [ ] UI change has explicit maintainer approval
- [ ] Behavior change has explicit maintainer approval

## Policy check

- [x] I have read and understood `CONTRIBUTING.md`.
- [x] This PR fits the current PR policy: localization/translation only or a critical bug fix.
- [x] This PR does not add features, UI changes, refactors, or other non-critical changes.
- [x] This PR is small, focused, and limited to one issue.
- [x] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
- [x] This PR includes a linked issue, reproduction steps, and testing notes if it is a critical bug fix.
- [x] I listed the testing performed below.

## Scope boundaries

Just the LazyListState ownership. Didn't touch ordering, filtering, the chips, or focus handling, and no styling moved. `StreamsList` still defaults to `rememberLazyListState()` so nothing else calling it notices.

## Testing

TCL C805 (G08_4K_GB), Android 14. 0.7.14-beta release APK, internal player, built on top of dev@7d735089.

Used two demo addons that just return numbered placeholder streams so the offset is obvious, 12 in Demo Alpha and 8 in Demo Beta. Scrolled Alpha down to 12, tabbed over to Beta, Beta came up at its own first stream. Went back to Alpha, still on 12. Did the same on a few other titles and provider combos and got the same thing every time. Focus order on the tabs and on the list is unchanged, and I didn't get stuck anywhere with the d-pad.

## Screenshots / Video

Before and after clips are both on #2538, recorded on the TCL.

[...]

PR #2776, opened 2026-07-27
fix(streams): reset scroll position when switching source tabs (#2538)

### LINKED ISSUE #2538: [Bug]: Streams screen keeps the previous source's scroll position when switching source tabs
### Pre-flight checks

### App version / commit

0.7.14-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

TCL C805 (G08_4K_GB)

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

UI / Layout

### Bug description

The Sources screen has one tab per provider at the top. The stream list keeps a single scroll position shared across all tabs instead of one per tab. Scrolling down one provider's list and switching to another provider's tab carries that scroll position onto the second list, so it opens where the first tab was rather than at that provider's own position. Switching back does not restore where you were. It happens with any title and any providers.

### Steps to reproduce

1. Open a title with two providers installed and go to the Sources screen.
2. On the first provider's tab, scroll the list down to the bottom.
3. Switch to the second provider's tab.
4. The second provider's list opens scrolled to the position from the first tab.

### Expected behavior

Each provider's tab shows its list from the top, or its own last position. Switching tabs does not carry one provider's scroll position onto another provider's list.

### Actual behavior

All tabs share one scroll position. Switching to another provider opens its list at the previous tab's scroll position instead of its own. Only the last tab you scrolled sits where you left it.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

Not content-specific. Reproduces with any title and any two providers.

```shell

```

### Anything else? (optional)

Two recordings attached, both on top of dev@7d735089, using demo addons that return numbered placeholder streams (Demo Alpha: 12, Demo Beta: 8) so the scroll position is easy to read

- First video (before): I scroll down the Demo Alpha list, switch to Demo Beta, and Demo Beta opens at Demo Alpha's scroll position instead of its own.

	https://github.com/user-attachments/assets/7740d18d-b7df-4ab1-8189-10a5976c3d68

- Second video (after): the same steps on a build compiled with my commit [6587dda](https://github.com/hxreborn/NuvioTV/commit/6587dda7e5fe280ee0bd42dc0249137d0a4a1024), which gives each source provider its own scroll state. Demo Beta opens at its first stream, and switching back to Demo Alpha returns to 12.

	https://github.com/user-attachments/assets/8636d3df-65d0-498f-918e-adf902b994e6

I'd gladly open a PR if needed.

supergera13:
## Summary

The streams/sources list now scrolls back to the top when switching between addon filter tabs, instead of carrying the previous tab's scroll position.

## PR type

- [x] Critical bug fix

## Why

The Streams screen uses filter chips to switch between addon sources (e.g. Torrentio, MediaFusion). All tabs shared a single implicit `LazyColumn` scroll state. Scrolling down one provider's stream list and switching to another tab kept the same scroll offset, showing the second provider's list already scrolled far down (or showing empty space if it had fewer items).

`StreamsList` composable used the default `rememberLazyListState()` without explicit keying. Since the `LazyColumn` persists across tab switches (same composable, different items), the scroll state was never reset.

## Issue or approval

Fixes #2538

## Reproduction steps

1. Open a title with two or more installed source addons
2. On the first addon's tab, scroll the stream list down several items
3. Switch to the second addon's tab
4. Observe: the list is already scrolled down, showing empty space or the wrong items
5. Expected: list should start from the top for each addon

## UI / behavior impact

## Policy check

## Scope boundaries

- Only `StreamScreen.kt` — `StreamsList` function: explicit `LazyListState` + scroll reset on filter change (+8 lines)
- No changes to the filter chip UI, stream item rendering, or data loading logic

## Testing

**Compilation verified:** `./gradlew :app:compileFullDebugKotlin` — BUILD SUCCESSFUL

**Device smoke test needed:**
1. Open a title with two installed source addons
2. On the first addon's tab, scroll the stream list down
3. Switch to the second addon's tab using Left/Right
4. Observe: the list should start from the top, not carry the previous scroll position
5. Switch back to the first addon — list should also start from the top

## Screenshots / Video

Before/after comparison of scroll position reset. The fix addresses a visual glitch where an empty or wrong section of the list was visible after switching tabs.

## Breaking changes

None.

## Linked issues

Fixes #2538

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/stream/StreamScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/stream/StreamScreen.kt
@@ -61,6 +61,7 @@ import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.lazy.itemsIndexed
@@ -1026,8 +1027,14 @@ private fun StreamsList(
     val firstCardFocusRequester = remember { FocusRequester() }
     val lastKeyRepeatDispatchRef = remember { java.util.concurrent.atomic.AtomicLong(0L) }
     val restoreFocusRequester = remember { FocusRequester() }
+    val streamListState = rememberLazyListState()
     val firstStreamKey = streams.firstOrNull()?.let { first ->
         "${first.addonName}_${first.url ?: first.infoHash ?: first.ytId ?: "unknown"}"
+    }
+
+    // Reset scroll position to the top when the addon filter changes (#2538).
+    LaunchedEffect(selectedAddonFilter) {
+        streamListState.scrollToItem(0)
     }
 
     LaunchedEffect(requestInitialFocus, firstStreamKey) {
@@ -1055,6 +1062,7 @@ private fun StreamsList(
     }
 
     LazyColumn(
+        state = streamListState,
         modifier = Modifier
             .fillMaxSize()
             .padding(NuvioTheme.spacing.lg)


skoruppa:
isn't that the same what #2749 did? ><

supergera13:
@skoruppa Close, but not the same — they're complementary approaches:

- **#2749** keeps a separate scroll position _per provider_ (each tab remembers where you were)
- **#2776** resets scroll to the top when switching tabs

#2538 (the issue this PR fixes) specifically reports that switching tabs keeps the _previous_ tab's scroll position, which feels wrong — the user expects to start from the top of the new provider's list.

If you prefer #2749's approach (persistent per-provider scroll), this PR becomes redundant. But if the expected behavior is "reset to top on tab switch" (which is what #2538 asks for), then this is the right fix.

Happy to close this in favor of #2749 if that's the behavior you want. Or if #2749 is stale/inactive, I can rebase this one.

skoruppa:
Good explanation. After thinking on it, I decided to go with your fix

PR #2764, opened 2026-07-27
fix(player): keep next-episode on the same addon source (#2756)

supergera13:
## Summary

When auto-playing the next episode inside the player, prefer streams from the same addon that is currently playing (e.g. stay on **Torrentio TV** instead of jumping to **Torrentio** because it is earlier in the installed-addon order), while still honoring binge-group continuity.

## PR type

- [x] Critical bug fix

## Why

`playNextEpisode` auto-selects via `StreamAutoPlaySelector` using installed-addon order, optional binge-group preference, and global autoplay filters. It did **not** consider the addon of the stream the user actually started.

With both Torrentio and Torrentio TV installed (or the same addon mirrored with different debrid), FIRST_STREAM / ordered selection often returns the earlier addon. After next-episode autoplay, the source silently switches away from the addon the user chose even when binge groups match.

## Issue or approval

Fixes #2756

## Reproduction steps

1. Install both **Torrentio** and **Torrentio TV** (with debrid)
2. Open a series, pick a stream from **Torrentio TV**, start internal player
3. Trigger next episode / auto-play from inside the player
4. Before: next episode can resolve from **Torrentio**
5. After: next episode stays on **Torrentio TV** when that addon returns a matching/playable stream; binge-group still preferred when the preferred addon has no matching binge

## UI / behavior impact

- [x] Behavior changed only to fix a documented bug/regression
- [x] No UI change

## Policy check

- [x] I have read and understood `CONTRIBUTING.md`.
- [x] This PR fits the current PR policy: localization/translation only or a critical bug fix.
- [x] This PR does not add features, UI changes, refactors, or other non-critical changes.
- [x] This PR is small, focused, and limited to one issue.
- [x] This PR does not bundle unrelated refactors, cleanups, formatting, or drive-by changes.
- [x] This PR includes a linked issue, reproduction steps, and testing notes if it is a critical bug fix.
- [x] I listed the testing performed below.

## Scope boundaries

- Only the autoplay stream selection logic used by in-player next-episode
- No new settings UI
- Selection priority:
  1. same addon + same binge group
  2. any addon with same binge group
  3. same addon via FIRST_STREAM / REGEX
  4. global mode fallback
- Stream-screen autoplay outside in-player next-episode is unchanged
- X-axis / stream-index preservation is **out of scope** (binge groups cover continuity per maintainer; #2778 closed)

## Testing

```
[...]

Issue #2756, opened 2026-07-26
[Bug]: Next episode in player switches from Torrentio TV to Torrentio (source preference lost)


Burgevi:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.7.20

### Install method

GitHub Release APK

### Platform

Google TV

### Device model

TCL T8C

### Android version

Android 12

### Player mode

Internal player

### Area (tag)

Next Episode / Auto-play

### Bug description

I have both **Torrentio** and **Torrentio TV** (with debrid) installed

I manually pick a stream from **Torrentio TV** and start playback. When the player advances to the **next episode** (auto-play / in-player next episode), the selected source/addon is **not preserved**: the next episode resolves/plays from **Torrentio** instead of staying on **Torrentio TV**.

This resets my preferred addon between episodes and often picks a different stream quality/provider than the one I started with.

### Steps to reproduce

1. Install/enable both **Torrentio** and **Torrentio TV with debrid**.
2. Open any series detail page.
3. Open streams for an episode and explicitly select a stream from **Torrentio TV**.
4. Start playback in the **internal player**.
5. Let the episode finish, or trigger **next episode / auto-play** from inside the player.
6. Observe which addon/source is used for the next episode.

### Expected behavior

The next episode should keep the same preferred source/addon context that was used for the current episode (**Torrentio TV**), or at least prefer the same addon when resolving the next stream.

### Actual behavior

After switching to the next episode inside the player, the source resets from **Torrentio TV** to **Torrentio**.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

- Content type: series
- Addons: Torrentio + Torrentio TV (debrid)
- Started with: Torrentio TV stream
- After next episode: switches to Torrentio
- Player: Internal

### Logs (required for crash reports)

```shell

```

PR #2778, opened 2026-07-27
fix(player): preserve stream selection (X-axis) across next-episode autoplay (#2756)

supergera13:
## Summary

Follow-up to PR #2764 (Y-axis: addon preference). This PR adds X-axis preservation: the next episode picks the **same stream** (by name) within the preferred addon, not just any stream from the addon.

This matches Stremio behavior as reported by @haveAnIssue: if the user chose the 2nd stream in Torrentio, next-episode autoplay should pick the same stream in Torrentio for the next episode.

## PR type

- [x] Critical bug fix

## Why

After PR #2764, next-episode stays on the same addon (e.g. Torrentio TV). But within that addon, it picks the first stream by the auto-play mode (usually FIRST_STREAM = top result). If the user deliberately chose the 2nd or 3rd stream (different quality/indexer), next-episode drops them back to the top stream.

`selectAutoPlayStream()` with `preferredAddonName` filters to the correct addon but then uses the standard `selectFromCandidates()` logic (FIRST_STREAM / REGEX / binge group), which doesn't consider the user's positional choice.

## Issue or approval

Enhances fix for #2756 (X-axis preservation per @haveAnIssue feedback on PR #2764)

## Reproduction steps

1. Install Torrentio addon with multiple streams per episode
2. Open a series and pick the 2nd or 3rd stream in Torrentio (not the top result)
3. Start internal player and let next-episode autoplay trigger
4. Observe: next episode picks the top stream again, not the one the user chose
5. With fix: next episode picks the same stream name within Torrentio if available

## UI / behavior impact

- [x] No UI change

## Policy check

## Scope boundaries

- `StreamAutoPlaySelector.kt` — new `preferredStreamName` parameter + name matching logic
- `PlayerRuntimeControllerStreams.kt` — passes current stream name to auto-play selection
- No changes to binge-group UI, settings, or addon management

## Testing

**Unit tests:** `StreamAutoPlaySelectorTest` — all 9 tests PASS (BUILD SUCCESSFUL)

**Static analysis / code path verification:**

1. User plays Episode 1 -> selects "Torrentio 1080p DTS remux" (2nd stream in Torrentio addon)
2. `currentStreamName` = "Torrentio 1080p DTS remux", `currentAddonName` = "Torrentio"
3. Episode 1 ends -> `autoPlayNextEpisode()` fires
4. `preferredAddonForNext` = "Torrentio", `preferredStreamNameForNext` = "Torrentio 1080p DTS remux"
5. Next episode streams arrive -> `selectAutoPlayStream()` filters to Torrentio candidates
6. Name match found: returns the stream with the same name
7. Episode 2 starts on the same stream the user chose

**Fallback behavior:**
- If the exact stream name doesn't exist for the next episode -> falls through to `selectFromCandidates()` (picks best available in the addon)
- If the addon has no streams for next episode -> falls through to global selection
- `preferredStreamName = null` -> no name matching attempted, existing behavior unchanged

## Screenshots / Video

Not a UI change

## Breaking changes

None.

## Linked issues

Enhances fix for #2756

---
*Note: CI build failed due to Java heap space OOM on the GitHub Actions runner, not a code issue. This edit triggers a rebuild.*

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/player/StreamAutoPlaySelector.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/StreamAutoPlaySelector.kt
@@ -62,7 +62,14 @@ object StreamAutoPlaySelector {
          * ranking. Fixes next-episode jumping from e.g. "Torrentio TV" to
          * "Torrentio" just because the latter is earlier in installed order.
          */
-        preferredAddonName: String? = null
+        preferredAddonName: String? = null,
+        /**
+         * When set (typically the name of the stream currently playing), prefer
+         * a stream with the same name within the preferred addon. This preserves
+         * the user's stream choice (quality/indexer) across episodes — matching
+         * Stremio behavior where next-episode picks the same stream position.
+         */
+        preferredStreamName: String? = null
     ): Stream? {
         if (streams.isEmpty()) return null
 
@@ -90,6 +97,16 @@ object StreamAutoPlaySelector {
         val preferredAddon = preferredAddonName?.trim().orEmpty()
         if (preferredAddon.isNotEmpty()) {
             val sameAddonCandidates = candidateStreams.filter { it.addonName == preferredAddon }
+            // X-axis: prefer the exact same stream name (quality/indexer) the user
+            // was watching. This mirrors Stremio behavior where the next episode
+            // picks the same stream position within the addon.
+            val preferredName = preferredStreamName?.trim().orEmpty()
+            if (preferredName.isNotEmpty() && sameAddonCandidates.isNotEmpty()) {
+                val sameNameMatch = sameAddonCandidates.firstOrNull { stream ->
+                    stream.name == preferredName && isPlayable(stream)
+                }
+                if (sameNameMatch != null) return sameNameMatch
+            }
             val fromPreferredAddon = selectFromCandidates(
                 candidateStreams = sameAddonCandidates,
                 mode = mode,
--- a/app/src/main/java/com/nuvio/tv/core/player/StreamAutoPlaySelector.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/StreamAutoPlaySelector.kt
@@ -55,7 +55,14 @@ object StreamAutoPlaySelector {
         selectedPlugins: Set<String>,
         preferredBingeGroup: String? = null,
         preferBingeGroupInSelection: Boolean = false,
-        bingeGroupOnly: Boolean = false
+        bingeGroupOnly: Boolean = false,
+        /**
+         * When set (typically the addon of the stream currently playing), prefer
+         * matching streams from that addon before falling back to the global
+         * ranking. Fixes next-episode jumping from e.g. "Torrentio TV" to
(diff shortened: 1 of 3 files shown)


skoruppa:
Where I find logic the logic from #2764 as something that can be added - for people who use duplicated addons where binge group could be duplicated, that changes from here are hard to understand. If user selected second stream, binge group should ensure it will follow with the same source. If such binge group does not exist I don't see any reason behind selecting the "second streams" just because user earlier also selected the "second stream"

haveAnIssue:
My argument is that it should also preserve the selected stream index within that source, at least when the stream ordering is stable. Remembering only the source may restore the same provider but not necessarily the same stream the user originally selected.

I tested Torrentio with Friends and the first stream index remains consistent from Season 1 Episode 1 through Episode 24 and even Season 2 Episode 1. If the first index can be reliably preserved, the same logic should apply to the second index as well.

https://torrentio.strem.fun/sort=seeders/manifest.json

skoruppa:
> My argument is that it should also preserve the selected stream index within that source, at least when the stream ordering is stable. Remembering only the source may restore the same provider but not necessarily the same stream the user originally selected.
> 
> I tested Torrentio with Friends and the first stream index remains consistent from Season 1 Episode 1 through Episode 24 and even Season 2 Episode 1. If the first index can be reliably preserved, the same logic should apply to the second index as well.
> 
> https://torrentio.strem.fun/sort=seeders/manifest.json

No, this doesn't make any sense. There is no way to ensure that source selected as n-th line will be also n-th for next episode. This is a task for the binge group

supergera13:
@skoruppa You make a fair point about binge groups. Let me clarify the specific case this PR addresses:

The X-axis preservation is meant for the scenario where:
1. The same addon returns multiple streams for the next episode (different qualities, releases, etc.)
2. The binge group matches the addon, but the specific stream index within that addon may differ between episodes

You're right that if binge groups are unique and well-defined, the source should match automatically. The case where this adds value is when binge groups are NOT unique (duplicate addons) or when the stream ordering changes between episodes — the user selected "stream 2" on episode 1 and expects "stream 2" on episode 2, not "stream 1".

@haveAnIssue tested this with Torrentio + Friends and confirmed the stream index is stable across episodes, which suggests the ordering IS consistent in practice. If that's the general case, then you're right — the binge group should handle it and this PR adds unnecessary complexity.

I'm happy to close this if you think the binge group coverage is sufficient. The base PR #2764 (Y-axis addon preference) is the important fix — this was the "nice to have" on top.

skoruppa:
Even with duplicated addons, binge groups should still address the same source...

We have a lot of available streaming addons and we can't make global decisions based on only one addon. I can ensure you that a lot of other addons DON'T have stable position

supergera13:
@skoruppa Understood. I'll close this PR then — the binge group coverage should be sufficient for the source-preference use case. Thanks for the detailed explanation.

supergera13:
Closing as discussed with @skoruppa — binge group coverage is sufficient for the source-preference use case, and global decisions based on stream index are not viable across all addons.

PR #2784, opened 2026-07-27
fix(player): restore ExoPlayer aspect modes broken by fixed-size surface buffers (#2761)

### LINKED ISSUE #2761: [Bug]: VIDEO ASPECT RATIO ISSUE
### Pre-flight checks

### App version / commit

0.7.20

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Xiaomi Mi Box S Gen 3

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

UI / Layout

### Bug description

When I change the video aspect ratio to CROP, the app stretches the video instead.

STRETCH also stretches videos.

This happens on Exoplayer, but works perfectly on libmpv, except that libmpv crops and resizes subtitles as I change aspect ratios.

### Steps to reproduce

Open Nuvio
Play a video
Change the aspect ratio of the video

### Expected behavior

CROP is supposed to fill the entire screen, not stretching the video as there is the option to do that as well. And in libmpv, changing aspect ratios is supposed to change only the video, and not with subtitles, because there is also an option to resize and adjust subtitle layout.

### Actual behavior

Crop aspect ratio stretches videos instead, and in libmpv, it resizes the video with subtitles.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

Springs-Tea:
## Summary

Fixes ExoPlayer aspect modes rendering as stretched video since 0.7.20. `SurfaceHolder.setFixedSize()` is now applied only in the default (`ORIGINAL`/Fit) aspect mode; every scaled mode (Crop, Stretch, zooms) switches the surface back to `setSizeFromLayout()`.

Root cause: 0.7.20 introduced `configureHardwareSurfaceOverlay()` (244379f3, PR #2719), which pins the video SurfaceView buffer to the video dimensions via `setFixedSize()` to engage the TV HWC/VPU scaler. When an aspect mode scales the surface past the screen edges (Crop scales `exo_content_frame` uniformly beyond the display bounds), the surface window gets clipped to the screen, and the composer then squeezes the entire fixed-size buffer into that clipped window. The result: Crop looks identical to Stretch (distorted picture). This matches the report exactly — broken on ExoPlayer, "Always", first reported the day 0.7.20 shipped, on an AMLogic-based box (Mi Box S), and 244379f3 is the only display/surface change between 0.7.19 and 0.7.20.

## PR type

- [x] Critical bug fix

## Why

Aspect ratio control is broken for ExoPlayer users on 0.7.20: Crop (and the zoom modes) distort the picture instead of cropping, with no workaround other than switching the whole engine to Libmpv.

## Issue or approval

Fixes #2761

## Reproduction steps

1. Set the internal player engine to ExoPlayer (0.7.20).
2. Play any title whose aspect ratio differs from the screen (e.g. a 2.35:1 movie on a 16:9 TV).
3. Cycle the aspect ratio to **Crop**.
4. Observe the picture is stretched/distorted (identical to Stretch) instead of cropped with correct proportions.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

- The HWC overlay optimization from #2719 is intentionally kept for the default `ORIGINAL` mode (the surface never overflows the screen there, so the fixed-size buffer is safe), and `setZOrderMediaOverlay()` / `Surface.setFrameRate()` are untouched in all modes.
- The Libmpv part of #2761 (subtitles scaling together with the video on aspect change) is a separate, pre-existing issue in the mpv surface scaling path and is not addressed here to keep this PR minimal.

## Testing

- Bisected the regression: `git log 0.7.19-beta..0.7.20-beta` shows 244379f3 is the only display/surface change between the last good and first bad release; the fix restores the exact pre-0.7.20 surface sizing (`setSizeFromLayout`) for all scaled modes.
- Verified every `configureHardwareSurfaceOverlay()` call site passes the current aspect mode, and that `applyExoAspectMode()` now re-syncs buffer sizing when the user cycles modes mid-playback (both directions: entering and leaving `ORIGINAL`).
- **Reproduced and verified on an Android TV emulator** (android-tv arm64 system image, Android 14 / API 34, 1920×1080), matching the reporter's Android 14 TV device:
  - Content: Tears of Steel 720p (1280×534, true 2.40:1 — open movie, CC-BY), served from a local Stremio addon; internal player on **ExoPlayer** (Media3 1.8.0 confirmed via logcat).
  - **Before (dev @ be335045, same code as 0.7.20):** Fit renders correctly letterboxed; switching to **Crop** fills the screen by vertically stretching the picture (identical to Stretch) instead of cropping — logcat confirms `Aspect mode toggled: ORIGINAL -> FULL_SCREEN (Crop)` while the frame is visibly distorted.
  - **After (this fix):** Crop fills the screen with a uniform zoom (sides cropped, proportions correct). Cycled through all 7 modes — Stretch still stretches (its intended behavior), Slight/Cinema Zoom, Vertical/Horizontal stretch, and Fit all behave as designed.
  - `assembleFullDebug` builds cleanly on both branches.

## Screenshots / Video

All from the same 2.40:1 test video on the internal ExoPlayer:

**1. Before — Fit: correct letterbox (baseline)**
1-before-fit-correct

**2. Before — Crop (OSD label visible): picture vertically stretched — the bug**
2-before-crop-BUG-stretched

**3. After — Crop: full-screen uniform crop, correct proportions**
3-after-crop-FIXED

**4. After — Stretch: still stretches (intended behavior preserved)**
4-after-stretch-intended

**5. After — Fit: still correct**
5-after-fit-correct

## Breaking changes

None.

## Linked issues

Fixes #2761

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerAspectScaleUtils.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerAspectScaleUtils.kt
@@ -131,6 +131,15 @@ internal fun applyExoAspectMode(playerView: PlayerView, mode: AspectMode) {
 
     applyAspectScale(targetView, mode, viewAspect, videoAspect)
     centerTargetInPlayer(playerView, targetView)
+    (surfaceView as? SurfaceView)?.let { surface ->
+        val videoSize = playerView.player?.videoSize
+        applySurfaceBufferSizing(
+            surfaceView = surface,
+            mode = mode,
+            videoWidth = videoSize?.width ?: 0,
+            videoHeight = videoSize?.height ?: 0
+        )
+    }
 }
 
 internal fun applyAspectMode(playerView: PlayerView, mode: AspectMode) {
@@ -225,14 +234,16 @@ internal fun configureHardwareSurfaceOverlay(
     playerView: PlayerView,
     videoWidth: Int,
     videoHeight: Int,
-    frameRate: Float = 0f
+    frameRate: Float = 0f,
+    aspectMode: AspectMode
 ) {
     val targetView = resolveVideoSurfaceView(playerView) as? SurfaceView ?: return
-    if (videoWidth > 0 && videoHeight > 0) {
-        runCatching {
-            targetView.holder.setFixedSize(videoWidth, videoHeight)
-        }
-    }
+    applySurfaceBufferSizing(
+        surfaceView = targetView,
+        mode = aspectMode,
+        videoWidth = videoWidth,
+        videoHeight = videoHeight
+    )
     runCatching {
         targetView.setZOrderMediaOverlay(false)
     }
@@ -247,6 +258,26 @@ internal fun configureHardwareSurfaceOverlay(
                     )
                 }
             }
+        }
+    }
+}
+
+private fun applySurfaceBufferSizing(
+    surfaceView: SurfaceView,
+    mode: AspectMode,
+    videoWidth: Int,
+    videoHeight: Int
+) {
+    // A fixed-size buffer is only safe while the surface stays inside the screen:
+    // once an aspect mode scales it past the screen edges, the composer squeezes the
+    // full buffer into the clipped window and distorts the picture (#2761).
+    if (mode == AspectMode.ORIGINAL && videoWidth > 0 && videoHeight > 0) {
+        runCatching {
+            surfaceView.holder.setFixedSize(videoWidth, videoHeight)
+        }
+    } else {
+        runCatching {
+            surfaceView.holder.setSizeFromLayout()
         }
     }
 }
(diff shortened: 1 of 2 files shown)


halibiram:
Thank you for the PR and the effort.
The root cause (#2719, configureHardwareSurfaceOverlay plus setFixedSize) was fully reverted in #2802  Because of that, the conditional `setFixedSize / setSizeFromLayout` logic based on aspect mode is no longer needed. Surface buffer behavior is back to the pre 0.7.20 state.
#2784 is therefore conflicting and no longer necessary. It can be closed. If the aspect ratio issue (#2761) is still present after the revert, it should just be re tested.
Thanks again.

Springs-Tea:
Thanks for the heads-up @halibiram! A full revert in #2802 makes sense — with `setFixedSize()` gone entirely, the aspect-mode-conditional buffer sizing in this PR is indeed redundant, and returning to the pre-0.7.20 surface behavior addresses the same root cause more thoroughly.

Closing this PR in favor of #2802. I have a reproducible emulator setup for #2761 (a true 2.40:1 test video with Crop/Stretch cycling on ExoPlayer), so I'm happy to re-test the aspect modes against a build that includes the revert and report back on the issue if that's useful.

Thanks again for taking the time to review!

PR #1279, opened 2026-04-08
Live AI Subtitle Translation via Gemini or Groq

silentbil:
## Summary

  AI Subtitle Translation

  Adds real-time AI-powered subtitle translation using Google Gemini directly inside the player — no external apps, no manual copy-paste.

  ✅ No more missing subtitles                                                                                                                                                                                                     
  ✅ No more subtitle timing issues

  How it works:
  - Enable AI Translation in  Settings → Integrations → AI Subtitles
  - Set your free Gemini API key (or scan a QR code from your phone)
  - Play any content with built-in subtitles — translation happens live as cues appear

  Features:
  - Auto-select mode: automatically activates AI translation on playback when enabled                                                                                                                                           
  - Strip hearing-impaired annotations from translated text
  - Subtle AI badge while translating, toast notification on first success or error
  - Key entry via on-screen keyboard or phone QR scan
  - Falls back to addon subtitles when no built-in tracks are available
  - Off by default — zero impact for users who don't opt in

**This couldn't be built as an addon — addons can fetch and serve subtitle files, but they have no access to the player's active cue stream at runtime. Translation requires intercepting subtitles frame-by-frame as they render,
   which is only possible from inside the player itself.**

## PR type
- Translations AI infra


## Why

Subtitles on streaming apps are almost always English-only. If your preferred language isn't available,
or the translation is poor — you're stuck. 
This feature lets you watch any content in your language, on the fly, 
powered by one of the best translation models available, without leaving the app.


## Policy check

<!-- Confirm these before requesting review -->
 - [x] This PR is not cosmetic-only, unless it is a translation PR.
- [x] This PR does not add a new major feature without prior approval.
- [x] This PR is small in scope and focused on one problem.
- [ ] If this is a larger or directional change, I linked the issue where it was approved.

<!-- PRs that do not match this policy will usually be closed without merge. -->

## Testing

  - Enabled AI Translation toggle and verified all sub-options appear/hide correctly                                                                                                                                                
  - Confirmed toggle off hides all AI options including hearing impaired filter                                                                                                                                                     
  - Played content with English built-in subtitles — translation appeared live in Hebrew                                                                                                                                            
  - Verified AI loading badge shows during translation and disappears after                                                                                                                                                         
  - Verified success/error toast on first batch result                                                                                                                                                                              
  - Tested QR code key entry flow via phone browser                                                                                                                                                                                 
  - Tested invalid/missing API key — error toast shown, fallback to original subtitle                                                                                                                                               
[...]

PR #2817, opened 2026-07-30
feat: AI subtitle auto-sync with jump fallback

theNuvioGuy:
### Problem
Addon-provided subtitle files are frequently out of sync with the video (different cut/release), and the existing manual delay slider requires trial and error. This feature automatically computes and applies the correct delay by matching **built-in (embedded) subtitle lines** — which are always correctly timed to the video — against the **addon subtitle file's lines**, using an LLM to do the semantic matching.

**This is the optimal approach for AI-assisted subtitle work: the original subtitle language/content is fully preserved (the AI is only used to find a handful of line-to-line matches for timing purposes), and no costly AI transcription/translation of the entire movie or episode audio is ever required — only a few short text snippets are sent to the LLM per sync, regardless of runtime.**

### High-Level Flow
1. When the user selects an addon subtitle (and Auto-Sync is enabled), the controller temporarily switches playback to a text-based **built-in** subtitle track (`beginTemporaryInternalCueGathering`) so ExoPlayer starts emitting `CueGroup`s for it.
2. As `CueGroup`s arrive, their text is buffered as timestamped "reference cues" (`recordSubtitleAutoSyncSourceCueGroup` → `appendSubtitleAutoSyncCue`), capped at the last 20 lines. A large jump in presentation time (>90s) clears the buffer, since that indicates a seek to a new region.
3. If the built-in track is image-based (PGS/DVB) with no text cues, the cue bitmap is OCR'd via ML Kit (`recognizeSubtitleBitmapText`) as a fallback.
4. Non-dialogue cues (`♪ music ♪`, `[Music]`, `(singing)`, etc.) are filtered out since they rarely have an equivalent line in the addon file.
5. Once ≥5 reference lines are buffered, the addon subtitle file is downloaded and parsed, and a window of ±12 lines around the estimated corresponding position is sent along with the reference lines to an LLM (Groq or Gemini, user-selectable, with user-supplied API keys).
6. The LLM is asked to return confident `{source_index, target_index}` pairs (semantic matching, not positional — line counts between built-in/addon subtitles aren't 1:1).
7. For each returned pair, an offset is computed as `builtInCueTime - addonCueTime` (adjusted for the delay already active on the built-in track's *rendered* position). The final delay is the **median** of all pooled offsets.
8. The process retries up to 3 attempts, pooling offsets across attempts until at least 3 data points are collected (or timing out), to reduce the chance of a single bad LLM match skewing the result.
9. Once a delay is computed, it's applied globally to the subtitle renderer and playback is switched back to the addon track (`endTemporaryInternalCueGathering`).

### Key Guards / Edge Cases Handled
- **Session/track identity locking**: a session key (`internalTrackKey|addonUrl`) prevents duplicate concurrent runs and remembers completed sessions so re-selecting the same pair doesn't re-trigger unless the user forces a re-run.
- **Stale cue protection**: switching tracks or addon subtitles clears the buffer so old cues never get matched against a new file.
- **MPV unsupported**: Auto Sync currently requires the ExoPlayer internal engine; MPV playback shows an explicit "unavailable" message.
- **No built-in track / non-text-only track**: surfaces a clear error instead of silently failing.
- **API key required**: if the selected provider (Groq/Gemini) has no key configured, the request fails fast with a message pointing to Settings.

### Remote Key Configuration (phone companion page)
`AddonConfigServer` was extended with a small web form (served over the existing local HTTP server) so the user can enter their Groq/Gemini API key and preferred model from their phone rather than typing it with a TV remote. Submitting the form POSTs to `/api/ai-keys`, which persists the values into `PlayerSettingsDataStore` (`subtitleAiGroqKey` / `subtitleAiGeminiKey`, `subtitleAiProvider`, `subtitleAiModel`).

### Settings Additions
New `PlaybackSettingsScreen` / `PlaybackSubtitleSettings` entries: AI provider picker (Groq/Gemini), per-provider API key entry dialogs, and toggles for `autoSyncExternalSubtitles` and `subtitleAiAutoSelect` (auto-trigger vs. manual "Apply").

### Testing
Fully tested on **Google TV on a TCL TV**, with two movies chosen specifically because synced Hebrew subtitles are notoriously hard to find for them: **The Shawshank Redemption** (חומות של תקווה) and **A Bronx Tale**. Testing was done with Hebrew as the built-in reference language. The only path **not** tested is the OCR/image-based subtitle fallback — no available source with an image-only (PGS/DVB) built-in track was found during testing. This fallback is an extreme edge case and is expected to be exercised rarely in practice.

> **Note:** The subtitle overlay currently contains extensive logging (`subtitleAutoSyncLastLlmRequest` / `subtitleAutoSyncLastLlmResponse` and related debug status text) to make it easy to follow exactly what the sync process is doing at each step. This was intentional for review/testing purposes — **it should be stripped down or gated behind a debug flag before this ships to production.**


theNuvioGuy:
@skoruppa @tapframe 

I think this is a strong addition to Nuvio; please include it ASAP, as maintaining changes is hard in such a big codebase.

Please don't let it die like https://github.com/NuvioMedia/NuvioTV/pull/1279
I think what was implemented is a very nice, much better solution (cheaper and more precise).

@silentbil will appreciate your feedback as this is inspired by your first work on this

haveAnIssue:
Beautiful.

hayekil:
I like that approach hopefully will be implemented in the coming versions

ofer45600:
יאללה משלנו! כל הכבוד אחשלי הלוואי והמפתחים ישלבו את זה בנוביו!

tapframe:
Thanks for the effort, I like the idea, but this PR is not ready yet.
It adds 2k+ lines and changes the player, UI, settings, and dependencies without linking an approved feature request. It also does not follow the PR template, has no screenshots, mixes in unrelated AccountViewModel changes and a large Spanish translation update, and adds no tests.[CONTRIBUTING.md]( https://github.com/NuvioMedia/NuvioTV/blob/dev/CONTRIBUTING.md)
The implementation also needs to be split up.` PlayerRuntimeControllerSubtitleTiming.kt` contains networking,ocr, and AI matching in the same file. I advise you to open up the PR after getting approval. So closing this one for now.

i know many users like this feature, but like we said, we are focusing on stability right now, and larger feature requests are currently being handled internally. we have implemented tons of features suggested by users over time, and if something has not been implemented yet, there are usually reasons for it. not every feature request or PR has to be merged just because the app is open source. maintaining stability and quality is our top priority. and there is much work to be done.

theNuvioGuy:
@tapframe   

Thanks for the detailed feedback — you're right, and I take it. Looking back at [CONTRIBUTING.md](https://github.com/NuvioMedia/NuvioTV/blob/dev/CONTRIBUTING.md), I skipped the process (no feature request issue, no approval, unrelated changes bundled in, no template, no formal tests, one giant file). That's on me, not a disagreement with the policy.

**Here's what I'll do to follow the process correctly (see note below as well):**

1. Open a **Feature Request issue** describing the AI subtitle auto-sync design, the approach (matching a few built-in subtitle lines against the addon file via LLM to compute a timing offset — no full-movie transcription/translation, so it stays cheap and preserves the original subtitle content), and the testing already done (Google TV on TCL, two Hebrew-subtitle-scarce titles: *The Shawshank Redemption* and *A Bronx Tale*; OCR fallback path not yet exercised).
2. **Wait for explicit approval** before writing/reopening any PR — I won't resubmit until you've said it's a go.
3. If approved, split the work into **small, focused, reviewable PRs** instead of one large diff:
   - Core sync logic (cue capture, track switching, offset calculation) separated from AI provider networking and from the OCR fallback, each in its own file/module
   - Settings UI as its own PR
   - Remote/phone key-entry server as its own PR
   - Drop the unrelated `AccountViewModel` auth-guard change and the Spanish translation update entirely from this work — I'll submit those separately if they're still needed (see below note on this)
4. Each PR will follow the template: linked approved issue, screenshots/short demo video, before/after behavior notes, and tests where they're feasible.

**One honest thing I want to raise before I invest more time into this:** a feature like this (player integration, multiple AI providers, OCR fallback, settings, key management) is a meaningful amount of ongoing maintenance. I'm happy to put in that effort and follow the process exactly as you've laid out, *but I'd like some genuine commitment on your side too*: that if I do this properly (approved issue, small PRs, tests, docs), it will actually get a real review pass rather than being closed again for reasons unrelated to quality. Closing a PR is easy; working together to get something like this into a mergeable state is the approach that actually benefits the project and the users who've been asking for it. If that's something you're up for, I'll work on this.

**Note on the unrelated changes you flagged:** you're right that the `AccountViewModel` auth-guard and the Spanish translation update don't belong in this PR — they were leftover local work that got swept into the same commit by mistake, not something intentionally bundled into the subtitle feature. For context on why they exist at all: `local.properties` (which backs the `release` build type) doesn't ship with real Supabase credentials — only `local.dev.properties` does — so building a `release`-type variant from a clean checkout without your own Supabase keys hits sign-in/QR-login with a blank `SUPABASE_URL`/`SUPABASE_ANON_KEY`, which previously wasn't guarded and would likely surface as a confusing crash or malformed-request error instead of a clear message. The guard just makes that fail gracefully with "Sign-in is not configured in this build." I'll pull it out into its own PR rather than leave it here.

Final thing: "thinking big" should not come at the price of stability. Thank you for your work :)

tapframe:
> One honest thing I want to raise before I invest more time into this: a feature like this (player integration, multiple AI providers, OCR fallback, settings, key management) is a meaningful amount of ongoing maintenance. I'm happy to put in that effort and follow the process exactly as you've laid out, but I'd like some genuine commitment on your side too: that if I do this properly (approved issue, small PRs, tests, docs), it will actually get a real review pass rather than being closed again for reasons unrelated to quality. Closing a PR is easy; working together to get something like this into a mergeable state is the approach that actually benefits the project and the users who've been asking for it. If that's something you're up for, I'll work on this.

this is the reason there are rules, so your time and effort won't go wasted.  for this particular case, this is not something in our priority list right now. and as I said, at this moment, larger changes are discussed and done internally only. hope that clarifies.

theNuvioGuy:
@tapframe

I was not complaining about my efforts; it was a fun project, and since I was able to run it perfectly on my side, I can even use it myself. I was asking about future effort; whether they would be wasted or not :)

However, I understand that this significant change is not welcome, even though you recognize the strong demand for it.

Thanks again

Akila3:
this is a must have in Nuvio.
those are directly related to the FR that was opened:
https://github.com/NuvioMedia/NuvioMobile/issues/1624

DantexDT:
Really necessary and useful. It would be a very nice addition to NUVIO wich is already great!

newspeer:
This would be a great addition!!!

PR #2639, opened 2026-07-15
Fix external player auto next handoffs, binge group continuity, and finale detection

Laskco:
## Summary

This fixes a regression in external player auto next completion. Nuvio now captures the next episode decision from the episode metadata already loaded by the player or Stream screen, then refreshes it in the background without discarding a known playable successor.

A confirmed finale, including episode 12 of 12 and any other final available episode, returns without presenting an auto next loader or attempting another launch. Unavailable and unaired episodes are excluded. When the episode list is genuinely unavailable, the existing metadata lookup remains as a fallback.

The transition loader now remains under one owner from the moment Nuvio returns to the foreground until the next external player covers the app. Its message and progress follow the active Stream loading stage, including source discovery, subtitle loading, and intro or outro lookup. Stream settlement cannot briefly expose the episode or source screen during an active handoff.

The binge group still belongs only to the stream that actually launched. If that group is unavailable for the next episode, normal autoplay selection remains the fallback. Back cancellation is checked before tracker setup, after settings are read, during launch preparation, and immediately before the external player intent, so cancelling cannot launch the pending continuation later.

## PR type

- [ ] Translation/localization only
- [x] Critical bug fix

## Why

Several lifecycle races affected the same external player completion path. The background metadata prefetch could fail or return a thinner episode list even though the current screen had already loaded a valid successor. Stream settlement callbacks could release the transition loader before navigation and source resolution finished. A cancelled continuation could also finish asynchronous launch preparation and open the external player after the user had backed out.

Together these races could return the user to the previous episode source selection, reuse an abandoned binge group, show a loader after a final episode, expose the episode or source screen between players, lose detailed loading status, or launch a continuation after cancellation.

The completed stream now owns the handoff state. A loaded episode snapshot supplies the strongest available successor or finale decision, the root transition loader keeps ownership until external playback covers Nuvio, and cancellation remains authoritative through the final launch boundary.

## Issue or approval

Fixes #2638

## Reproduction steps

1. Enable auto play next episode and select an external player.
2. Launch an episode from one binge group, back out, then launch another group and finish the episode naturally.
3. Confirm the next episode uses the group that actually launched or falls back to normal autoplay selection when that group is unavailable.
4. Observe the transition through episode navigation, source discovery, subtitle loading, and intro or outro lookup.
5. Repeat with episode 12 of 12 or another final available episode.
6. During a pending next episode transition, press Back and confirm no external player opens afterward.

[...]

PR #2830, opened 2026-07-30
fix(player): keep external-player handoff alive after leaving PlayerScreen (#2560)

### LINKED ISSUE #2560: [Bug]: App exits to previous screen when attempting to switch to an external player
### Pre-flight checks

### App version / commit

0.7.15 Beta

### Install method

GitHub Release APK

### Platform

Fire TV

### Device model

Amazon Fire TV Cube (3rd Gen) - Connected to Hisense 65U8NQ TV and Samsung HW-Q800C Soundbar

### Android version

Fire OS 7

### Player mode

Internal player

### Area (tag)

### Bug description

When the default player in Nuvio settings is set to "Internal", starting a stream works perfectly. However, if I try to switch to an external player from within the internal player's UI during active playback, the app abruptly stops the video and returns to the previous screen.

Important note: If I set the default player to "External" in Nuvio settings beforehand, the app correctly prompts for the external player when opening a video and it works without any issues. The bug specifically happens during the handoff from the internal player to an external one.

I was not on the home for last weeks so I'm not quite sure but I think this feature was working great on [Beta 0.7.11] or [Beta 0.7.10].

### Steps to reproduce

1. Go to Nuvio settings and ensure the default player is set to "Internal".
2. Open any movie or series and start playing it.
3. While the video is playing, use the on-screen UI to select the option to switch to an External Player.
4. Observe that instead of opening the external player, the app immediately exits playback and drops back to the previous page.

### Expected behavior

The internal player should seamlessly close and hand over the stream to the selected external player without dropping back to the menu.

### Actual behavior

The application exits the playback entirely and returns to the previous screen.

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

I have taken a recording of this issue occurring. 
https://drive.google.com/file/d/1JpolUFp7zu2bJYR0E3_peT4B6vJuMQCb/view?usp=sharing

```shell

```

### Anything else? (optional)

Image

supergera13:
## Summary

Fixes switching from the internal player to an external player mid-playback. The external intent is no longer cancelled when PlayerScreen is left immediately after "Open in External Player".

## PR type

- [x] Critical bug fix

## Why

`onOpenInExternalPlayer` stopped the internal player, launched via `viewModelScope`, then called `onBackPress` right away. Navigating away cleared the ViewModel and cancelled the coroutine before `ExternalPlaybackTracker.launchPlayer` could fire the external intent — so the user only saw playback exit back to the previous screen.

Default "External" player from Stream still works because that path launches before leaving a different screen/lifecycle.

## Issue or approval

Fixes #2560

## Reproduction steps

1. Settings → default player = **Internal**
2. Start any stream in the internal player
3. Open the player UI and choose **Open in External Player**

Before: playback stops and the app returns to the previous screen; external player never opens.  
After: internal player stops and the selected external player opens on the same stream (with resume position).

## UI / behavior impact

- [x] No UI redesign — restores intended external handoff behavior
- Progress saving remains owned by `ExternalPlaybackTracker` on return

## Policy check

- [x] Critical bug fix with linked GitHub issue
- [x] Minimal and focused on the handoff path
- [x] No new dependencies
- [x] Stream-screen external launch path unchanged

## Scope boundaries

- Internal → external handoff from PlayerScreen / PlayerViewModel / ExternalPlaybackTracker only

## Testing

- Capture metadata + URL before `stopAndRelease`
- Enqueue launch on tracker process-scoped scope (`launchPlayerInBackground`) so ViewModel clear cannot cancel it
- Subtitle prep still runs before intent when configured

## Screenshots / Video

Not a layout change. Behavior: internal player → external player handoff succeeds instead of popping back.

## Breaking changes

None.

## Linked issues

Fixes #2560

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
@@ -330,39 +330,6 @@ class ExternalPlaybackTracker @Inject constructor(
      * Optional [prepareSubtitles] runs on this same scope before the intent is fired
      * (subtitle downloads must not be cancelled by ViewModel clear).
      */
-    fun launchPlayerInBackground(
-        metadata: ExternalPlaybackMetadata,
-        url: String,
-        title: String?,
-        headers: Map<String, String>?,
-        resumePositionMs: Long = 0L,
-        subtitles: List<SubtitleInput>? = null,
-        autoLaunch: Boolean = false,
-        nextEpisodeSnapshot: ExternalNextEpisodeSnapshot? = null,
-        context: Context,
-        prepareSubtitles: (suspend () -> List<SubtitleInput>?)? = null
-    ) {
-        scope.launch {
-            val resolvedSubtitles = try {
-                prepareSubtitles?.invoke() ?: subtitles
-            } catch (e: Exception) {
-                Log.w(TAG, "Subtitle prep for external player failed; launching without", e)
-                subtitles
-            }
-            launchPlayer(
-                metadata = metadata,
-                url = url,
-                title = title,
-                headers = headers,
-                resumePositionMs = resumePositionMs,
-                subtitles = resolvedSubtitles,
-                autoLaunch = autoLaunch,
-                nextEpisodeSnapshot = nextEpisodeSnapshot,
-                context = context
-            )
-        }
-    }
-
     /**
      * Launch external player with progress tracking.
      * Uses the Activity-level launcher for ActivityResult, or fire-and-forget on Zidoo.
--- a/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
@@ -323,6 +323,47 @@ class ExternalPlaybackTracker @Inject constructor(
     }
 
     /**
+     * Enqueues [launchPlayer] on the tracker's process-scoped [scope] so the launch
+     * survives ViewModel / composition clear (e.g. leaving PlayerScreen immediately
+     * after "Open in External Player" — see #2560).
+     *
+     * Optional [prepareSubtitles] runs on this same scope before the intent is fired
+     * (subtitle downloads must not be cancelled by ViewModel clear).
+     */
+    fun launchPlayerInBackground(
+        metadata: ExternalPlaybackMetadata,
+        url: String,
+        title: String?,
+        headers: Map<String, String>?,
+        resumePositionMs: Long = 0L,
+        subtitles: List<SubtitleInput>? = null,
+        autoLaunch: Boolean = false,
+        nextEpisodeSnapshot: ExternalNextEpisodeSnapshot? = null,
+        context: Context,
+        prepareSubtitles: (suspend () -> List<SubtitleInput>?)? = null
+    ) {
+        scope.launch {
+            val resolvedSubtitles = try {
+                prepareSubtitles?.invoke() ?: subtitles
+            } catch (e: Exception) {
+                Log.w(TAG, "Subtitle prep for external player failed; launching without", e)
+                subtitles
+            }
+            launchPlayer(
+                metadata = metadata,
+                url = url,
+                title = title,
+                headers = headers,
+                resumePositionMs = resumePositionMs,
+                subtitles = resolvedSubtitles,
+                autoLaunch = autoLaunch,
+                nextEpisodeSnapshot = nextEpisodeSnapshot,
+                context = context
+            )
+        }
+    }
+
+    /**
      * Launch external player with progress tracking.
      * Uses the Activity-level launcher for ActivityResult, or fire-and-forget on Zidoo.
      * If resumePositionMs is 0, fetches the saved position from the repository.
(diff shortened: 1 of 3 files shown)


Laskco:
External Player guy here. One thing to say firstly, be extremely careful with anything to do with the external player support code. It's fragile and there's a shitload of issues that come if you're not cautious with it. 

Things that I will need you to test and report back,
1. **Internal to external handoff:** Start with the internal player, choose **Open in External Player**, and confirm the external player opens without showing the details or source screen in between.

2. **Auto next transition:** Finish an episode and confirm the next episode launches while the loading screen remains visible and stable throughout the transition.

3. **Binge group continuity:** Back out of one source group, launch from another, and confirm auto next uses the group that actually launched. If it is unavailable, normal auto play selection should take over.

4. **Final episode handling:** Finish the last available episode, such as episode 12 of 12, and confirm no auto next loader appears and Nuvio returns normally.

5. **Cancellation and stale launches:** Cancel an auto next transition or start another stream while a handoff is preparing. Confirm the cancelled or older stream never launches afterward.



Anyways,
The `viewModelScope` cancellation looks like the correct root cause, and capturing the stream information before stopping the internal player makes sense. I’m just concerned that moving the whole launch into a process scoped background job could cause some of the older external player issues to return.

`launchPlayerInBackground()` has no way to cancel or replace an older launch. Since `PlayerScreen` closes immediately, the first launch could still be preparing subtitles, resume data, or skip segments when the user starts another stream. It could then open over the newer stream or replace its tracking information. This is similar to the stale handoff and wrong episode problems fixed in #2639.

`onBackPress` also runs before the external player has actually opened. If subtitle or intro and outro preparation takes a few seconds, the previous screen may appear before the external player launches. That is the same type of transition issue addressed in #2463.

There is also no launch result returned to `PlayerScreen`. If the URL or content ID is missing, or the external player fails to open, the internal player still closes and the user is sent back.

I think the safer approach would be:

1. Keep `PlayerScreen` open while the handoff is being prepared.
2. Wait for `ExternalPlaybackTracker.launchPlayer()` to return its result.
3. Only call `onBackPress` after the external player intent has been sent successfully.
4. Give subtitle preparation an overall timeout and continue without forwarded subtitles if it takes too long.
5. Block repeated clicks while the handoff is running.

If the process scoped launch needs to stay, it should keep track of the active handoff so a newer launch can cancel the older one and prevent stale cleanup from affecting the current stream.

supergera13:
@Laskco Thanks for the detailed warning and test list. I updated this PR so the PlayerScreen stays alive while subtitle preparation and the external handoff are in progress, and it only navigates back after ExternalPlaybackTracker.launchPlayer() reports a successful launch. Repeated handoff clicks are ignored while one is pending, subtitle preparation is bounded to 10 seconds, and missing/failed launches no longer navigate away silently. The manual path now waits in the ViewModel scope instead of using a fire-and-forget process-scoped launch.

I cannot run the real-device checks from this environment, so the five scenarios you listed still need maintainer/device verification, especially cancellation/stale-launch behavior and external auto-next continuity.

supergera13:
@Laskco thanks for the careful review — agreed the process-scoped path was the wrong direction.

The PR now follows the safer handoff you described:

1. `PlayerScreen` stays open while the handoff runs (`externalHandoffInProgress`)
2. `launchInExternalPlayer` waits for `ExternalPlaybackTracker.launchPlayer()` (with a 10s subtitle prep timeout) and reports success/failure via callback
3. `onBackPress` only runs after a successful launch
4. Repeated open-in-external clicks are blocked while the handoff is in progress
5. Removed the unused `launchPlayerInBackground` helper so stale process-scoped launches cannot return

I still cannot fully exercise all 5 external-player device scenarios on my side (no external player installed on the emulator). If you have a chance to re-check the handoff + auto-next cancellation paths on a real box, that would be the definitive sign-off.

Laskco:
> @Laskco thanks for the careful review — agreed the process-scoped path was the wrong direction.
> 
> The PR now follows the safer handoff you described:
> 
> 1. `PlayerScreen` stays open while the handoff runs (`externalHandoffInProgress`)
> 2. `launchInExternalPlayer` waits for `ExternalPlaybackTracker.launchPlayer()` (with a 10s subtitle prep timeout) and reports success/failure via callback
> 3. `onBackPress` only runs after a successful launch
> 4. Repeated open-in-external clicks are blocked while the handoff is in progress
> 5. Removed the unused `launchPlayerInBackground` helper so stale process-scoped launches cannot return
> 
> I still cannot fully exercise all 5 external-player device scenarios on my side (no external player installed on the emulator). If you have a chance to re-check the handoff + auto-next cancellation paths on a real box, that would be the definitive sign-off.

I'll make an APK later today and I’ll test it and report back here.

Laskco:
So, I'm unable to reproduce the issue due to not owning a FireOS device. I believe this is exclusive to that hardware. 
Will need the original bug poster to test if this fixes their issue. @ibrakadabracoding 

Please test this debug build linked here in the gofile. 
https://gofile.io/d/Rf5NlL

ibrakadabracoding:
Hi @Laskco and @supergera13, thank you for your attention and solutions. I did my tests with today's release 0.8.1 and also with your debug version. Here is my findings on my Fire Cube device:

The problem still exists on 0.8.1. I still cannot switch to external player when I started with internal.

The good news is, your debug version is **working great**. Now I can open a movie/series with internal player and then switch to external without any issue. I checked with one movie and one series episode. One of them from Real Debrid and the other one from Web sources. Both are working great. I think I can confirm that your solution fixes the bug.

supergera13:
Very good to hear man 👍🏻

PR #2854, opened 2026-08-01
fix(continue-watching): restore episode videoId for legacy progress (#2841)

### LINKED ISSUE #2841: [Bug]: Wrong episode played on resume
### Pre-flight checks

### App version / commit

0.7.20

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Google TV Streamer

### Android version

14

### Player mode

Internal player

### Area (tag)

### Bug description

I went to resume using Continue Watching an episode we were halfway through. Upon playing we realised it was playing the first episode available, not the fourth/fifth episode. It said in the UI it was the episode we expected.

### Steps to reproduce

1. watch series until mid-way through
2. pause halfway through
3. resume using Continue Watching

### Expected behavior

Resumes the episode you were on

### Actual behavior

It resumes at the point you expect, but on the wrong episode.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

I Will Find You S1E4 or 5.

```shell

```

### Anything else? (optional)

_No response_

supergera13:
## Summary

Fix Continue Watching resuming the wrong episode when legacy local watch-progress entries lack a proper episode-specific `videoId` (or store the series `contentId`).

## PR type

- [x] Critical bug fix

## Why

Continue Watching navigates to Stream using `progress.videoId`. Older local snapshots either omitted `videoId` or stored the series `contentId`. Stream/addons then resolved episode 1, while the UI still showed the correct season/episode labels and resume position.

In 0.8, remote progress (Trakt/Simkl) already prefers the **meta-addon** episode video id during Continue Watching enrichment. Local progress with a missing/series-level videoId needs the same path. Inventing `contentId:season:episode` is wrong for anime trackers (`mal:` / `kitsu:`) that have no season segment and can disagree with what the meta catalog exposes.

## Issue or approval

Fixes #2841

## Reproduction steps

1. Have series progress mid-way through an episode (e.g. S1E4/E5) from an older app version / legacy progress snapshot where `videoId` is missing or equals the series id.
2. From Home, open that item via Continue Watching (after enrichment / with a previously enriched cache).
3. Before: streams/playback load episode 1 (or the first available episode) instead of the episode shown in the UI.
4. After: stream route uses the meta-addon episode video id for that S/E.

## UI / behavior impact

- [x] No UI change

## Policy check

## Scope boundaries

- Extends the existing Continue Watching enrichment preference for meta episode video ids (remote always; local only when videoId is blank or equals series `contentId`).
- Reuses the same rule for in-memory/disk enrichment overlays so a previously resolved meta id is applied before live enrichment finishes.
- Does **not** artificially build `contentId:season:episode`.
- Does not change stream selection logic beyond using the corrected videoId on the CW route.

## Testing

- `./gradlew :app:testFullDebugUnitTest --tests com.nuvio.tv.ui.screens.home.ContinueWatchingVideoIdTest :app:compileFullDebugKotlin` — **passed**
  - Remote progress prefers meta video id
  - Local series-level / blank videoId prefers meta (including `mal:` / `kitsu:` absolute forms)
  - Local stored episode videoId is retained
  - Fallback when meta video id is unavailable
- Manual device check still recommended: resume a series episode via Continue Watching with legacy local progress (missing/series-level videoId) and confirm streams match the shown episode after CW enrichment.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #2841


skoruppa:
I'm not sure about this fix. And I'm not even sure if it is needed. First, I think it will not work using ids that contain no season like mal or kitsu. Second, videoid should be taken from a meta addon and not be artificially build, of course only if videoid is missing in the progress entry.

With 0.8 I remember I added taking videoid from meta, but only for remote progress like simk/trakt. But it seems it should be also used when videoid is missing (I had no idea such thing can happen)

supergera13:
@skoruppa Thanks — agreed on both points.

The previous approach of inventing `contentId:season:episode` was wrong: MAL/Kitsu (and similar) have no season segment, and the real episode id should come from the meta addon.

Reworked this PR to follow the 0.8 path you added for remote progress:

1. **Removed** artificial videoId rebuild on JSON parse
2. **Extended** meta videoId preference in CW enrichment:
   - remote (Trakt/Simkl): still always prefer meta when available
   - local: prefer meta **only** when `videoId` is blank or equals the series `contentId` (legacy missing episode id)
3. Apply the same rule when overlaying a previously enriched/cached videoId so a cold resume is less dependent on inventing ids

`resolveVideoForProgress` already finds the episode via S/E when the stored videoId does not match a meta entry; we then keep that meta `video.id`.

Verified:
```
./gradlew :app:testFullDebugUnitTest --tests com.nuvio.tv.ui.screens.home.ContinueWatchingVideoIdTest :app:compileFullDebugKotlin
```
passed (includes mal/kitsu-style meta ids without season segments).

Honest limit: full end-to-end resume still needs a device/emulator with a meta addon + legacy local progress snapshot; the pure preference rules and compile are covered here.

app/src/main/java/com/nuvio/tv/ui/screens/home/ContinueWatchingVideoId.kt
```diff
@@ -0,0 +1,34 @@
+package com.nuvio.tv.ui.screens.home
```
skoruppa: I'm not sure if I like creating a whole new file, for a little function that has bigger description that whole logic instead of adding one additional if...

supergera13: Refactored per your point: dropped the standalone `ContinueWatchingVideoId.kt` and inlined `resolveContinueWatchingVideoId` directly in `HomeViewModelContinueWatching.kt` (same file as all three call sites), trimmed the KDoc down to the actual logic. The helper stays `internal` so the existing unit tests still cover it without needing a dedicated source file.

supergera13:
@skoruppa Status after your feedback (no new code since the last push):

1. Dropped artificial \`contentId:season:episode\` rebuild (would break mal/kitsu).
2. Prefer meta-addon episode \`videoId\` for local progress only when blank or equal to series \`contentId\` (same idea as the 0.8 remote path).
3. Inlined the helper into \`HomeViewModelContinueWatching.kt\` — no dedicated source file.

Ready for another look when you have time. Still not claiming full device/e2e verification of every legacy snapshot shape.

skoruppa:
I was playing a bit with the CW Progress items, and I don't think your changes make any difference. In Progress items without video id should not exist. Only an external tool could have created it. And after assuring the video id is cached and taken from meta, I don't think there is anything else here to fix

supergera13:
Closing based on your findings — agreed that progress entries without a videoId shouldn't exist in-app, and that caching the meta addon videoId covers the real path. Thanks for digging into CW progress and for the earlier guidance on not inventing ids for mal/kitsu.

PR #2860, opened 2026-08-01
fix(player): bind exact addon subtitle track on manual switch (#2601)

### LINKED ISSUE #2601: Isn't selecting the subtitle I've selected
### Pre-flight checks

### App version / commit

0.7.16-beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Google TV (SW-22AE NF)

### Android version

Android TV OS version 11

### Player mode

Internal player

### Area (tag)

### Bug description

https://github.com/user-attachments/assets/7dab0672-81c7-416f-b542-fb04048e9601

By default, it selects a subtitle from OpenSubtitles v3, which is fine. But when I manually select a subtitle from "SubMaker | ElfHosted" instead, it doesn't actually switch. The UI shows it as selected, but when I play the video, it still plays the previous OpenSubtitles v3 subtitle.

### Steps to reproduce

1. Install both of these add-ons:-

    • OpenSubtitles v3
    • SubMaker | ElfHosted
2. Play an anime 
3. Switch the subtitle to "SubMaker | ElfHosted"

### Expected behavior

It should switch to whichever subtitle I chose

### Actual behavior

Isn't switching to whichever subtitle I chose (initially it shows that it has selected a subtitle from "SubMaker | ElfHosted" but doesn't play the video using it's subtitle)

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

Playing Hell's paradise (S2E1) where the internal engine is set to "Auto (Best for content)" and the addon subtitle loading is set to preferred languages

```shell

```

### Anything else? (optional)

_No response_

supergera13:
## Summary

When the user manually switches from one preferred-language addon subtitle (e.g. OpenSubtitles v3) to another in the same language (e.g. SubMaker | ElfHosted), ExoPlayer now binds the **exact** selected sidecar track instead of re-selecting the first same-language match. The subtitle overlay already updated `selectedAddonSubtitle`, so the UI looked correct while playback kept the previous track.

## PR type

- [x] Critical bug fix

## Why

With addon subtitle startup on preferred languages, every matching-language sidecar is attached up front. Manual switch used a language-only `TrackSelectionOverride` fallback when applying a pre-attached track, which always hit the first same-language track (often the auto-selected OpenSubtitles entry). Users saw SubMaker selected in the UI while OpenSubtitles cues kept playing.

## Issue or approval

Fixes #2601

## Reproduction steps

1. Install both OpenSubtitles v3 and SubMaker | ElfHosted (or any two subtitle addons for the same preferred language).
2. Set addon subtitle loading to preferred languages (as in the report).
3. Play content where both addons return subtitles (e.g. anime S2E1 as reported).
4. Confirm auto-select picks OpenSubtitles (or any first match).
5. Open the subtitle overlay and select the SubMaker (or second-addon) entry for the same language.
6. Observe cues still come from the first addon while the UI shows the second selected.

After this fix, step 6 applies the selected track id override (or reloads media and binds that pending id); cues match the chosen addon.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

- ExoPlayer addon-sidecar selection only (`selectAddonSubtitle`, track-id / language override helpers, pending track application after media refresh).
- Does not change MPV external-subtitle path, auto-select preference order, subtitle UI layout, or startup attachment policy.
- Language-only override still works when exactly one same-language addon track exists, or when a preferred track id is supplied (timing refresh / pending bind).

## Testing

- Static review of ExoPlayer switch path against #2601: preferred-language multi-addon attach → language-first fallback → UI/player desync.
- Confirmed no-reload path no longer calls language-only apply; reload path keeps `pendingAddonSubtitleTrackId` and binds by id (with preferred-id language helper as secondary).
- Confirmed timing-refresh restore passes `preferredTrackId` so delay bounce cannot re-pick the wrong same-lang sidecar.
- No device/emulator run in this environment; relies on CI `fullDebug` build and logic review.

## Screenshots / Video

Not a UI change (selection highlight already updated; this restores player track binding to match).

## Breaking changes

None.

## Linked issues

Fixes #2601

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerTrackSelection.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerTrackSelection.kt
@@ -5,6 +5,7 @@ import androidx.media3.common.C
 import androidx.media3.common.MediaItem
 import androidx.media3.common.Player
 import androidx.media3.common.TrackSelectionOverride
+import androidx.media3.common.Tracks
 import com.nuvio.tv.domain.model.Subtitle
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.update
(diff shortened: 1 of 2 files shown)


skoruppa:
I again think the issue was misdiagnosed and never reproduced. I never had an issue when switching from one addon to another when picking subtitles, and proposed solution is made like it is broken for everyone. Reporter in the bug's description mentioned that the issue happens when the engine is set to auto and when watching anime - it means the engine playing video was mpv, and probably somewhere there we can find the issue, not in exo

supergera13:
@skoruppa You're right — closing this.

I never reproduced the switch failure myself. The reporter's notes (anime + engine **Auto (Best for content)**) point at **mpv**, not Exo. This PR only hardens the Exo addon-subtitle override path and treats a general Exo first-match bug that I did not actually observe end-to-end.

I'll leave #2601 open for a proper mpv-side investigation (or a confirmed Exo-only repro with engine forced to ExoPlayer). Sorry for the noise.

supergera13:
Closing as misdiagnosed: reporter conditions imply mpv under Auto engine, not an Exo first-match bug. Will revisit only with a confirmed engine-specific repro.

PR #2880, opened 2026-08-01
fix(subtitles) hide "�" from subtitles

haveAnIssue:
There are subtitles that display the character � (uFFFD). I have simply removed it from the subtitle display.

## Summary

Added a stripReplacementCharacter() step to CueNormalizingTextOutput.processCue() in PlayerRuntimeControllerInitialization.kt, the single pipeline that every subtitle cue passes through before being shown on screen during playback.

## PR type

- [x] Critical bug fix

## Why

Some subtitle files contain bytes that don’t decode cleanly with the character encoding being used (e.g. a subtitle saved in a legacy encoding like Windows-1255/1256 being read as UTF-8, or a genuinely corrupted byte). When that happens, the decoder can’t map that byte sequence to a real character, so it substitutes the “�” placeholder instead of silently dropping it. That placeholder was passing straight through to the on-screen subtitle renderer, showing up as a distracting stray character in otherwise normal dialogue lines.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerRuntimeControllerInitialization.kt
@@ -2176,6 +2176,10 @@ private fun FfmpegAudioRenderer.applyDownmixSettings(
     }
 }
 
+// The Unicode replacement character ("�") that shows up when a subtitle byte sequence fails to
+// decode with the detected/assumed charset.
+private const val REPLACEMENT_CHARACTER = '\uFFFD'
+
 private class CueNormalizingTextOutput(
     private val delegate: TextOutput,
     private val shouldNormalizeCuePositionProvider: () -> Boolean,
@@ -2194,7 +2198,8 @@ private class CueNormalizingTextOutput(
     }
 
     private fun processCue(cue: Cue): Cue {
-        var processed = fixRtlCueText(cue)
+        var processed = stripReplacementCharacter(cue)
+        processed = fixRtlCueText(processed)
         if (shouldNormalizeCuePositionProvider()) {
             processed = normalizeCuePosition(processed)
         }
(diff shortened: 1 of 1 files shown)


haveAnIssue:
[445BACF134291168CC05CDB25BA5FC43.txt](https://github.com/user-attachments/files/30630577/445BACF134291168CC05CDB25BA5FC43.txt)

example. Read row 954 to understand.

haveAnIssue:
when a subtitle file’s byte sequence fails to decode with the detected/assumed charset (e.g. a legacy-encoded SRT read as UTF-8), the decoder substitutes the Unicode replacement character (“�”) in place of the original character, and that placeholder is currently passed straight through to the viewer during playback. There are a few ways this could theoretically be addressed — attempting to re-detect the correct charset and re-decode, trying to guess/recover the original character from context, and so on — but those are heuristic, unreliable, and add real complexity and risk of introducing new bugs for something the source data doesn’t give us enough information to solve properly. The correct and simplest approach is simply to strip the character before it reaches the viewer, which is what this PR does. Would appreciate a review when possible. @skoruppa

haveAnIssue:
@halibiram I don't understand why you deleted this code? It's important for Hebrew speakers.

halibiram:
> @halibiram I don't understand why you deleted this code? It's important for Hebrew speakers.

My bad, since I didn't merge the branch, I might have accidentally wiped out your fix. Could you open a PR for it again?

haveAnIssue:
@halibiram Actually, I don't know how to proceed after your PR.
The code is in PlayerRuntimeControllerInitialization.kt but it has no effect at all on the Subtitle/Preview.

**_Line 2185_**: private const val REPLACEMENT_CHARACTER = '\uFFFD'

and _**Line 2307**_: private fun stripReplacementCharacter(cue: Cue): Cue

PR #2900, opened 2026-08-03
feat: Continue Watching card style and info options, with corner radius and artwork fixes

### LINKED ISSUE #2911: [Bug]: Continue Watching ignores the poster corner radius setting
### Pre-flight checks

### App version / commit

0.8.1

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Xiaomi Mi TV

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

UI / Layout

### Bug description

The poster corner radius setting under Settings > Layout applies to every catalog row, but not to the Continue Watching row. With the radius set to sharp, catalog posters render square while Continue Watching cards stay rounded, so the home screen looks inconsistent.

The rounding also shows on the card's focus ring and on the artwork clip, not just the card background. No error text appears; it is purely a visual inconsistency.

### Steps to reproduce

1. Open Settings > Layout.
2. Set the poster corner radius to sharp.
3. Return to the home screen.
4. Compare a catalog row against the Continue Watching row.

### Expected behavior

Continue Watching cards use the same corner radius as catalog posters, so the whole home screen follows the setting.

### Actual behavior

Continue Watching cards use the same corner radius as catalog posters, so the whole home screen follows the setting.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

ram130:
## Summary
Continue Watching cards can now follow the same card styles NuvioMobile offers, instead of being locked to the landscape card.
One setting under Settings - Layout - Continue Watching:
- **Card Style**: Card (landscape, current default), Wide, or Poster (portrait)
The default is unchanged, so nothing moves until a user opts in. Poster and Wide are ported from NuvioMobile and follow its layouts rather than introducing TV-specific variants.
Also fixes two pre-existing bugs found while building this, both in the same Continue Watching rendering path.

## PR type

- [x] Critical bug fix

Contains two bug fixes alongside the approved card style feature request linked below.

## Why
A user reported that the Continue Watching row does not follow the poster sizing used by the catalogs below it, and how it is in NuvioMobile. Continue Watching already scales with the poster width setting and is pixel-identical to a catalog card in landscape mode, but stays locked to 16:9 in portrait because `continueWatchingScale` and the aspect ratio were hardcoded. NuvioMobile solves this with a dedicated card style picker rather than by inheriting the catalog orientation, so this follows the same approach and naming.

**Corner radius was ignored.** `ContinueWatchingSection` used hardcoded `CwCardShape` and `CwClipShape` values for the card, its focus border, its depth surface and the artwork clip, so users with poster corners set to sharp still saw rounded Continue Watching cards while every catalog row respected the setting.

**Artwork went stale after toggling episode thumbnails.** Episode artwork is chosen during enrichment in `HomeViewModelContinueWatching` and persisted in `ContinueWatchingEnrichmentCache`, but toggling the setting only updated ui state without clearing that cache or retriggering the pipeline. Items kept whichever artwork was baked in at their last enrichment, so the row showed a mix of old and new choices depending on when each item was last refreshed. The sibling `nextUpFromFurthestEpisode` setting already clears the caches on change; this now does the same.

## Issue or approval
Fixes #2911

## Reproduction steps
Corner radius:

1. Set Settings - Layout - poster corner radius to sharp.
2. Return to home and compare the catalog rows against the Continue Watching row.
3. Catalog posters are square-cornered, Continue Watching cards stay rounded.

Stale artwork:

1. Toggle "Use Episode Thumbnails in Continue Watching".
2. Some items update and others keep their previous artwork, varying by when each was last enriched.

## UI / behavior impact

The new styles are opt-in behind a single setting that defaults to current behavior.

## Policy check

## Poster card style
Follows the NuvioMobile poster card rather than reusing the landscape layout:
- Title sits below the artwork, outside the card depth surface, so the focus ring and depth wrap only the poster.
- Title wraps to two lines, in a fixed height row so a short title does not make its card shorter than the rest.
- Season and episode code sits to the right of the title on the same row, and is omitted for movies.
- Progress bar is lifted off the bottom edge and sits on a rounded pill inside the artwork.
- Artwork uses the portrait catalog dimensions, so it lines up with the catalog posters below it. The two line title row adds roughly 28dp to the row height, which is inherent to placing the title under the artwork and matches how the mobile card is measured.
- The episode thumbnail setting does not apply. Poster cards always show poster art, and the toggle hides itself while Poster is selected.

## Wide card style
Ported from NuvioMobile at a reviewer's request. Artwork sits in a poster-shaped strip on the left, with the title, season and episode code, episode title and progress bar stacked beside it. Keeps the 2.5:1 shape of the mobile card.
It reuses the existing artwork resolution, cache key and prefetch path rather than introducing a second one, so the prefetch warms exactly what the card renders. Dimensions derive from the poster width setting instead of the fixed pixel sizes the mobile card uses, so Wide scales with the user's poster size like the other styles.
## Artwork precedence, and why it is not simply consistent
Poster and Wide cards show 2:3 artwork while episode thumbnails and backdrops are 16:9, so landscape artwork loses roughly 45 percent of each side when cropped into a poster frame. The precedence is therefore deliberate rather than uniform:
| | thumbnails on | thumbnails off |
| --- | --- | --- |
| Card | episode thumbnail, backdrop for movies | backdrop |
| Poster | poster | poster |
| Wide | episode thumbnail for series, poster for movies | poster |

Poster always shows poster art and ignores the episode thumbnail setting, matching mobile, and the option hides itself while Poster is selected. Wide keeps the setting because its artwork strip is small enough that a cropped still reads. 

Making every style resolve identical artwork was tried and reverted during testing: it produced heavily cropped backdrops in the poster frame.

## Scope boundaries
Intentionally not changed:
- Modern home layout only. Classic and Grid render Continue Watching through separate composables and keep current behaviour, since the new parameter defaults to today's value.
- The landscape Card style is untouched, including its title size, overlay text and flush progress bar.
- No changes to catalog cards, the hero, sorting, or the Continue Watching data pipeline beyond the cache invalidation described above.

## Breaking changes
none

## Testing
Device: Xiaomi Mi TV (MiTV-AFMU0), Android 14

- Verified all three card styles against the resolved artwork URLs with the episode thumbnail setting on and off, confirming series take the episode still when enabled and movies keep their poster.
- Confirmed the focus ring and card depth surround only the artwork on Poster, and the whole card on Card and Wide.
- Confirmed every Poster card is the same size whether or not it carries a season and episode code, and every Wide card is the same size for in progress, Up Next and New Season items.
- Set poster corners to sharp and confirmed the cards, focus ring and artwork clip square off with the catalogs.
- Toggled the episode thumbnail setting repeatedly and confirmed every item updates immediately rather than a subset.
- Confirmed the unwatched episode blur applies only when a card resolved to an episode thumbnail.
- Confirmed long titles scroll on focus after a short delay and reset when focus moves away on Card and Wide, and ellipsize on Poster where the title wraps to two lines.
- Confirmed the prefetch and the card resolve the same artwork and cache key in every style, so switching style does not leave the row decoding on the scroll frame.
- Enabled episode thumbnails on Card, switched to Poster, and confirmed the cards still show poster art and the toggle disappears from the settings list.

## Screenshots / Video

- Home Screen - Poster

- Home Screen - Wide

- Settings 

Card style previews contributed by @halibiram

- Mobile app

- Bug fix - Corner radius was ignored

## Linked issues

Fixes #2911

Also addresses user reports from discord:
- Corner radius ignored on Continue Watching and Upcoming — the user report was "I have posters set to 'sharp' giving square corners, and that works on all my catalogs, but not continue watching or upcoming."
- Continue Watching artwork going stale when the episode-thumbnail toggle changes.

Implements [discord](https://discord.com/channels/1379902184207941732/1529681453909020703)


skoruppa:
I don't like those additional options :P 

And if we are doing this, we need to port the "wide" as well

@tapframe need your input ;) Maybe you want to take care of it?

ram130:
Let me update now.

ram130:
> 
> And if we are doing this, we need to port the "wide" as well
> 
> 

Ported Wide as requested. 



----

skoruppa:
> > And if we are doing this, we need to port the "wide" as well
> 
> Ported Wide as requested.
>  

why is your wide, so wide? >< It also seems to have less height? Season/Episode Number should be in different line than episode title. 

And just a pro tip - you can make screenshots through adb 
```adb exec-out screencap -p > screenshot.png```

With the "Poster" style, you also need to make a perfect copy of the mobile design, right now it just looks off. The title should be under the card depth, not inside. It is missing season and episode number (or at least no items with it are shown on you picuters), the progress bar should be a bit higher

tapframe:
yeah, no additional options. if we're trying to bring it to parity with mobile, we shouldn't introduce extras. the other inputs have already been addressed by skoruppa.

ram130:
> > > And if we are doing this, we need to port the "wide" as well
> > 
> > 
> > Ported Wide as requested.
> >  
> 
> why is your wide, so wide? >< It also seems to have less height? Season/Episode Number should be in different line than episode title.
> 
> And just a pro tip - you can make screenshots through adb `adb exec-out screencap -p > screenshot.png`
> 
> With the "Poster" style, you also need to make a perfect copy of the mobile design, right now it just looks off. The title should be under the card depth, not inside. It is missing season and episode number (or at least no items with it are shown on you picuters), the progress bar should be a bit higher


> yeah, no additional options. if we're trying to bring it to parity with mobile, we shouldn't introduce extras. the other inputs have already been addressed by skoruppa.


Thanks, I made the changes, removed the extra card info option (title options) and redid the implementation to match mobile, with slight adjustments for the bigger screen on TV.

skoruppa:
I think it is still too wide. Even with longer titles you have so much empty space on the right. And on mobile the "Poster" style has title in two lines.

ram130:
> I think it is still too wide. Even with longer titles you have so much empty space on the right. And on mobile the "Poster" style has title in two lines.

Ok I've made the change to wide and updated poster to two lines.

wide_card_sm
poster_card_2lines

tapframe:
looks good. one small thing. in the poster layout, the “use episode thumbnail” option should be ignored, like on mobile. it doesn’t look good there.
moving the Card Style option to the top of the list and hiding “use episode thumbnail” when “poster” is selected would be nice.mobile doesn’t hide the option either, but i’ll take care of that.

halibiram:
What if we made the card style like this?


cw_layout_v2

ram130:
> What if we made the card style like this?
> 

I liked it, It was on my list to do. Feel free to commit it here.

ram130:
> looks good. one small thing. in the poster layout, the “use episode thumbnail” option should be ignored, like on mobile. it doesn’t look good there. moving the Card Style option to the top of the list and hiding “use episode thumbnail” when “poster” is selected would be nice.mobile doesn’t hide the option either, but i’ll take care of that.

Ok, changes made.

halibiram:
> > What if we made the card style like this?
> 
> I liked it, It was on my list to do. Feel free to commit it here.

I opened a pr on your fork thinking it might help you. if @tapframe and @skoruppa like it, you can merge it

skoruppa:
> > > What if we made the card style like this?
> > 
> > 
> > I liked it, It was on my list to do. Feel free to commit it here.
> 
> I opened a pr on your fork thinking it might help you. if @tapframe and @skoruppa like it, you can merge it

I like it :) lets add it to the PR. Although the placeholders on mobile looks better :p

ram130:
> > > > What if we made the card style like this?
> > > 
> > > 
> > > I liked it, It was on my list to do. Feel free to commit it here.
> > 
> > 
> > I opened a pr on your fork thinking it might help you. if @tapframe and @skoruppa like it, you can merge it
> 
> I like it :) lets add it to the PR. Although the placeholders on mobile looks better :p

ok made the changes and added to PR.

Good one @halibiram

PR #2936, opened 2026-08-06
fix(search): only remember explicitly submitted searches in recent history (#2928)

### LINKED ISSUE #2928: [Bug]: Unwanted entries are added to "Recent Searches"
### Pre-flight checks

### App version / commit

0.8.2

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Kinhank G1

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

Other

### Bug description

There are entries that are added to my "recent searches" without me having searched for them...

### Steps to reproduce

1. Go to "Search".
2. Search "dog" and delete character by character.
3. there is "do" and "dog" in my recent search.

### Expected behavior

Nothing Added

### Actual behavior

"do" and "dog" are added.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

supergera13:
## Summary

Search only records an entry in Recent Searches when the user explicitly confirms the query (Done / voice / selecting a suggestion or recent item / moving focus from the text input into live results). The live search that runs per keystroke no longer saves every intermediate prefix into history.

## PR type

- [x] Critical bug fix

## Why

Typing e.g. "dog" and then deleting character by character saved "do" and "dog" into Recent Searches even though the user never confirmed a search. Because the live search fires per keystroke (debounced), and every run that returned results persisted its query, any prefix with results became a separate history entry. The expected behaviour, per the report, is that simply typing and deleting adds nothing to history.

## Issue or approval

Fixes #2928

## Reproduction steps

1. Open Search.
2. Type "dog", then delete one character at a time until empty.
3. Open Recent Searches.

Result before fix: "do" and "dog" both appear, even though neither was confirmed.
Result after fix: nothing is added unless the user explicitly submits the search.

## UI / behavior impact

- [x] Behavior change: Recent Searches is now only updated on explicit search submission, not on every live-search keystroke, so partial/prefix queries the user did not confirm no longer pollute history.

## Policy check

- [x] Fix is limited to the described bug (single issue).
- [x] Change is minimal and does not overlap other open PRs.
- [x] Equivalent duplicate issues check-up before working.

## Scope boundaries

No change to suggestion generation, discover flow, or how results render. The change only gates where the query gets written into Recent Searches: live-search runs no longer persist; explicit submit and D-pad focus transfer from the text input into live results do. Opening a specific result is not a separate history trigger.

## Testing

Verified with `./gradlew :app:testFullDebugUnitTest --tests "com.nuvio.tv.ui.screens.search.*" :app:compileFullDebugKotlin` (**BUILD SUCCESSFUL**, 5 tests, 0 failures). `SearchViewModelConcurrencyTest` now covers the review case: after live-search results arrive, moving from the text input into results saves the confirmed query exactly once.

## Screenshots / Video

Not a UI change.

## Breaking changes

No breaking changes.

## Linked issues

- Fixes #2928

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/search/SearchScreen.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/search/SearchScreen.kt
@@ -528,7 +528,12 @@ fun SearchScreen(
                     isVoiceListening = isVoiceListening,
                     voiceRmsLevel = voiceRmsLevel,
                     onVoiceSearch = launchVoiceSearch,
-                    onMoveToResults = { focusResults = true },
+                    onMoveToResults = {
+                        // D-pad down from the text field is the user's confirmation that the
+                        // live-search results are useful, even before a particular card opens.
+                        viewModel.onEvent(SearchEvent.RememberSearchFromTextInput)
+                        focusResults = true
+                    },
                     onOpenDiscover = onOpenDiscover,
                     showDiscoverButton = uiState.discoverLocation == DiscoverLocation.IN_SEARCH,
                     keyboardController = keyboardController,
@@ -722,9 +727,6 @@ fun SearchScreen(
                                 },
                                 onItemClick = { id, type, addonBaseUrl ->
                                     lastFocusedRowKey = catalogKey
-                                    // Entering a result confirms the query for recent history
-                                    // (keyboard Done is not required after live search).
-                                    viewModel.onEvent(SearchEvent.RememberSearchFromResults)
                                     // Save focus state to ViewModel before navigating
                                     viewModel.savedFocusRowKey = catalogKey
                                     viewModel.savedFocusItemIndex = searchRowFocusedItemIndex[catalogKey] ?: 0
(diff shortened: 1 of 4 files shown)


skoruppa:
Entering the search result should also count as "submitting"

supergera13:
## Review follow-up

@skoruppa addressed: entering a search result now counts as submitting for recent history.

Current behavior:

1. Live search does not write history, avoiding intermediate prefixes such as `do` / `dog`.
2. Keyboard Done / Submit writes history when results are non-empty.
3. Opening a result also writes the current query to recent history.

The branch is rebased, mergeable, and the current Full Debug workflow passes.

app/src/main/java/com/nuvio/tv/ui/screens/search/SearchScreen.kt
```diff
@@ -722,6 +722,9 @@ fun SearchScreen(
                                 },
                                 onItemClick = { id, type, addonBaseUrl ->
                                     lastFocusedRowKey = catalogKey
+                                    // Entering a result confirms the query for recent history
```
skoruppa: What I meant is going out from the text input into results view below it, not "entering a specific result"

supergera13: Addressed in `0146f86e`: pressing D-pad Down to move focus from the search text field into the live results now confirms and saves the query. Opening a specific result no longer performs that save. I also added a ViewModel regression test; the Search unit-test suite and `compileFullDebugKotlin` pass.

PR #2938, opened 2026-08-06
fix(search): prevent stale searches from hanging (#2917)

### LINKED ISSUE #2917: Multi-word Search Broken on Fire TV
### Pre-flight checks

### App version / commit

Latest release 

### Install method

Other (please describe below)

### Platform

Android TV

### Device model

Fire TV Stick 4K 

### Android version

Android 11

### Player mode

Internal player

### Area (tag)

Streams / Sources (selection, loading, errors)

### Bug description

The search function has stopped working correctly on Firestick.

I am running the latest version of Nuvio on an Amazon Fire TV Stick. The issue appears to have started today (4 August 2026). Search was working normally previously.

The key issue is that single-word searches work correctly, but searches containing multiple words result in an infinite loading state.

### Steps to reproduce

1. Open Nuvio on Firestick.
2. Navigate to Search.
3. Search for a single-word title, e.g.:
    * Deep
4. Results appear immediately.
5. Now search for the full title (voice search):
    * Deep Cover
6. The search remains on the loading screen indefinitely and no results are returned.

### Expected behavior

Multi-word search queries should return results in the same way as single-word searches.

For example:

Deep Cover → should return the movie Deep Cover.

### Actual behavior

Single-word searches work normally, but adding a second word causes the search to enter an infinite loading state.

Image
Image

### Frequency

Always

### Did this work before?

Yes, it used to work

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

supergera13:
## Summary

Prevent an older live-search run from resuming after a newer query and cancelling the newer query's catalog jobs. The complete search run is now tracked and cancelled as one structured coroutine scope, while a generation token prevents stale results or completion handlers from updating current UI state.

A deterministic JVM regression test reproduces the out-of-order `Deep` -> `Deep Cover` sequence reported on Fire TV.

## PR type

- [x] Critical bug fix

## Why

`performSearch` previously launched the installed-addon lookup in an untracked `viewModelScope` coroutine and tracked only the later catalog jobs. If the lookup for an older query resumed after a newer multi-word query had already started, the older run could cancel the newer catalog jobs and replace them with stale work.

The stale run's completion was then ignored because `submittedQuery` already contained the newer query, leaving `isSearching` set to `true` indefinitely. This matches the reported behavior where a single-word search succeeds but submitting the full multi-word title remains on the loading state.

The Stremio path encoding was also checked independently: Cinemeta returned HTTP 200 with eight results for `Deep Cover` using both `%20` and `+`, so the failure is not caused by encoding.

## Issue or approval

Fixes #2917

## Reproduction steps

1. Open Search on Fire TV.
2. Submit a single-word query such as `Deep`.
3. Quickly replace it with and submit a multi-word query such as `Deep Cover` (the Fire TV keyboard or voice input can trigger this timing).
4. Under the affected timing, the older addon lookup resumes after the newer search starts.
5. Before this fix, the newer catalog jobs are cancelled and Search remains loading indefinitely.
6. After this fix, only `Deep Cover` completes and the loading state clears.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

This changes only Search run lifecycle and stale-result protection. It does not change query encoding, catalog URL construction, result ordering, search presentation, suggestions, history behavior, addon configuration, or any UI styling.

## Testing

- Regression proof before the production change: `SearchViewModelConcurrencyTest` failed after the delayed old lookup resumed because `isSearching` became `true` and stale `Deep` work was observed.
- `./gradlew :app:testFullDebugUnitTest --tests "com.nuvio.tv.ui.screens.search.SearchViewModelConcurrencyTest" --rerun-tasks` — passed after the fix.
- `./gradlew :app:testFullDebugUnitTest --tests "com.nuvio.tv.ui.screens.search.*"` — passed.
- `./gradlew :app:compileFullDebugKotlin` — passed.
- `git diff --check` — passed.
- No Fire TV, Android emulator, or physical device was used; the reported ordering is reproduced deterministically with controlled coroutine scheduling and addon lookup completion.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

- Fixes #2917

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/search/SearchViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/search/SearchViewModel.kt
@@ -29,6 +29,7 @@ import com.nuvio.tv.domain.model.PLACEHOLDER_IMAGE_URL
 import com.nuvio.tv.domain.repository.CatalogRepository
 import dagger.hilt.android.lifecycle.HiltViewModel
 import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -71,6 +72,9 @@ class SearchViewModel @Inject constructor(
     private val catalogOrder = mutableListOf<String>()
 
     private var activeSearchJobs: List<Job> = emptyList()
+    private var searchRunJob: Job? = null
+    private var activeSearchQuery: String? = null
+    private var searchGeneration = 0L
     private var discoverJob: Job? = null
     private var catalogRowsUpdateJob: Job? = null
     private var suggestionJob: Job? = null
@@ -201,6 +205,7 @@ class SearchViewModel @Inject constructor(
                 // An explicit retry must refetch even though nothing about the request changed.
                 lastRequestKey = null
                 lastCompletedRequestKey = null
+                cancelSearchRun()
                 performSearch(uiState.value.submittedQuery.ifBlank { uiState.value.query })
             }
         }
@@ -221,8 +226,7 @@ class SearchViewModel @Inject constructor(
         }
 
         // Drop in-flight requests for the previous keystroke before scheduling the next run.
-        activeSearchJobs.forEach { it.cancel() }
-        activeSearchJobs = emptyList()
+        cancelSearchRun()
 
         // Live search: results follow what you type, like mobile. Debounced because each run hits
         // every enabled addon catalog.
@@ -337,6 +341,15 @@ class SearchViewModel @Inject constructor(
         pendingCatalogResponses = 0
     }
 
+    private fun cancelSearchRun() {
+        searchGeneration++
+        searchRunJob?.cancel()
+        searchRunJob = null
+        activeSearchJobs.forEach { it.cancel() }
+        activeSearchJobs = emptyList()
+        activeSearchQuery = null
+    }
+
     /**
      * Identifies a search by everything that changes what it would return: the query, the released
      * filter, and the exact set of catalogs it would hit. Enabling an addon or flipping the filter
@@ -370,8 +383,7 @@ class SearchViewModel @Inject constructor(
         }
 
         if (query.length < MIN_SEARCH_QUERY_LENGTH) {
-            activeSearchJobs.forEach { it.cancel() }
-            activeSearchJobs = emptyList()
+            cancelSearchRun()
             catalogRowsUpdateJob?.cancel()
             resetCatalogAccumulator()
             lastRequestKey = null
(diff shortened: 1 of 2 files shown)


skoruppa:
@amrhik can you test the apk from https://github.com/NuvioMedia/NuvioTV/actions/runs/31091507760/artifacts/8963974015 to confirm if the fix changes anything on your side?

amrhik:
> @amrhik can you test the apk from https://github.com/NuvioMedia/NuvioTV/actions/runs/31091507760/artifacts/8963974015 to confirm if the fix changes anything on your side?

-
@skoruppa 
Tested. Multi word search is working on this.

amrhik:
@skoruppa  The APK linked had an issue with log in. No QR code was loading to let me log in. I tested the search function while I was not logged in.

amrhik:
@skoruppa Ive updated to 0.8.3 Beta and the issue still persists for me

supergera13:
@amrhik thanks for the follow-up. `0.8.3-beta` does contain the #2938 fix, so this is likely a second path rather than the old stale-query race returning.

The important difference is that the PR APK worked while you were signed out, whereas the release fails after your account/addons are restored. I am investigating whether one enabled search catalog keeps the aggregate search in loading.

Could you please check these two points on 0.8.3 or newer?

1. Temporarily disable every non-default addon, leaving Cinemeta enabled, then retry `Deep` → `Deep Cover`.
2. If that works, re-enable the search-capable addons one at a time and tell us which addon makes it fail (configuration URL/details can be sanitized). Also note whether loading eventually ends after about 60 seconds or truly never ends.

That will identify the failing catalog without exposing account data. I am preparing a deterministic test for the authenticated multi-addon case in parallel.

supergera13:
Follow-up PR is now open as draft: #3017.

It covers the deterministic authenticated multi-addon failure: one synced search catalog can stay open forever and keep the whole screen loading after other catalogs returned. The patch gives each catalog a 30-second budget and preserves responsive results.

@amrhik once the Full Debug artifact for #3017 is available, testing it while signed in with your normal synced addons would be the most useful confirmation. The disable/re-enable check from my previous comment can still identify the specific stalled addon.

supergera13:
@amrhik the signed-in test APK for follow-up PR #3017 is ready:

https://github.com/NuvioMedia/NuvioTV/actions/runs/31681305735/artifacts/9173904730

Please test it while signed in with your normal synced addons, using the same `Deep` → `Deep Cover` sequence. If it still fails, please note whether it settles after 30 seconds and whether disabling non-default addons changes the result.

CI Full Debug and the deterministic regression test both pass.

amrhik:
@supergera13 

Firstly I tried disabling all addons except Cinemeta on the latest version but the issue still persists. 

The test version is not letting me log in as the QR code does not load even after refreshing 
IMG_0218

amrhik:
@supergera13 Update:

i reinstalled Cinemeta from the official source on my latest version of Nuvio and this has resolved the issue

supergera13:
@amrhik thank you — reinstalling Cinemeta from the official source and seeing the issue disappear on the current release is the decisive isolation step.

That changes the diagnosis: your remaining case was tied to the previously installed/restored Cinemeta entry or configuration, rather than the stale-search race fixed by #2938. No further test is needed from you unless the problem returns after the official reinstall.

The missing QR in the contributor PR APK is also explained: fork PR artifacts are built without the repository's private authentication configuration, so they cannot be used for a signed-in account test. Sorry for sending you through that dead end.

I am closing #3017 rather than presenting its generic timeout as a confirmed fix for your report. If the issue returns with the official Cinemeta installation, the installed manifest URL/version and a debug log would be the useful next evidence.

supergera13:
A narrower follow-up is now available as draft #3021.

Unlike the closed timeout proposal, it covers the production path supported by the Cinemeta reinstall result: a catalog emits `Loading` and then a terminal `Error`, but Search was leaving that catalog's old loading row in the UI. The fix clears the failed row, preserves results from working catalogs, and shows the global error only when no catalog returned results.

Regression coverage executes both mixed success/error and all-error cases (3 Search tests, 0 failures/errors). No further action is needed from @amrhik unless the issue returns with the official Cinemeta installation.

PR #3015, opened 2026-08-12
fix(continue-watching): resume from saved position instead of 0:00

halibiram:
## Summary

Continue Watching resume no longer starts at `0:00` when a saved position exists.

1. **Merge local progress with tracking providers.** Resume no longer uses the first empty Trakt/Simkl episode snapshot. `getResumeProgress` waits for provider `allProgress` (up to 2s) and picks with `WatchProgressLookup`: a local millisecond `position` wins over a percent-only Trakt/Simkl row (`pos=0`). That is the `pick=local_position_over_provider_percent` path in NUV-0EECB3 / NUV-CDB7E3. The player loads that row in `loadSavedProgressSuspend` *before* `initializePlayer`, then `resolveInitialResumePosition` passes it into `setMediaSource`.
2. **Keep pending resume until the stream is seekable.** HLS/debrid often is not seekable on the first `STATE_READY`. The old code cleared `pendingResumeProgress` there, so the seek was lost. `WatchProgressResumePolicy.decideReadySeek` now returns `KeepPending` until the item is seekable (or duration is known for percent-only rows), then seeks.
3. **Ignore empty remote sync rows.** A newer remote row with `position=0, duration=0` no longer overwrites a usable local save (`WatchProgressSyncReducer`). `isInProgress()` also treats `position > 0` with unknown duration as in-progress, so paused saves remain resume targets.

NUV-0EECB3 and NUV-CDB7E3 both land on (1): local ms start position, no extra READY seek needed. (2) and (3) cover the other 0:00 

## PR type

- [x] Critical bug fix

## Why

Opening a title from Continue Watching showed the in-progress card, but playback started at the beginning. 

The failure was confirmed on device with a **temporary playback issue report** (existing `schemaVersion: 1` payload, no new backend fields). Those captures are **NUV-0EECB3** and **NUV-CDB7E3**: series episodes on Amazon Fire TV, ExoPlayer, Trakt connected, with both a local millisecond save and a Trakt percent-only row (`pos=0`). Before this fix that mix dropped resume: the episode snapshot emitted empty before Trakt loaded, or the percent-only Trakt row replaced the local position, so `setMediaSource` started at `0`.

- **NUV-0EECB3** — local `pos=721260` (~12:01). After this fix the first frame is at `721260 ms` (`pick=local_position_over_provider_percent`, `cause=APPLIED`).
- **NUV-CDB7E3** — local `pos=944151` (~15:44, ~77% of the real file). Trakt is percent-only `32.53%` on a `1500000 ms` estimate (~6:40 on the real duration). After this fix playback starts at `944151 ms`, not the Trakt percent. This is the case where preferring local milliseconds actually changes the resume point.

The temporary report path is not part of this PR; it was only used to collect those IDs.

## Issue or approval

Field reports (temporary playback issue report): NUV-0EECB3, NUV-CDB7E3

## Reproduction steps

1. Sign in to Trakt (or Simkl) so Continue Watching uses a tracking provider.
2. Play a series episode for several minutes, then exit before the end.
3. Confirm the title appears on Continue Watching with progress.
4. Open it again from Continue Watching (`startFromBeginning=false`).
5. **Before this fix:** player starts at `0:00` even though a local save exists .
6. **After this fix / NUV-0EECB3:** playback starts at the saved local position (`721260 ms` / ~12:01), not `0:00`.
7. **After this fix / NUV-CDB7E3:** playback starts at the local save (`944151 ms` / ~15:44), not Trakt’s `32.53%` (~6:40). Local millisecond progress is kept over the Trakt percent-only row.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

- Player UI, overlays, and settings are unchanged.
- No new playback-report schema fields; backend payload stays `schemaVersion: 1`.
- Does not change Continue Watching row layout or catalog ranking beyond using the same saved progress the player now honors.
- The temporary playback issue report used to collect NUV-0EECB3 / NUV-CDB7E3 is not included in this PR.

## Testing

- Unit: `WatchProgressResumePolicyTest`, `WatchProgressSyncReducerTest`, extra `WatchProgressTest` / `WatchProgressPreferencesStorageTest` cases for in-progress-with-unknown-duration and empty-remote merge.
- Field reports were taken with a temporary playback issue report (same device: Amazon Fire TV, ExoPlayer, Trakt):
  - **NUV-0EECB3** — resume applied at `721260 ms`; report position `728969 ms`; first frame already at the saved position. `loadSavedProgress=true`, `resumePositionMs=721260`.
  - **NUV-CDB7E3** — local `944151 ms` preferred over Trakt `pct=32.5302` / `dur=1500000`; first frame and `resumePositionMs=944151`; report position `949614 ms`. Using the Trakt percent on the real file duration would have resumed near `6:40` instead of `15:44`.
- Manual path covered by both reports: series episode, local position preferred over Trakt percent-only, no tunneling, `setMediaSource` start position sufficient (no extra READY seek).

Relevant excerpts from the temporary playback issue reports (titles omitted):

**NUV-0EECB3**

```json
{
  "device": { "manufacturer": "Amazon", "model": "AFTKRT", "androidRelease": "11" },
  "content": { "contentType": "series", "season": 5, "episode": 9 },
  "player": { "engine": "EXOPLAYER", "positionMs": 728969, "durationMs": 2581704 },
  "playbackAnalytics": {
    "startPositionMs": 721260,
    "rawEventLines": [
      "PREPARE_PLAYBACK: loadSavedProgress=true currentSeason=5 currentEpisode=9",
      "PLAYER_INIT: engine=EXOPLAYER resumePositionMs=721260 tunneling=false",
      "EXO_EVENT: name=rendered_first_frame elapsedMs=9416 positionMs=721260",
      "RESUME_DIAG: cause=APPLIED engine=EXOPLAYER S5E9 providerId=trakt providerTimedOut=false pick=local_position_over_provider_percent local=src=local;pos=721260;dur=2581704;pct=-1.0;S5E9 provider=src=trakt_playback;pos=0;dur=2580000;pct=27.9367;S5E9 startFromBeginning=false loadRequested=true resumable=true initialMs=721260 keptPending=false seekIssued=true seekTargetMs=721260 firstFrameMs=721260 settledMs=721260 expectedMs=721260 reportMs=728969 reportDurMs=2581704"
    ]
  }
}
```

**NUV-CDB7E3**

```json
{
  "device": { "manufacturer": "Amazon", "model": "AFTKRT", "androidRelease": "11" },
  "content": { "contentType": "series", "season": 9, "episode": 12 },
  "player": { "engine": "EXOPLAYER", "positionMs": 949614, "durationMs": 1230176 },
  "playbackAnalytics": {
    "startPositionMs": 944151,
    "rawEventLines": [
      "PREPARE_PLAYBACK: loadSavedProgress=true currentSeason=9 currentEpisode=12",
      "PLAYER_INIT: engine=EXOPLAYER resumePositionMs=944151 tunneling=false",
      "EXO_EVENT: name=rendered_first_frame elapsedMs=9013 positionMs=944151",
      "RESUME_DIAG: cause=APPLIED engine=EXOPLAYER S9E12 providerId=trakt providerTimedOut=false pick=local_position_over_provider_percent local=src=local;pos=944151;dur=1230176;pct=-1.0;S9E12 provider=src=trakt_playback;pos=0;dur=1500000;pct=32.5302;S9E12 startFromBeginning=false loadRequested=true resumable=true initialMs=944151 keptPending=false seekIssued=true seekTargetMs=944151 firstFrameMs=944151 settledMs=944151 expectedMs=944151 reportMs=949614 reportDurMs=1230176"
    ]
  }
}
```

## Screenshots / Video

Not a UI change

## Breaking changes

None.

## Linked issues

Playback issue reports (temporary capture): NUV-0EECB3, NUV-CDB7E3

### DIFF
--- a/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
+++ b/app/src/main/java/com/nuvio/tv/core/player/ExternalPlaybackTracker.kt
@@ -12,6 +12,7 @@ import com.nuvio.tv.core.tracking.buildTrackingMediaReference
 import com.nuvio.tv.data.local.PlayerSettingsDataStore
 import com.nuvio.tv.domain.model.Video
 import com.nuvio.tv.domain.model.WatchProgress
+import com.nuvio.tv.domain.model.WatchProgressLookup
 import com.nuvio.tv.domain.repository.MetaRepository
 import com.nuvio.tv.domain.repository.WatchProgressRepository
 import com.nuvio.tv.data.repository.SkipIntroRepository
@@ -629,12 +630,11 @@ class ExternalPlaybackTracker @Inject constructor(
     }
 
     private suspend fun currentSavedProgress(metadata: ExternalPlaybackMetadata): WatchProgress? {
-        val flow = if (metadata.season != null && metadata.episode != null) {
-            watchProgressRepository.getEpisodeProgress(metadata.contentId, metadata.season, metadata.episode)
-        } else {
-            watchProgressRepository.getProgress(metadata.contentId)
-        }
-        return flow.firstOrNull()
+        return watchProgressRepository.getResumeProgress(
+            contentId = metadata.contentId,
+            season = metadata.season,
+            episode = metadata.episode
+        )
     }
 
     private suspend fun fetchRuntimeMsFromMeta(metadata: ExternalPlaybackMetadata): Long {
@@ -1158,13 +1158,12 @@ class ExternalPlaybackTracker @Inject constructor(
     }
 
     private suspend fun getResumePosition(metadata: ExternalPlaybackMetadata): Long {
-        val flow = if (metadata.season != null && metadata.episode != null) {
-            watchProgressRepository.getEpisodeProgress(metadata.contentId, metadata.season, metadata.episode)
-        } else {
-            watchProgressRepository.getProgress(metadata.contentId)
-        }
-        val wp = flow.firstOrNull() ?: return 0L
-        if (wp.isCompleted()) return 0L
+        val wp = watchProgressRepository.getResumeProgress(
+            contentId = metadata.contentId,
+            season = metadata.season,
+            episode = metadata.episode
+        ) ?: return 0L
+        if (!WatchProgressLookup.isResumable(wp)) return 0L
         return if (wp.duration > 0L) {
             wp.resolveResumePosition(wp.duration)
         } else {
--- a/app/src/main/java/com/nuvio/tv/core/sync/WatchProgressSyncReducer.kt
+++ b/app/src/main/java/com/nuvio/tv/core/sync/WatchProgressSyncReducer.kt
@@ -1,0 +1,156 @@
+package com.nuvio.tv.core.sync
+
+import com.nuvio.tv.domain.model.WatchProgress
+
+/**
+ * Pure watch-progress sync merge rules used by [WatchProgressSyncService] and
+ * [com.nuvio.tv.data.local.WatchProgressPreferences]. Kept free of Android /
+ * Supabase so the Continue Watching resume-from-0 regressions can be reproduced
+ * in unit tests.
+ */
+internal object WatchProgressSyncReducer {
+    const val MIN_USABLE_POSITION_MS = 1_000L
+    const val MIN_USABLE_PERCENT = 2f
+
+    fun hasUsableResumeData(progress: WatchProgress): Boolean {
+        if (progress.position >= MIN_USABLE_POSITION_MS) return true
+        val percent = progress.progressPercent ?: return false
+        return percent >= MIN_USABLE_PERCENT && percent < 100f
+    }
+
+    fun shouldPushToRemote(progress: WatchProgress): Boolean {
+        val dummyCompletedSentinel = progress.position <= 1L &&
+            progress.duration <= 1L &&
+            progress.duration > 0L
+        if (dummyCompletedSentinel) return false
+        return hasUsableResumeData(progress) || progress.isCompleted()
+    }
+
+    fun episodeKey(contentId: String, season: Int, episode: Int): String {
+        return "${contentId}_s${season}e${episode}"
+    }
+
+    fun isSeriesType(contentType: String): Boolean {
+        return contentType.lowercase() in setOf("series", "tv")
+    }
+
+    fun canonicalizeForRemote(rawEntries: Map<String, WatchProgress>): Map<String, WatchProgress> {
+        if (rawEntries.isEmpty()) return rawEntries
+
+        val canonical = rawEntries.toMutableMap()
+        rawEntries.forEach { (key, progress) ->
+            val isSeriesMirrorKey = key == progress.contentId &&
+                isSeriesType(progress.contentType) &&
+                progress.season != null &&
+                progress.episode != null
+            if (!isSeriesMirrorKey) return@forEach
+
+            val season = progress.season ?: return@forEach
+            val episode = progress.episode ?: return@forEach
+            val episodeProgress = rawEntries[episodeKey(progress.contentId, season, episode)]
+                ?: return@forEach
+
+            val exactMirror = progress.position == episodeProgress.position &&
+                progress.duration == episodeProgress.duration &&
+                progress.lastWatched == episodeProgress.lastWatched
+            val episodeIsAtLeastAsFresh = episodeProgress.lastWatched >= progress.lastWatched - 1_000L
+
+            if (exactMirror || episodeIsAtLeastAsFresh) {
+                canonical.remove(key)
+            }
+        }
+
+        return canonical.filterValues(::shouldPushToRemote)
+    }
+
+    fun normalizePulledEntries(
+        entries: List<Pair<String, WatchProgress>>
+    ): List<Pair<String, WatchProgress>> {
+        if (entries.isEmpty()) return entries
+
+        val byKey = linkedMapOf<String, WatchProgress>()
+        entries.sortedByDescending { it.second.lastWatched }
+            .forEach { (key, progress) ->
+                val existing = byKey[key]
+                if (existing == null || progress.lastWatched > existing.lastWatched) {
+                    byKey[key] = progress
+                }
+            }
+
+        val latestEpisodeByContent = byKey.entries
+            .asSequence()
+            .mapNotNull { (key, progress) ->
+                if (isSeriesType(progress.contentType) &&
+                    progress.season != null &&
+                    progress.episode != null &&
+                    key != progress.contentId
+                ) {
+                    progress
+                } else {
+                    null
+                }
+            }
+            .groupBy { it.contentId }
+            .mapValues { (_, episodes) ->
+                episodes.maxWithOrNull(
+                    compareBy<WatchProgress> { it.lastWatched }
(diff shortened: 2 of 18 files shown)


ram130:
I was working on a fix, let me review your changes as well.

ram130:
@halibiram 

Nice work — this matches what I traced independently, and the pure-policy
extraction with tests is a better shape than patching the call sites. The
Nuvio-sync `KeepLocalEmptyRemote` case and the ExternalPlaybackTracker paths
are both good catches. Four things before this comes out of draft:

1. `WatchProgressLookup.pickResumeProgress` — the
   `localHasPosition && !providerHasPosition -> resumableLocal` branch ignores
   `lastWatched`. If I watch on another device, the Trakt row is newer but
   percent-only, and a stale local position wins. Can we gate that branch on
   `local.lastWatched >= provider.lastWatched` and fall through to the provider
   otherwise?

2. `ReadySeekDecision.KeepPending` has no retry driver.
   `tryApplyPendingResumeProgress` is only called from the STATE_READY branch of
   `onPlaybackStateChanged` and from `loadSavedProgressFor`. If duration isn't
   resolved at the first STATE_READY and no second one fires, we keep pending
   forever and still start at 0:00. Could we also call it from
   `onTimelineChanged` / `onEvents(EVENT_TIMELINE_CHANGED)`? That's when
   duration actually lands. Relatedly, `!isSeekable -> KeepPending` never
   clears for live/non-seekable media, where the old `Clear` was right.

3. `getResumeProgress` awaits `provider.allProgress.first()` under a 2s timeout,
   and it's awaited in `preparePlaybackBeforeStart` before `initializePlayer`.
   Worst case that's 2s on player startup. Could the local row be used
   immediately with the provider value applied async once it arrives?

4. `isInProgress` now returns true for `position > 0 && duration <= 0` at any
   depth, and `isResumable` skips `STARTED_THRESHOLD` for any `position > 0`.
   `continueWatching` filters on `isInProgress`, and those rows have
   `progressPercentage == 0` so `isCompleted()` is never true. Is that intended,
   and do they still age out of the CW row?

Separately, worth knowing why this got more frequent recently, I mentioned on discord: 1fa091d1a made
`TraktSettingsDataStore.watchProgressSource` a `stateIn(Eagerly,
DEFAULT_WATCH_PROGRESS_SOURCE)` StateFlow (default TRAKT) and
`activeProgressProviderState` a `stateIn(Eagerly, null)` StateFlow, so
`.firstOrNull()` consumers latch an eagerly-seeded TRAKT provider before the
stored setting loads. Out of scope here — this PR makes the resume path robust
enough that it matters much less but it probably deserves its own issue.

tapframe:
reproduced this with extra logs and found a path difference that is not mentioned here. Direct Continue Watching failed, while starting an episode from the details screen worked. The details screen observes getAllEpisodeProgress() early, which warms the per-show cache before playback. Direct CW skips that step, so getEpisodeProgress().firstOrNull() takes the first empty episode emission even though allProgress already has the episode and percentage used by the CW card.
for the case I reproduced, combining episodeProgress with allProgress inside getEpisodeProgress() fixes it. I tested with an empty episode snapshot and 53% in allProgress, and direct CW resumed correctly.
Is there a reason this needs the broader player lifecycle, retry, and sync changes instead of fixing the lookup in the repository? Are those changes covering separate reproduced cases?

halibiram:
> reproduced this with extra logs and found a path difference that is not mentioned here. Direct Continue Watching failed, while starting an episode from the details screen worked. The details screen observes getAllEpisodeProgress() early, which warms the per-show cache before playback. Direct CW skips that step, so getEpisodeProgress().firstOrNull() takes the first empty episode emission even though allProgress already has the episode and percentage used by the CW card. for the case I reproduced, combining episodeProgress with allProgress inside getEpisodeProgress() fixes it. I tested with an empty episode snapshot and 53% in allProgress, and direct CW resumed correctly. Is there a reason this needs the broader player lifecycle, retry, and sync changes instead of fixing the lookup in the repository? Are those changes covering separate reproduced cases?

You're right, path difference you found is correct.
I couldn't clearly pinpoint exactly where issue was, so I focused on the player side resume path. It ended up feeling a bit like a forced resume attempt. field reports also showed local ms losing to Trakt percent only and the pending drop on non seekable READY, but in your case the real trigger was the empty episode snapshot in the repository lookup.
Fixing the lookup, meaning merging episodeProgress with allProgress inside getEpisodeProgress, is probably enough

ram130:
Yeah, that's a big part of it.

`observeEpisodeProgress` warms via `.onStart { scope.launch { ... } }` while the
combine emits right away from `state[cacheKey] ?: emptyMap()`, so the bare
`.firstOrNull()` in `loadSavedProgressSuspend` gets null and
`pendingResumeProgress` never gets set. Seek logic never runs at all, so the
resume policy can't help here.

MetaDetailsViewModel:888-895 already guards the same thing with
`withTimeoutOrNull(150L) { flow.first { it.isNotEmpty() } }`. The player's lookup
just never got it. Timing lines up too, since line 576 merges `optimisticProgress`
in regardless of snapshot warmth. Works inside the 3-min `optimisticTtlMs` window,
fails on a cold show after it expires.

I went back and read the description properly. NUV-0EECB3 is this same case, so it
was already covered here, and what @tapframe added is which nav path triggers it.
my earlier point about the sync reducer wanting its own PR not needed.

What's left:

`getResumeProgress` already does the lookup fix, and I think that's enough on its
own for both NUV-0EECB3 and NUV-CDB7E3.

Does `KeepPending` still have a case of its own once the lookup returns a local ms
position? Seems like it'd only bite when there's no local row at all (episode
watched on another device) and duration hasn't resolved. If it stays, there's no
retry driver. `tryApplyPendingResumeProgress` only fires from the STATE_READY
branch, so if no second STATE_READY comes it just sits pending and you still get
0:00. `onTimelineChanged` would cover that. Also `!isSeekable - KeepPending` never
clears for live/non-seekable stuff where `Clear` was the right call.

`isInProgress` widening changes `continueWatching` membership and not just resume.
Is that intended?

tapframe:
> > reproduced this with extra logs and found a path difference that is not mentioned here. Direct Continue Watching failed, while starting an episode from the details screen worked. The details screen observes getAllEpisodeProgress() early, which warms the per-show cache before playback. Direct CW skips that step, so getEpisodeProgress().firstOrNull() takes the first empty episode emission even though allProgress already has the episode and percentage used by the CW card. for the case I reproduced, combining episodeProgress with allProgress inside getEpisodeProgress() fixes it. I tested with an empty episode snapshot and 53% in allProgress, and direct CW resumed correctly. Is there a reason this needs the broader player lifecycle, retry, and sync changes instead of fixing the lookup in the repository? Are those changes covering separate reproduced cases?
> 
> You're right, path difference you found is correct. I couldn't clearly pinpoint exactly where issue was, so I focused on the player side resume path. It ended up feeling a bit like a forced resume attempt. field reports also showed local ms losing to Trakt percent only and the pending drop on non seekable READY, but in your case the real trigger was the empty episode snapshot in the repository lookup. Fixing the lookup, meaning merging episodeProgress with allProgress inside getEpisodeProgress, is probably enough

https://github.com/NuvioMedia/NuvioTV/commit/5faf2be5f55b62a4892d44167f665bc05e07e019, commit for reference.

PR #3017, opened 2026-08-13
fix(search): time out unresponsive addon catalogs (#2917 follow-up)

supergera13:
## Summary

Prevent one unresponsive search catalog from keeping the entire Search screen in a loading state indefinitely.

Each searchable addon catalog now gets a 30-second collection budget. If one catalog does not terminate, its placeholder is removed, responsive addon results stay visible, and the overall search settles normally.

## PR type

- [x] Critical bug fix

## Why

Follow-up to #2917 / merged PR #2938.

#2938 fixed a stale-search coroutine race. The reporter confirmed that PR APK worked while signed out, but later reported that 0.8.3-beta still failed after login. Login restores the synced addon list, so the authenticated session can query additional configurable search catalogs.

The remaining failure is reproducible deterministically: one catalog returns results while another emits Loading and never completes. Search currently waits for every catalog job with `joinAll()`, so the second catalog leaves `isSearching = true` forever even though usable results already exist.

## Issue or approval

Follow-up to #2917 and merged fix #2938.

## Reproduction steps

1. Use two enabled searchable catalogs.
2. Let one return normal results.
3. Let the other start but never complete.
4. Before: responsive rows appear, but global search loading/shimmer never settles.
5. After 30 seconds: the stalled catalog is dropped for that search, responsive rows remain, and loading clears.

## UI / behavior impact

- [x] No UI redesign

## Policy check

- [x] Critical bug follow-up, limited to Search request completion.
- [x] Two files only; no unrelated cleanup.
- [x] Deterministic regression coverage included.

## Scope boundaries

No change to query encoding, addon synchronization, catalog ordering, result ranking, or the stale-generation protection from #2938. Cancellation of an obsolete query still propagates normally; only the per-catalog timeout is converted into a settled catalog result.

A very slow but otherwise valid catalog may be omitted after 30 seconds for that search.

## Testing

- Added `SearchViewModelConcurrencyTest.unresponsive synced catalog cannot keep a search loading`.
- The deterministic fixture has one responsive catalog and one flow that suspends indefinitely. It verifies that loading clears and only the responsive row remains after the timeout.
- Targeted Search test: **2 tests, 0 failures/errors**.
- `./gradlew :app:compileFullDebugKotlin` — **BUILD SUCCESSFUL**.
- `git diff --check` — clean.
- Current `dev` contains four unrelated test-fixture type errors (TMDB/trailer `Flow` vs `StateFlow`). For local execution only, I applied the mechanical fixture corrections, ran the targeted Search test successfully, then removed those unrelated changes before committing.

## Device confirmation

The exact reporter addon/configuration is not available yet. The reporter has been asked to retry with only Cinemeta, then re-enable synced search addons one at a time. This PR remains draft pending that confirmation or maintainer review.

## Screenshots / Video

Not a visual change.

## Breaking changes

None.

## Linked issues

- Follow-up to #2917
- Follow-up to #2938

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/search/SearchViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/search/SearchViewModel.kt
@@ -42,6 +42,7 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.joinAll
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeoutOrNull
 import javax.inject.Inject
 
 @HiltViewModel
@@ -98,6 +99,9 @@ class SearchViewModel @Inject constructor(
          * it waits longer than the suggestion debounce to avoid a request storm per keystroke.
          */
         const val LIVE_SEARCH_DEBOUNCE_MS = 350L
+        // A synced configurable addon can leave a catalog request open indefinitely. Keep one
+        // unresponsive catalog from holding the entire search screen in its loading state.
+        const val SEARCH_CATALOG_TIMEOUT_MS = 30_000L
 
         const val MAX_SUGGESTIONS = 8
         const val MAX_RECENT_SEARCHES = 8
@@ -551,7 +555,13 @@ class SearchViewModel @Inject constructor(
 
             val jobs = searchTargets.map { (addon, catalog) ->
                 launch {
-                    loadCatalog(addon, catalog, query, generation)
+                    val completed = withTimeoutOrNull(SEARCH_CATALOG_TIMEOUT_MS) {
+                        loadCatalog(addon, catalog, query, generation)
+                        true
+                    } ?: false
+                    if (!completed) {
+                        onCatalogTimeout(addon, catalog, generation, query)
+                    }
                 }
             }
             pendingCatalogResponses = jobs.size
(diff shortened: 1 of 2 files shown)


supergera13:
## Verification update

- PR Template Check: **passed**
- PR Full Debug Build: **passed**
- Targeted `SearchViewModelConcurrencyTest`: **2 tests, 0 failures/errors**
- Tester APK: https://github.com/NuvioMedia/NuvioTV/actions/runs/31681305735/artifacts/9173904730

The reporter has been asked to test this APK while signed in with the synced addon configuration that still reproduced the issue on 0.8.3-beta. Keeping the PR draft pending that device confirmation.

supergera13:
Closing this draft after the reporter isolated the actual remaining case: reinstalling Cinemeta from the official source resolves the issue on the current release.

The proposed test used a synthetic catalog flow that emits `Loading` and then never terminates. That is not the production `CatalogRepositoryImpl` contract: its flow calls `safeApiCall`, emits a terminal `Success` or `Error`, and completes. The main OkHttp client also allows a 60-second read timeout, while this patch would cancel every catalog at 30 seconds, potentially discarding a valid slow response.

Therefore this should not be merged or presented as a #2917 follow-up without real evidence of an unbounded production request. The branch is being retained for reference; a future timeout change should be driven by logs/reproduction and aligned with the network-layer timeout policy.

PR #3088, opened 2026-08-19
fix(cw): Next Up offering unavailable episodes when "Show unaired next up episodes" is off

### LINKED ISSUE #3087: [Bug]: Next Up shows unaired (no date) episodes with "show unaired next up episodes" off
### Pre-flight checks

### App version / commit

0.8.6

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Mi Box

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

### Bug description
Continue Watching's Next Up offers an episode that has no air date and is marked unavailable, even with "Show unaired next up episodes" turned off in Layout settings.

Next will show unaired episode whenever the show no air date and is listed (like a mid season break)

### Steps to reproduce

Turn off "Show unaired next up episodes"  in Layout settings.

Search show Sistas, marked all watched to S10E11, note EP12 will show in next cue, even though no air date is available and content is unavailable. 

### Expected behavior

Not show in next up since the user has show unaired next up toggle off.

### Actual behavior

Shows the episode.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

ram130:
## Summary

Two fixes in `resolveNextUpVideoFromMeta`:

1. A missing air date now counts as unaired, so it falls under the "show unaired next up episodes" setting instead of passing as aired. This matches the season-rollover branch, which already rejects a missing date.
2. Episodes with `available == false` are skipped, with a `reason=unavailable` log line.

## PR type

- [x] Critical bug fix

## Why

With the setting off, `releaseDate?.isAfter(today) == true` evaluated to false for a null date, so the same-season branch returned `true` before `showUnairedNextUp` was ever consulted. An episode with no air date could never be filtered, regardless of the setting.

Separately, next-up never read `available`, so an episode the details screen labels "Unavailable" could still be offered.

## Issue or approval

Fixes #3087

## Reproduction steps

1. Layout settings, CW  "Show unaired next up episodes" off
2. Watch a show whose next same-season episode has no `released` date and `available: false` (Sistas, next up resolves to S10E12)
3. Open Continue Watching -> S10E12 is offered

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

No skip-forward logic added. `firstOrNull` already advances to the next qualifying episode, and returns null when none qualifies, which drops the show from the row.

`hasEpisodeAired(raw, fallback = true)` has the same null-means-aired assumption and was left alone. 
## Testing

Box S, Android 14 

- With the setting off, the reported show no longer offers the dateless unavailable episode and drops off Continue Watching.
- With the setting on, unaired next-up still appears, so the toggle is unchanged in that direction.
- Added `NextUpEpisodeEligibilityTest` (5 cases: past/today/future/missing date, and availability).

Note: `:app:testFullDebugUnitTest` does not currently compile on `dev` for reasons unrelated to this PR. `TmdbCollectionSourceResolverTest`, `TrailerServiceUseTrailersGateTest`, and `TrailerServiceYouTubeSessionCacheTest` pass `Flow<TmdbSettings>` where `StateFlow` is now expected, and `SearchViewModelConcurrencyTest` is missing a `metaRepository` argument. The new test file compiles clean, but the suite cannot be run until those are fixed.

## Screenshots / Video

Not a UI change.

Bug:
cw-before
cw-details

## Breaking changes

None.

## Linked issues

Fixes #3087

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/home/HomeViewModelContinueWatching.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/home/HomeViewModelContinueWatching.kt
@@ -2218,9 +2218,6 @@ private const val CW_NEXT_UP_NEW_SEASON_UNAIRED_WINDOW_DAYS = 7
 internal fun isNextUpEpisodeUnaired(releaseDate: LocalDate?, today: LocalDate): Boolean =
     releaseDate == null || releaseDate.isAfter(today)
 
-// Addons report available=false for episodes they cannot serve, which is the same flag the details screen labels Unavailable.
-internal fun isNextUpEpisodeAvailable(available: Boolean?): Boolean = available != false
-
 private fun resolveNextUpVideoFromMeta(
     progress: WatchProgress,
     meta: CwMetaSummary,
@@ -2274,13 +2271,6 @@ private fun resolveNextUpVideoFromMeta(
     val todayLocal = LocalDate.now(ZoneId.systemDefault())
     val watchedEpisodeSeason = episodes[watchedIndex].season
     val nextVideo = episodes.drop(watchedIndex + 1).firstOrNull { video ->
-        if (!isNextUpEpisodeAvailable(video.available)) {
-            logNextUpDecision(
-                "skip contentId=${progress.contentId} name=${progress.name} reason=unavailable " +
-                    "seed=${seedSeason}x${seedEpisode} next=${video.season}x${video.episode}"
-            )
-            return@firstOrNull false
-        }
         val releaseDate = parseEpisodeReleaseDate(video.released)
         val isSeasonRollover = video.season != watchedEpisodeSeason
         if (isSeasonRollover) {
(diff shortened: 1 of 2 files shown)


skoruppa:
The null-date -> unaired change makes sense, let's align mobile with that too (right now it treats missing date as aired from what I see).

On the available skip though - I'd rather not hard-block on that flag. Addons misuse it all the time and release date should be enough imho

ram130:
> The null-date -> unaired change makes sense, let's align mobile with that too (right now it treats missing date as aired from what I see).
> 
> 
> 
> On the available skip though - I'd rather not hard-block on that flag. Addons misuse it all the time and release date should be enough imho 

I agree on mobile.  I'll see if I can submit a pr after this merged to align it as well. 

I'll submit a new commit with the change when I get home.

ram130:
> The null-date -> unaired change makes sense, let's align mobile with that too (right now it treats missing date as aired from what I see).
> 
> On the available skip though - I'd rather not hard-block on that flag. Addons misuse it all the time and release date should be enough imho

Dropped the available check. 

Left the null-date -> unaired change as-is. 

Checked mobile — WatchingPolicies.shouldSurfaceNextEpisode has the same thing. The same-season branch ends in isReleasedBy -> isEpisodeReleaseAired(...) ?: true, while the season-rollover branch uses isExplicitlyReleasedBy -> ?: false. 

I'll open a PR there after this merges.

One question while I'm in there: mobile does gate on available (isReleasedBy returns false when available is false, plus the !available branch). Do you want that removed for consistency with this PR, or left as is?

ram130:
Confirmed mobile operating the same.  Don't think a pr is needed for mobile. This will bring it to parity.

PR #2630, opened 2026-07-14
fix(player): Shared progressive MP4 parallel session & buffer refcounting

halibiram:
## Summary

This PR fixes a critical crash (SIGSEGV) and bandwidth inefficiency issues during progressive MP4 playback when seeks/reopens occur, by sharing parallel chunk sessions across ExoPlayer reopens and introducing reference counting for chunk buffers.

Key changes:
1. **Shared MP4 Parallel Session**: Survives ExoPlayer seek reopens via a process-wide progressive chunk session keyed by `(chunkSize, index)` to prevent CDN/debrid URI churn from restarting downloads for the same range. In-flight downloads are kept across reopens, and shared futures are never cancelled when a single reader closes.
2. **Buffer Refcounting**: Replaces unsafe `Cleaner.clean` with a robust reference counter (`tryRetain` and `release`) with an epoch check on recycle. This ensures zero-copy `ByteBuffer` reads in `LoadTask` do not hit segment faults (SIGSEGV) after a long pause or resource recycling.
3. **Non-Faststart Optimization**: Prefetches head + tail moov islands as soon as the file length is known, prioritizing metadata TCP until a tail chunk completes, then unlocks playhead multi-ahead (capped at 4) on the active playhead corridor. Side scatter windows remain single-stream.
4. **Retry-After Header Handling & Clamping**: Parses HTTP `Retry-After` headers (supporting delta-seconds and RFC 1123 dates) for rate limits. If rate limits are triggered repeatedly (3 times in a 10s window), the session is temporarily clamped to a single connection.
5. **Progressive MP4 Session Mode**: Extends chunk session logic to progressive MP4 streams even when parallel connections are disabled, running a single-connection 8MB chunk session source to cache chunks across seeks.
6. **Connection Pool & Socket Isolation**: Allocates a separate `parallelPool` (size >= 32) in `PlayerMediaSourceFactory` and increases OkHttp pool size dynamically to isolate player range GET requests from default shared sockets and avoid socket starvation.
7. **Auth / Redirection Expiration**: Invalidates resolved URIs on HTTP 401/403/410 errors to trigger redirection re-probing on the next connection attempt.
8. **Subtitles & Fallback**: Bypasses parallel mode for subtitles (`nuvio_type=subtitle`) and falls back to a single upstream connection if range requests are not accepted.
9. **Memory Budget & Connection Tuning**: Adds MP4 session overhead when parallel is off to the Settings UI estimate.
[...]

PR #3102, opened 2026-08-20
fix(player): play non-faststart MP4s on the default path via session-owned chunks

### LINKED ISSUE #3100: [Bug]: Non-faststart MP4 (moov atom at end) fails to play or constantly rebuffers on default settings
### Pre-flight checks

### App version / commit

0.8.6 beta

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Ugoos AM9 Pro

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

### Bug description

Progressive MP4 files whose moov atom sits at the end of the file (non-faststart — common in web-DL releases) fail to start, or start and then rebuffer continuously, on the default playback path (Parallel Connections OFF). 

ExoPlayer recreates the progressive DataSource on every seek; a non-faststart MP4 forces the extractor to seek to the file tail to read moov before seeking back to the front for mdat, and the default single-connection path pays a full close→reopen per hop with nothing retained, so the scatter-read pattern these files force never stabilises. 

Delivery is healthy — ffprobe, curl, and AVFoundation-based players range-seek to the tail moov and play the same file fine, and the CDN honours range requests (HTTP 206). This affects all non-faststart MP4s regardless of source. With Parallel Connections ON the file is watchable; with it OFF (the default) it is not.

### Steps to reproduce

Install stock 0.8.6-beta and leave Parallel Connections OFF (the default).
Play a large non-faststart MP4 (moov atom at file tail) over HTTP from a debrid CDN.
Observe playback fails to reach a first frame, or starts and rebuffers within ~1 minute, worsening over time.

### Expected behavior

The file plays like any faststart MP4 — the player range-seeks to the tail moov, parses it, and plays, the way ffprobe/curl/Infuse do.

### Actual behavior

Playback hangs at buffering with no first frame, or plays briefly then rebuffers unwatchably.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

Test file is a non-faststart MP4, confirmed via ffprobe -v trace:

Atom order: ftyp (32 B) -> mdat (17.28 GiB) -> moov (20.9 MiB) at 99.88% into the 18.58 GB file.
Not fragmented (no sidx/moof) - the track index a player must read before decoding sits after the entire payload.

Delivery confirmed healthy (curl): CDN serves ~47 MB/s (~17x the file's 21.6 Mbps bitrate) and honours range requests (206 Partial Content + accept-ranges: bytes), so a seek-to-tail for the moov is possible at the network layer.

Cross-player: the same URL plays/parses in curl (linear read), ffprobe (range-seeks to tail moov, parses in seconds), and Infuse/AVFoundation (range-seeks, plays). It fails only in ExoPlayer-based players.

```shell

```

### Anything else? (optional)

_No response_

ysosrs123:
## Summary

Progressive MP4 files whose `moov` atom sits at the end of the file (non-faststart, common in web-DL releases) fail to play smoothly on the default playback path — they either never reach a first frame or rebuffer constantly. This PR makes them play by giving the chunk-download session a lifetime that survives ExoPlayer's per-seek data-source recreation, and by engaging that session for progressive MP4 even when parallel connections are off (pinned to a single connection and an 8 MB chunk, so request concurrency and retained memory stay at plain-path levels).

## PR type

- [x] Critical bug fix

## Why

A non-faststart MP4 has its `moov` atom (the track index a player must read before it can decode) at the tail of the file. To start playback the extractor must seek to the end for the `moov`, then back to the front for `mdat`, and it continues to scatter-read as it demuxes. ExoPlayer recreates the progressive `DataSource` on every seek, closing the previous one.

On the default single-connection path each of those recreations previously cancelled all in-flight downloads and re-probed the URL, retaining nothing across the seek. The scatter-read pattern these files force therefore never stabilises: the same ranges restart repeatedly and the buffer cannot fill, presenting as an unstartable stream or constant rebuffering.

Delivery is not the cause — `ffprobe`, `curl`, and AVFoundation-based players range-seek to the tail `moov` and play the same files fine, and the CDN honours range requests (HTTP 206). The fix is to make the downloaded chunks (completed *and* in-flight) belong to a companion-level session that survives the close/open boundary, so the seek back to the front reuses the tail read instead of discarding it, and to route progressive MP4 through that session on the default path where it previously used the plain source.

## Issue or approval

Fixes #3100

## Reproduction steps

1. Install a stock build and leave **Parallel Connections OFF** (the default).
2. Play a large non-faststart MP4 (moov atom at the file tail) over HTTP.
3. Observe playback fails to reach a first frame, or starts and then rebuffers continuously.

`ffprobe -v trace` on a representative file shows atom order `ftyp` → `mdat` → `moov`, with the `moov` at ~99.9% of the file.

## UI / behavior impact

- [x] No UI change

## Policy check

## Scope boundaries

This PR ports only the minimal mechanism that makes non-faststart MP4 playable on the default path, plus the small correctness fixes that mechanism needs:

- **Included:** session-owned chunk downloads that survive the per-seek data-source recreation; single-connection MP4 session mode for progressive MP4 when parallel connections are off; reads through a duplicated `ByteBuffer` on both read paths (shared chunks are read concurrently); rejection of short/EOF-truncated chunks and a zero-progress read guard; cancel-before-drop of failed chunk futures with an ownership-gated release; length-state consistency on the non-range fallback and on fresh reopen; request headers folded into the session identity; touch-LRU eviction bounded at cap+2 with position-aware victim selection that never evicts the imminent prefetch window. `MemoryBudget.BUFFER_OVERHEAD` is corrected `1 → 2` so the settings-screen memory estimate matches the data source's actual retained-chunk ceiling (with its unit test updated).
- **Not included:** no new settings, UI, or user-facing options. HLS/DASH, non-MP4 progressive, subtitle, and the plain single-connection path for other containers are unchanged. The parallel-on path gains the same session survival but is otherwise unchanged.
- **Deliberately held back for a separate follow-up:** buffer lifetime here is protected by a short touch-guard on eviction, which is adequate for the common case and has been in downstream production use. A structurally stronger approach — reference-counting the session buffers with epoch invalidation (as explored in #2630) — eliminates the free-during-read race by construction rather than by timing and is the better long-term fix, but it is a larger change that warrants its own soak testing and focused review. It is intentionally out of scope here to keep this the smallest change that fixes #3100, and is intended as a follow-up crediting @halibiram's design. The core mechanism in this PR is the same one reviewed in #2544.

## Testing

**Device:** Ugoos AM9 Pro (Amlogic, `arm64-v8a`), Android 14, internal player, default settings (Parallel Connections **off**).

**Build / unit tests (branch off current `dev`):**
- `./gradlew assembleFullDebug` → **BUILD SUCCESSFUL**.
- `./gradlew testFullDebugUnitTest` → 12 failures, but a control run on unmodified `dev` produced the **identical 12 failures** (same test names and line numbers, 863 total / 12 failed / 1 skipped both ways). None are in files this PR touches; they are pre-existing on `dev`. The one test this PR modifies (`MemoryBudgetTest`) passes.

**On-device, before/after — same file, same settings, same single connection.** Test file: a ~17.7 GB non-faststart MP4 (`moov` atom at ~99.9% of file), ~20 Mbps, well within single-connection bandwidth. Buffer-ahead and rebuffering sampled from the player over ~2 minutes of playback:

| | Rebuffer events | Total rebuffering | Buffering stalls | Buffer-ahead behaviour |
|---|---|---|---|---|
| **Stock (`dev`)** | 7 | ~62 s | 14 | Oscillates 1–20 s, repeatedly collapses; never stabilises |
| **This PR** | 0 | 0 s | 0 | Climbs to ~47 s ahead and holds flat for the whole run |

Because both runs use one connection at the same bitrate, the difference is not bandwidth — stock's default path cannot build a buffer because it pays the data-source reopen cost on every seek the non-faststart file forces; with the session-owned chunks it builds and holds a ~47 s buffer and plays without a single stall. Logs confirm the session engages (`MP4_SESSION engaged`), performs the tail `moov` read (reopen near the end of the file), then streams the front sequentially with each chunk fetched once.

The underlying mechanism (session survival + the review fixes above) has been running in a downstream fork build in daily use for several weeks; this PR rebases that proven mechanism onto current `dev` and adds the position-aware eviction fix.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Fixes #3100

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/ParallelRangeDataSource.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/ParallelRangeDataSource.kt
@@ -111,7 +111,48 @@ internal class ParallelRangeDataSource(
             return configuredDepth.coerceAtMost(rateLimitDepth).coerceAtLeast(1)
         }
 
-        private const val EVICTION_TOUCH_GUARD_MS = 2_000L
+        private const val TAIL_CHUNK_COUNT = 4L
+
+        internal fun isTailChunk(chunkIndex: Long, totalChunks: Long): Boolean {
+            return totalChunks > 0L && chunkIndex >= totalChunks - TAIL_CHUNK_COUNT
+        }
+
+        internal fun shouldMoveMainCursor(
+            lastReadChunkIndex: Long,
+            chunkIndex: Long,
+            prefetchWindow: Int,
+            sequentialOpen: Boolean,
+            currentChunkComplete: Boolean,
+            totalChunks: Long
+        ): Boolean {
+            if (isTailChunk(chunkIndex, totalChunks)) {
+                if (lastReadChunkIndex < 0L) return false
+                val delta = chunkIndex - lastReadChunkIndex
+                val distance = if (delta >= 0L) delta else -delta
+                return distance <= prefetchWindow.toLong()
+            }
+            if (lastReadChunkIndex < 0L) return true
+            val delta = chunkIndex - lastReadChunkIndex
+            val distance = if (delta >= 0L) delta else -delta
+            if (distance <= prefetchWindow.toLong()) return true
+            return sequentialOpen && currentChunkComplete
+        }
+
+        private const val EVICTION_TOUCH_GUARD_MS = 15_000L
+        private const val PLAYHEAD_BACK_CHUNKS = 2L
+        private const val MAX_PINNED_SIDE_CHUNKS = 2
+
+        internal fun isInPlayheadWindow(
+            readerIdx: Long,
+            chunkIndex: Long,
+            prefetchWindow: Int,
+            backChunks: Long = PLAYHEAD_BACK_CHUNKS
+        ): Boolean {
+            if (readerIdx < 0L) return false
+            val floor = (readerIdx - backChunks).coerceAtLeast(0L)
+            val ceil = readerIdx + prefetchWindow.toLong()
+            return chunkIndex in floor..ceil
+        }
         private const val RATE_LIMIT_MAX_BACKOFF_RETRIES = 3
         private const val RATE_LIMIT_BACKOFF_BASE_MS = 500L
         private const val RATE_LIMIT_BACKOFF_CYCLE_CAP_MS = 3_000L
@@ -145,10 +186,37 @@ internal class ParallelRangeDataSource(
             }
 
             @Volatile var lastReadChunkIndex: Long = -1L
+            val pinnedSideChunks: MutableSet<Long> = java.util.concurrent.ConcurrentHashMap.newKeySet()
 
-            fun noteRead(chunkIndex: Long) {
+            fun noteRead(
+                chunkIndex: Long,
+                sequentialOpen: Boolean,
+                currentChunkComplete: Boolean,
+                totalChunks: Long
+            ) {
                 touch(chunkIndex)
-                lastReadChunkIndex = chunkIndex
+                if (shouldMoveMainCursor(
+                        lastReadChunkIndex,
+                        chunkIndex,
+                        prefetchWindow,
+                        sequentialOpen,
+                        currentChunkComplete,
+                        totalChunks
+                    )
+                ) {
+                    lastReadChunkIndex = chunkIndex
+                    pinnedSideChunks.remove(chunkIndex)
+                } else {
+                    pinSideChunk(chunkIndex)
+                }
+            }
+
+            fun pinSideChunk(chunkIndex: Long) {
+                pinnedSideChunks.add(chunkIndex)
+                while (pinnedSideChunks.size > MAX_PINNED_SIDE_CHUNKS) {
+                    val drop = pinnedSideChunks.minByOrNull { lastTouch[it] ?: 0L } ?: break
+                    if (!pinnedSideChunks.remove(drop)) break
+                }
             }
 
             val rateLimitDepthCap = AtomicInteger(Int.MAX_VALUE)
@@ -234,6 +302,7 @@ internal class ParallelRangeDataSource(
         ) {
             val future = session.futures.remove(chunkIndex) ?: return
             session.lastTouch.remove(chunkIndex)
+            session.pinnedSideChunks.remove(chunkIndex)
             if (!future.cancel(true) && future.isDone && !future.isCancelled) {
                 try {
                     releaseSessionBuffer(future.get().buffer, session.chunkSize, poolCap)
@@ -254,6 +323,7 @@ internal class ParallelRangeDataSource(
             }
             session.futures.clear()
             session.lastTouch.clear()
+            session.pinnedSideChunks.clear()
         }
 
         private fun obtainSession(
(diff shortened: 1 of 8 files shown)


ysosrs123:
Heads-up: the PR Full Debug Build Comment workflow shows a red X, but it's the post-build comment step failing, not the build. It errors with HttpError: Resource not accessible by integration — the fork-PR GITHUB_TOKEN is read-only, so the workflow can't post the build-result comment back here. The pr-full-debug-build run itself succeeded; the debug APK is available as an artifact on that run for anyone wanting to test. Happy to adjust anything on the PR if needed.

skoruppa:
@halibiram want to check it out? ;)

halibiram:
@ysosrs123 yes, this works. If you don't mind, I need to apply a few adjustments for error cases, or I can tell you what needs to be fixed if you'd like to handle it

ysosrs123:
> @ysosrs123 yes, this works. If you don't mind, I need to apply a few adjustments for error cases, or I can tell you what needs to be fixed if you'd like to handle it

@halibiram is it the #2630 additions? I'm happy either way mate. If you have a lot of other work on your plate, just let me know what you need and I'll include those in a separate PR now. Otherwise, please feel free to adjust on your side.

ram130:
Looks good, I have a buffer engine patch that had included fixes for the main issue as well coming, and improve things all users reported, along with a patch for the VOD cache system I needed to update since it wasn't working. I'll review the code here and see what's up. I'll report back, hold on merging @skoruppa @halibiram

ram130:
Reviewed code, it aligns with a fix I had as well. This PR is good, only thing it needed some other fixes which I think @halibiram  is highlighting. 

> @ysosrs123 yes, this works. If you don't mind, I need to apply a few adjustments for error cases, or I can tell you what needs to be fixed if you'd like to handle it

@halibiram on the error-case adjustments, do they cover 429? assuming it's for the Retry-After that clamps to single-stream, to fix the observed intervals in the log pin at ~5s indefinitely.

per my testing, The MKV opened at 0, read the cues, seeked once to the play position, then streamed the rest of the run without reopening. The MP4 reopened 141 times across 44 offsets

data for anyone who needed it 
| | opens | failures | HTTP 429 | distinct positions |
|---|---|---|---|---|
| MP4, 27.9 GB, non-faststart | 141 | 64 | 63 | 44 |
| MKV, 30.3 GB | 4 | 1 | 0 | 4 |

Lastly, my VOD fix will also improve this issue since it will prevent a fresh HTTP request.

halibiram:
@ysosrs123 The core logic looks great. I think adding just the 429 / retry after handling from #2630 into this PR would be really good. It's just a small helper to clamp connections when a CDN rate limits. (correct me if I'm wrong)

@ram130 It's not in this PR yet, but if @ysosrs123 adds that small retry-after piece here, it should take care of the 429 clamping as well.

halibiram:
You don't necessarily have to follow that PR either, there might be a better solution as well

ysosrs123:
> You don't necessarily have to follow that PR either, there might be a better solution as well

@halibiram  -all understood. I'll update this PR and get you to take a look

ysosrs123:
Added the 429/503 handling as discussed — one commit on top: fix(player): adaptive 429/503 handling for session chunk downloads.

@ram130 to your question: yes, this covers the 429 case, and specifically the ~5s-indefinite pin. Root cause of that pin wasn't only the missing Retry-After — the per-chunk backoff ladder reset every time the outer retry machinery re-entered it, so against a hard limiter it hammered at a fixed period forever, header or no header. The fix: waits honour a server-stated Retry-After (both RFC 7231 forms, capped so a broken header can't stall a real-time pipeline), and without a header they escalate across episodes and decay with quiet. Prefetch depth adapts AIMD-style — halve on a rate-limit episode, +1 per quiet interval — so a session converges just under the provider's actual request budget instead of clamping to a single connection for the rest of playback. All state is per-session; providers that never rate-limit pay nothing.

The Retry-After parser and its unit tests are adopted from #2630 — credit @halibiram (co-authored on the commit).

Testing: assembleFullDebug green; unit suite 867 tests with the same 12 pre-existing dev failures, the four parser tests pass. On-device (Ugoos AM9 Pro): played sessions across three provider types — TorBox, Premiumize, and Emby — with the new path live; no false activations, no regressions, correctly dormant on sources that don't rate-limit. TorBox declined to actually 429 during testing because it's having DDoS issues again, so the live AIMD sequence wasn't captured on-device — your 429-reproducing setup is the definitive exercise for that, and the CI APK from this PR should be ideal for it.

skoruppa:
@halibiram @ram130 any objections?

ram130:
Good to go.

halibiram:
@ysosrs123 it's your turn to review :d I simulated a slow internet connection and made a few improvements, could you include these in the PR?

https://github.com/ysosrs123/NuvioTV-Fork/pull/1

ysosrs123:
> @ysosrs123 it's your turn to review :d I simulated a slow internet connection and made a few improvements, could you include these in the PR?
> 
> [ysosrs123#1](https://github.com/ysosrs123/NuvioTV-Fork/pull/1)

checking now mate. Thanks for simulating a slow connection, it completely slipped my mind, and is the main thing that will help people.

halibiram:
> @halibiram @ram130 any objections?

@ysosrs123  If you don't have anything else to add, I think it's ready to merge :)

ysosrs123:
> > @halibiram @ram130 any objections?
> 
> @ysosrs123 If you don't have anything else to add, I think it's ready to merge :)

Good to merge! Nice working with you both :)

I think this will help a bunch of people who report buffering but can't work out why

ram130:
It should on the mp4 side for sure.

ram130:
@ysosrs123 

Please double-check with my pr as I did some patches to the buffer system before you and on the same files you touch, updated memory budget, and fixed some stuff. 

#3130

ysosrs123:
> @ysosrs123
> 
> Please double-check with my pr as I did some patches to the buffer system before you and on the same files you touch, updated memory budget, and fixed some stuff.
> 
> #3130

Checked — I fetched your branch and ran a merge simulation against mine. Three of the four shared files auto-merge cleanly; the one conflict was a single hunk in MemoryBudget.kt where my tier-cap declaration abutted a line you rewrote. I've relocated that declaration (pure move, no semantic change) and re-ran the simulation in both merge orders: zero conflicts either way, and the merged result is the correct union — your budget floor and conversion ratios compose with my BUFFER_OVERHEAD fix and the low-RAM chunk cap.

Also gave the merged PlayerMediaSourceFactory a pass: your VOD cache wraps my chunk session as its upstream, which is complementary — your disk layer absorbs backward-seek re-reads, so a cached tail moov never re-hits the network on rewatch.

Net: both PRs land in the same build with nothing for either of us to resolve at merge time.

ram130:
Thanks.

Issue #22, opened 2026-02-07
Unable to Add Repository


fabears4us:
Unable to Add Repository in plugins section also how to remove any continue watching content....

Issue #11, opened 2026-02-07
Trakt integration


CrissZollo:
This should bring integration to trakt with syncing watching and watched content.

BLHRAM13:
Definitely looking forward to Trakt integration. Thanks for the great work so far.

tapframe:
Closed #82

PR #3107, opened 2026-08-20
fix(audio): pace bitstream passthrough on content time

### LINKED ISSUE #3086: [Bug]: Video stutters and loses A/V sync after around 1 hour of continuous playback
### Pre-flight checks

### App version / commit

0.8.6

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

Xiaomi TV Box Gen 3

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

### Bug description

When watching a movie using the internal player, playback runs completely smoothly until it reaches around the 1-hour mark (e.g., at 01:00:00, 02:00:00). At that exact point, the video suddenly begins to stutter, becomes laggy, and the audio and video abruptly lose synchronization.
Temporary Workaround:
Rewinding the video by 5 seconds instantly fixes the stutter and restores proper audio/video synchronization, but the issue reoccurs after another hour of continuous playback.

### Steps to reproduce

1. Start playing a long video or movie in the internal player.
2. Let it play continuously without pausing or seeking.
3. Observe playback around the 01:00:00 mark.

### Expected behavior

The movie should play smoothly until the end.

### Actual behavior

The video becomes laggy and the A/V lose sync after around a continuous hour of playback 

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

When the issue happened I pressed the report button to record the logs (NUV-371AA2). Note: it’s not always on exactly 1 hour, it usually happens around 40-80 minutes of CONTINUOUS playback

halibiram:
## Summary

Encoded HDMI passthrough (DTS, DTS-HD, AC-3, E-AC-3, AC-4, TrueHD) could let a byte-paced AudioTrack HAL sprint through VBR frames. The player clock then ran fast, video dropped frames to catch up, and A/V sync collapsed after tens of minutes of uninterrupted playback.

This branch now:

- Paces bitstream writes on content time (200 ms water level; 800 ms for RAW TrueHD, 200 ms when IEC-packed) and clamps `getCurrentPositionUs` to wall time and last written PTS. PCM decode is unchanged.
- Packs TrueHD into MAT / IEC 61937 when a compressed HBR track can be opened (`ENCODING_DOLBY_MAT`, or `ENCODING_IEC61937` after a background probe confirms it initializes). DTS-HD uses compressed RAW `ENCODING_DTS_HD` until that probe succeeds. Passthrough stays enabled; nothing is forced to PCM.
- Selects bitstream tracks on first `selectAllTracks` when the audio renderer is already alive, so playback does not start on AC-3/PCM and reselect DTS-HD seconds later.
- Caps RAW passthrough `AudioTrack` buffers per codec so the HAL does not allocate a 2+ MB DTS-HD buffer before the first sample is heard.

## PR type

- [x] Critical bug fix

## Why

#3086 reports that internal-player movies play smoothly until roughly 40–80 minutes of continuous playback, then video stutters, A/V sync is lost, and a 5-second rewind temporarily restores it until the next hour of uninterrupted play. The reporter reproduced this on Xiaomi TV Box Gen 3 (Android 14) in 0.8.6 and attached playback report NUV-371AA2.

That report is DTS-HD 7.1 passthrough on Amlogic (`audio/vnd.dts.hd`, 8ch, `DEVICE_ONLY`, zero audio underruns, accumulating dropped video frames). Lossless DTS-HD MA frame size grows with loudness. If the HAL advances time by bytes instead of access-unit duration, the media clock drifts until lipsync fails. The same class of clock applies to other encoded passthrough formats.

Separately, `ENCODING_IEC61937` `AudioTrack.Builder` can block the playback thread for several seconds on HALs that advertise the encoding and then reject the track. That delayed DTS-HD lock. HDMI audio encodings can also arrive after the first track, so the selector started on AC-3 and switched to DTS-HD later. Media3 sizes DTS-HD passthrough buffers from an 18 Mbps max rate (~2.25 MB), which delayed audible lock.

This is a mid-movie playback break and a delayed bitstream lock on the internal player, not a UI or feature change.

## Issue or approval

Fixes #3086

## Reproduction steps

1. On Android TV, play a long movie in the internal ExoPlayer with an encoded soundtrack (DTS-HD / DTS / TrueHD / AC-3 / E-AC-3 passthrough).
2. Let it run continuously with no pause and no seek.
3. Before this fix: around 40–80 minutes (often near 01:00:00) video stutters, A/V sync is lost. Rewinding ~5 seconds restores sync until the next uninterrupted hour. DTS-HD (and other bitstream tracks) could also take several seconds to lock after start because of an IEC61937 open attempt and/or an oversized AudioTrack buffer.
4. After this fix: bitstream passthrough remains active. Writes cannot run more than the content-time water level ahead of playing wall clock, and `getCurrentPositionUs` cannot run ahead of wall time or last written PTS. Continuous playback should keep A/V aligned past the one-hour mark. DTS-HD / AC-3 / E-AC-3 / TrueHD should lock on the compressed path without a multi-second IEC61937 stall, with per-codec RAW buffer caps (AC-3 12 KiB, E-AC-3 21 KiB, DTS 32 KiB, DTS-HD/TrueHD 120 KiB).
5. Confirm PCM / FFmpeg decode paths are unaffected (Bluetooth force-PCM, `audio/raw`). AVR codec badges stay on compressed encodings (`ENCODING_DTS_HD`, `ENCODING_AC3`, `ENCODING_DOLBY_TRUEHD` / `ENCODING_DOLBY_MAT`).

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

AudioTrack still receives encoded bitstream. Write-ahead, playhead, first-track selection, and RAW buffer size are corrected so video does not chase a sprinting HAL and bitstream lock is not delayed.

## Policy check

## Scope boundaries

- Limited to ExoPlayer encoded passthrough: content-time pacing, HBR MAT/IEC packing when the compressed track opens, first-selection of bitstream tracks, and per-codec RAW AudioTrack buffer caps.
- Does not add settings toggles or UI.
- Does not force PCM, disable passthrough, or change decoder priority.
- MPV is unchanged.
- AC-3 / E-AC-3 / AC-4 / DTS / DTS-HD / TrueHD stay on the bitstream path.
- PCM (`audio/raw`) is not paced and is not buffer-capped by the passthrough provider.
- `ENCODING_IEC61937` is never created on the playback thread; a background probe records whether it initializes, and DTS-HD uses compressed RAW until that probe succeeds.

## Testing

- `./gradlew :app:testFullDebugUnitTest --tests com.nuvio.tv.ui.screens.player.PassthroughWaterLevelPacerTest --tests com.nuvio.tv.ui.screens.player.BluetoothAudioRoutePolicyTest --tests com.nuvio.tv.ui.screens.player.iec.*`
- `PassthroughWaterLevelPacerTest`:
  - AC-3, E-AC-3, AC-4, DTS family, and TrueHD are treated as passthrough; PCM and AAC are not.
  - DTS-HD and AC-3 use the 200 ms water level; TrueHD uses 800 ms unless IEC-packed (then 200 ms).
  - Pre-start and playing sprints beyond the water level are rejected; realtime writes are accepted.
  - Playhead is capped at written PTS and wall time; lag is left alone.
  - Pause does not grow the allowed window; seek/discontinuity relatches PTS.
  - Sink integration: AC-3 `handleBuffer` rejects a pre-start sprint; PCM does not.
- `Iec61937PackerTest` / `TrueHdMatPackerTest` / `IecPassthroughAudioSinkTest`: MAT frame size and codes, IEC TrueHD preamble/byte-swap, DTS-HD 8ch burst 32768 bytes, TrueHD writes IEC bursts on a fake track, PCM is forwarded, factory-null falls back to the inner sink.
- Existing Bluetooth PCM/passthrough policy tests still pass.
- On-device (TCL / Realtek): DTS-HD opens on compressed `ENCODING_DTS_HD` (`HBR RAW mime=audio/vnd.dts.hd`) without an IEC61937 create on the playback thread; codec badge remains DTS.

## Screenshots / Video

Not a UI change

## Breaking changes

None

## Linked issues

Fixes #3086


Perza83:
Can't believe we're getting flawless truehd dolby branching audio on Nuvio 💙 Thank you @halibiram

ysosrs123:
@halibiram Nice to see this — the #3086 drift matches what I've seen from byte-paced passthrough clocks too, so content-time pacing feels like the right fix.

I shipped an app-side TrueHD MAT path in my fork recently and got burnt by two subtle issues (payload byte-swap, and play state being lost across a mid-start pipeline rebuild), so I checked this branch for both — all good here on both counts.

One thing worth a look from my experience: on the DOLBY_MAT path the frame keeps the 8-byte IEC preamble gap at the front of the buffer — might be worth confirming on hardware that the encoding expects that framing, if the test boxes happened to take the IEC route instead.

Happy to add Amlogic coverage — I've got an S905X5-J box into an eARC soundbar chain that's proven to lock app-packed TrueHD MAT. Planning to build this branch and test the passthrough formats, pause/seek, display switching, and a long run for the #3086 drift. Will post results for you.

Perza83:
@ysosrs123 Hey legend, I just wanted to say I love your fork and the optimisations you're doing with it. It's my daily goto streaming app atm and playback is flawless on my xiaomi 3rd gen 💙

ram130:
This is great, we can all build on top of this as well, I'll begin work on the auto mode. @ysosrs123 @halibiram

halibiram:
> This is great, we can all build on top of this as well, I'll begin work on the auto mode. @ysosrs123 @halibiram 

Since our work complements each other, I suggested to @ysosrs123 that we continue on this PR to test everything together. If you'd like to add anything as well, feel free to send a PR to my fork

ram130:
> > This is great, we can all build on top of this as well, I'll begin work on the auto mode. @ysosrs123 @halibiram 
> 
> Since our work complements each other, I suggested to @ysosrs123 that we continue on this PR to test everything together. If you'd like to add anything as well, feel free to send a PR to my fork

I talked with him.  Everything is looking great. Once apk is ready I'll confirm no issue on all my devices.

ram130:
Adding audio portion of HUD work, and reviewing optimizations and code for the audio engine.

ram130:
PR commit submitted to @halibiram , based on the optimizations and code review, we have some issues to work on, especially with extra data in the audio pipeline. But will save full review report until we are near ready. 

few call outs I need to make right now

- I notice a delay with audio after a few seconds of a playback. Delay seem to be about 1-2s behind. Regardless of format. 
- DV7 file + True HD seems to be giving me black screen, audio only. 
- DTS-HD is working on fire tv, but same as above, blank screen, audio only

Note: DTS-HD didn't work before on the Fire TV due to amazon restrictions.

halibiram:
> PR commit submitted to @halibiram , based on the optimizations and code review, we have some issues to work on, especially with extra data in the audio pipeline. But will save full review report until we are near ready.
> 
> few call outs I need to make right now
> 
> * I notice a delay with audio after a few seconds of a playback. Delay seem to be about 1-2s behind. Regardless of format.
> * DV7 file + True HD seems to be giving me black screen, audio only.
> * DTS-HD is working on fire tv, but same as above, blank screen, oudio only
> 
> Note: DTS-HD didn't work before on the Fire TV due to amazon restrictions.

merged :). Was tunneled playback active? A conflict with that might be causing this

ram130:
> > PR commit submitted to @halibiram , based on the optimizations and code review, we have some issues to work on, especially with extra data in the audio pipeline. But will save full review report until we are near ready.
> 
> > 
> 
> > few call outs I need to make right now
> 
> > 
> 
> > * I notice a delay with audio after a few seconds of a playback. Delay seem to be about 1-2s behind. Regardless of format.
> 
> > * DV7 file + True HD seems to be giving me black screen, audio only.
> 
> > * DTS-HD is working on fire tv, but same as above, blank screen, oudio only
> 
> > 
> 
> > Note: DTS-HD didn't work before on the Fire TV due to amazon restrictions.
> 
> 
> 
> merged :). Was tunneled playback active? A conflict with that might be causing this

Yep.  Come to think of it the log was flooding with a lot of info when I went to check. I wonder if it's related to the mp4 issue. Hmm. I'll try with it off and report back later.

ysosrs123:
Was this on a Fire Stick 4k 2nd Gen? Or Fire Stick 4 Max? or another kind?
If you can let me know also what file you played i'll test as well on my Amlogic devices

ram130:
> > > > PR commit submitted to @halibiram , based on the optimizations and code review, we have some issues to work on, especially with extra data in the audio pipeline. But will save full review report until we are near ready.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > few call outs I need to make right now
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > * I notice a delay with audio after a few seconds of a playback. Delay seem to be about 1-2s behind. Regardless of format.
> 
> > > 
> 
> > > 
> 
> > > > * DV7 file + True HD seems to be giving me black screen, audio only.
> 
> > > 
> 
> > > 
> 
> > > > * DTS-HD is working on fire tv, but same as above, blank screen, oudio only
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > Note: DTS-HD didn't work before on the Fire TV due to amazon restrictions.
> 
> > > 
> 
> > > 
> 
> > > merged :). Was tunneled playback active? A conflict with that might be causing this
> 
> > 
> 
> > Yep. Come to think of it the log was flooding with a lot of info when I went to check. I wonder if it's related to the mp4 issue. Hmm. I'll try with it off and report back later.
> 
> 
> 
> Was this on a Fire Stick 4k 2nd Gen? Or Fire Stick 4 Max? or another kind?
> 
> If you can let me know also what file you played i'll test as well on my Amlogic devices

Correct.  Fire TV 4K Max 2nd gen. Honestly any content with HD audio should trigger the issue easily I've noticed.

ysosrs123:
Thanks. I did 11 test runs w/captures. 
I'll put the findings and details in the fork PR comments so i don't flood this one - https://github.com/halibiram/NuvioTV/pull/1#issuecomment-5493024218

halibiram:
> I'll put the findings and details in the fork PR comments so i don't flood this one - [halibiram#1 (comment)](https://github.com/halibiram/NuvioTV/pull/1#issuecomment-5493024218)

Looks like we have some serious issues... >< Is it better to separate the audio track for tunneled playback right now, or should we try to fix it?

ysosrs123:
> > I'll put the findings and details in the fork PR comments so i don't flood this one - [halibiram#1 (comment)](https://github.com/halibiram/NuvioTV/pull/1#issuecomment-5493024218)
> 
> Looks like we have some serious issues... >< Is it better to separate the audio track for tunneled playback right now, or should we try to fix it?

I don’t think they’re serious issues. If this doesn’t need to be merged urgently, let’s try to fix it. I’m happy to work on it tomorrow if you want.

ram130:
Will test the new version in a few hours after work.  Otherwise I may also have a fix for the blank screen issue as well. Will report back.

ysosrs123:
> Will test the new version in a few hours after work. Otherwise I may also have a fix for the blank screen issue as well. Will report back.

I think I've sorted it. But only tested on amlogic

https://github.com/halibiram/NuvioTV/pulls

ram130:
> > Will test the new version in a few hours after work. Otherwise I may also have a fix for the blank screen issue as well. Will report back.
> 
> 
> 
> I think I've sorted it. But only tested on amlogic
> 
> 
> 
> https://github.com/halibiram/NuvioTV/pulls

Curious, didn't @halibiram recent commits and apk resolve it?

Either way I'll test it and if you have an apk for your pr send it here so I can test after.

ysosrs123:
> Curious, didn't @halibiram recent commits and apk resolve it?
> 
> Either way I'll test it and if you have an apk for your pr send it here so I can test after.

It didn't resolve for my device (Amlogic). It worked fine with tunneled playback OFF, but with it ON, black screen and no movement, but audio came through.  Let me know how that apk works on your firestick. Would be good to know if it works OK on mediatek SoC.

Here's the apks that include my PR's #4 and #5 
https://cloud.nordlocker.com/shares/unlock/a0011ce2-98d4-435e-912c-bf238a09d1e0#XNRFoJAdJASl74DBhbYWZYeljFQcCY5EdAHxsruYiEw

ram130:
> > Curious, didn't @halibiram recent commits and apk resolve it?
> > Either way I'll test it and if you have an apk for your pr send it here so I can test after.
> 
> It didn't resolve for my device (Amlogic). It worked fine with tunneled playback OFF, but with it ON, black screen and no movement, but audio came through. Let me know how that apk works on your firestick. Would be good to know if it works OK on mediatek SoC.
> 
> Here's the apks that include my PR's #4 and #5 https://cloud.nordlocker.com/shares/unlock/a0011ce2-98d4-435e-912c-bf238a09d1e0#XNRFoJAdJASl74DBhbYWZYeljFQcCY5EdAHxsruYiEw

Two builds tested, both with tunnelling off and on, switching between AC3, DTS-HD and TrueHD.

**Build 1, halibiram's latest commits (c9492d65a).** Tunnelling on gave static on every format and no
video. Tunnelling off played correctly but stuttered and went unresponsive after a couple of minutes.

**Build 2, the same plus #4 and #5.** The static is gone and playback works, tunnelling on or off,
apart from the items below.

---

**The static was the hw_av_sync binding.** 683d577f set `FLAG_HW_AV_SYNC` on the track but
`IecAudioTrack.write()` still wrote raw payload with no 16 byte AV sync header, so the HAL read the
first bytes of the MAT frame as that header and mis-framed everything after it. #4 reverts the
binding and the symptom goes with it. `hwAvSync=true` no longer appears and the IEC path still
works: 10 track opens across TrueHD and DTS-HD, `IEC61937 probe: usable`.

**The freezing is unchanged by #4 and #5, and is not a tunnelling bug.** Normalised for session
length:

| build | span | large object space freed | rate |
| --- | --- | --- | --- |
| halibiram tip | 6m56s | 819 MB across 47 GCs | 1.97 MB/s |
| with #4 and #5 | 16m25s | 2009 MB across 115 GCs | 2.04 MB/s |

Identical within noise, which is what the code says should happen since neither PR touches the
packers. Both runs ended in two ANRs each, same signature:

```
64% com.nuviodebug.com: 18% user + 45% kernel / faults: 23833 minor 66903 major
60% 127/kswapd0: 0% user + 60% kernel
Load: 17.83 / 14.96 / 10.62
```

No `FATAL EXCEPTION` and no `OutOfMemoryError` in either log. The pressure is large object space
rather than the java heap, which is why nothing throws. `kswapd0` pinned and 66903 major faults means
the main thread is waiting on page faults, so input dispatch times out at 5s and the app is killed.

Large object space is where allocations over roughly 12 KB go. `TrueHdMatPacker.writeHeader()`
allocates a fresh 61440 byte array per MAT frame, `Iec61937Packer.packTrueHd` does
`matFrame.copyOf()` for another 61440, and `packDtsHd` allocates up to 32 KB per access unit. Those
sizes are compile time constants, so they can be reused buffers rather than per frame allocations.
This looks like the cause of the freezes and needs fixing independently of the tunnel work.

**#4's watchdog did not fire under tunnelling here.** No `TUNNEL_AV_SYNC` line of either kind, while
tunnelling was definitely active (`open() set hdmi tunnelMode = 1`, `TUNNELED video`,
`hdmi_dolby_tunnel_out`). On build 1 the frame count version fired correctly twice,
`zero tunneled video frames after 8000ms`, disabled tunnelling for the stream and rebuilt into
working audio and video. So the frame count check is not universally blind, it is blind on the Ugoos
and works here. Worth keeping in mind before replacing it outright.

**Underruns are HAL side, not media3.** 14 of them, all
`alsaUnderFlowCheck() Underrun Happen: resetCase DSwitch`, each within about a second of an
`opened IEC61937`. They are device switch transients on IEC track teardown and reopen, so every
format switch produces one. Media3's own `onAudioUnderrun` fired zero times across the whole run.
Not buffer starvation, but every format change currently costs an audible one.

**#5 is untested from my side.** `DV_RUNTIME_CHECK_FALLBACK` appears zero times in either log, so the
misfire it targets never triggered here. Also no `less than 500ms of buffered media data` in either
run, so that part looks specific to your host rather than shared.

Otherwise for the most part when every thing works, audio and video is going fine with both tunneled on or off, with under runs every few mins. 

EDIT: Forgot to add, AC3 is being transcoded as PCM rather than DD+/DD 5.1/7.1

ysosrs123:
@ram130 / @halibiram 

New build at v0.9.0 and with PR 6,7,8,9,10 included -

https://www.dropbox.com/scl/fo/6xmuytvhdtvpu1zg5du15/AHna15f1EbMgGsACRjlMGsI?rlkey=jmpszhsr7j97bqp3ogiq8wv2f&st=fw22xovu&dl=0

ram130:
Thanks. Will test today.  Also need to submit a pr for some additional options. I'll report back later.

ram130:
Testing now.

ram130:
One call out from the latest commits, .idea/deploymentTargetSelector.xml is per-developer IDE state, like workspace.xml which is already ignored. It's currently tracked with a hardcoded emulator path (/Users/amarjitsingh/.android/avd/Television_1080p.avd), so every branch switch resets my run target to an emulator that doesn't exist on my machine, and Android Studio boots a default AVD instead of the connected device. Suggest adding it to .gitignore and git rm --cached it.

halibiram:
> One call out from the latest commits, .idea/deploymentTargetSelector.xml is per-developer IDE state, like workspace.xml which is already ignored. It's currently tracked with a hardcoded emulator path (/Users/amarjitsingh/.android/avd/Television_1080p.avd), so every branch switch resets my run target to an emulator that doesn't exist on my machine, and Android Studio boots a default AVD instead of the connected device. Suggest adding it to .gitignore and git rm --cached it.

It was already in repo, not added in this PR

ram130:
Tested the latest branch on a Fire TV 4K Max, native on and off, tunneling on, DTS-HD and TrueHD
remuxes plus an AC-3 file, 30 min.

The packer buffer reuse worked. Large object space turnover, normalized per second of session:

| build | LOS freed | rate |
| --- | --- | --- |
| before #4/#5 | 819 MB / 47 GCs | 1.97 MB/s |
| with #4/#5 | 2009 MB / 115 GCs | 2.04 MB/s |
| this build | 1284 MB / 34 GCs | 0.73 MB/s |

Down 64 percent, and the GCs carrying large objects went from 115 to 34 in a run nearly twice as
long. One ANR instead of two, and the paging pressure behind them is largely gone: major faults
66903 to 12086, kswapd0 60 percent to 20 percent.

The bottleneck moved from memory to CPU. The ANR profile inverted, 18 percent user / 45 percent
kernel before, 32 percent user / 16 percent kernel now. `Iec61937Packer.swapEndian16` is still a
byte at a time swap over 61440 bytes per MAT frame, about 30k iterations per frame on the audio
thread, and `packDtsHdInto` now adds an `out.fill(0)` per burst. Allocation is fixed, the per frame work is not. `ByteBuffer.asShortBuffer()` or writing the payload pre-swapped would remove it.

Caveat on that ANR: `com.amazon.whad` was at 62 percent (58 user), above Nuvio's 48 percent, and the
Amazon background stack summed to roughly 103 percent with load at 16.43. My device was very much hotter after a few minutes during to the higher cpu activity. 

AC-3 comes out as PCM 
E-AC3 passed through fine as `audio/eac3 channelCount=6`. Media3 logs `ac3.decoder, supported via passthrough: true`, so the chain accepts it

Still seeing dropouts, and they are not underruns. The HUD underrun counter stayed at 0 across every test, and the whole log has one `Dropped 10 frames`. So whatever the audio dropouts are, they are not buffer starvation and not mass frame drops. The same files on the public build on the same device do not do it.

No `TUNNEL_AV_SYNC` line of either kind with tunneling on, same as my last run.

ram130:
Looking into the dropouts now. Will report back once I have something concrete.

Separately I have a small follow up to [#7](https://github.com/NuvioMedia/NuvioTV/pull/3107/changes/d4234acbb8d0fd80925fda7d99b814ea4646a667) coming for the byte swap left in the packers, which is where the CPU went after the allocations were fixed.

ram130:
testing new commits

ram130:
Follow up after the new commits. The audio rejection and IEC fallback fixes sorted what I was
hitting before, both TrueHD and DTS-HD are selecting cleanly now. Great improvements to playback as well.

Reduced frame drops, only saw 1-2 during a longer playback of 20min. I need to test again to be
sure. The audio dropouts now line up with those frame drops, I only heard one at the moment a frame
dropped, so it looks like a single stall hitting both rather than two separate problems.

The new iec_health line may explain it. Across one 20 minute playback, tunnelling on, TrueHD:

```
iec_health mode=TRUEHD payload=IEC_BURST underruns=0 ... stalls=0
... stalls=611 ... stalls=3606 ... stalls=7424 ... stalls=10668
```
Switching DTS-HD to TrueHD  mid playback I saw a random stall
for a while before recovering.

Stalls climb continuously while underruns stay at 0 for the whole run. Switching again didn't reproduce the same issue, so seems random for now.

Profiled where the playback thread CPU goes, tunnelling both off and on. Same file per group,
`top -H` on the app.

Tunnelling off:

| audio | ExoPlayer:Playback |
| --- | --- |
| TrueHD / DTS-HD, no DV | 33-37% |
| TrueHD / DTS-HD, DV7 + libdovi | 30-34% |
| AC-3 | 9-10% |

Tunnelling on:

| audio | ExoPlayer:Playback |
| --- | --- |
| TrueHD, DV profile 8 | 24-28% |
| TrueHD, DV profile 7 + libdovi | 27-33% |
| AC-3 | 8-10% |

Dolby Vision makes no difference either way, profile 7 with conversion sits on top of profile 8. The
audio track is what moves it, tunnelling on or off. In an earlier build where the sink fell back to
RAW and the IEC packer never ran, TrueHD sat at 10% as well, so the cost only appears when IEC
packing actually executes.

That matters because the playback thread also releases video frames, so roughly 15 to 25 points
depending on tunnelling is a plausible cause of the frame drops, and public has no IEC packing at
all.

I have not profiled inside the packer, so I cannot say which part it is: MAT assembly, the burst
copies, the per burst fill, or what is left of the swap after [#11](https://github.com/NuvioMedia/NuvioTV/pull/3107/changes/573d8e67c6fe64e451abd205e8ed23165ef66079).

One question on the AC-3 path. In `PlayerRuntimeControllerInitialization.kt` around line 814,
`isForcePassthroughActive` feeds `EXTENSION_RENDERER_MODE_PREFER`, so enabling force optical
passthrough prefers the FFmpeg decoder. Is that deliberate? I am still seeing AC-3 decode rather
than pass through and I am trying to work out whether that is my settings or the code.

RES during playback is 289-340M against 151M on my build from two days ago, and this device has
around 40MB free. 

In terms of the frame drops, a question came up, are the IEC track sizing and the 200 ms water level deliberate given the stall counts, or is the buffer simply too small for this device? Will look into this further.

ysosrs123:
Thanks very much @ram130 
Did your test build include the 2x unmerged PRs in halil’s repo?

ram130:
> Thanks very much @ram130 Did your test build include the 2x unmerged PRs in halil’s repo?

no, just what has been committed already. I'll do another test with those tonight.

KevinMeinon:
Another data point for this clock, with a different symptom than #3086.

Setup: SHIELD TV Pro 2019, Nuvio 0.9.2, internal player on ExoPlayer, bitstream passthrough to a Denon AVR-X4500H over HDMI (Ethernet, no Bluetooth in the chain).

UHD BluRay REMUX, HEVC 4K HDR10, 51.8 GB, with both a TrueHD Atmos 7.1 track and a separate AC-3 5.1 track. The title ships both a theatrical and an extended cut on the disc, in case the playlist structure matters.

What happens, same file, only the audio track changed:

- TrueHD Atmos 7.1: audio goes silent for about 1 to 2 seconds, then returns out of sync with the video. Seeking back re-anchors it. Video never stutters and there is no rebuffering, so this is not the player waiting on data.
- AC-3 5.1: plays clean for the whole film. No dropout, no drift, no seek needed.

Same container, same source, same HDMI chain, so the only variable is the encoded passthrough path. If a byte-paced AudioTrack clock can also overshoot in short bursts instead of only after 40 to 80 minutes, that would fit what I see. TrueHD has by far the largest frame size of the encodings you cap here, and it is the one that fails.

Happy to test a build from this branch and re-run both tracks on the same file. I can also check whether the dropouts hit the same timestamps on a rewatch, if that is useful for narrowing it down.

ysosrs123:
@KevinMeinon thanks for the feedback.
Have you tried this build yet - https://discord.com/channels/1379902184207941732/1543735887916175400/1548048915801051156

ram130:
> Another data point for this clock, with a different symptom than #3086.
> 
> Setup: SHIELD TV Pro 2019, Nuvio 0.9.2, internal player on ExoPlayer, bitstream passthrough to a Denon AVR-X4500H over HDMI (Ethernet, no Bluetooth in the chain).
> 
> File: UHD BluRay REMUX, HEVC 4K HDR10, 51.8 GB, with both a TrueHD Atmos 7.1 track and a separate AC-3 5.1 track. The title ships both a theatrical and an extended cut on the disc, in case the playlist structure matters.
> 
> What happens, same file, only the audio track changed:
> 
> - TrueHD Atmos 7.1: audio goes silent for about 1 to 2 seconds, then returns out of sync with the video. Seeking back re-anchors it. Video never stutters and there is no rebuffering, so this is not the player waiting on data.
> - AC-3 5.1: plays clean for the whole film. No dropout, no drift, no seek needed.
> 
> Same container, same source, same HDMI chain, so the only variable is the encoded passthrough path. If a byte-paced AudioTrack clock can also overshoot in short bursts instead of only after 40 to 80 minutes, that would fit what I see. TrueHD has by far the largest frame size of the encodings you cap here, and it is the one that fails.
> 
> Happy to test a build from this branch and re-run both tracks on the same file. I can also check whether the dropouts hit the same timestamps on a rewatch, if that is useful for narrowing it down.
> 

Try the new build. Should fix this.

ram130:
Retesting.

halibiram:
> Retesting.

Could you test this? I recently received feedback that TrueHD is not working on the FireStick
https://github.com/halibiram/NuvioTV/pull/22

ram130:
> > Retesting.
> 
> 
> 
> Could you test this? I recently received feedback that TrueHD is not working on the FireStick
> 
> https://github.com/halibiram/NuvioTV/pull/22

Ok I'll do both this and it.

ram130:
> > Retesting.
> 
> Could you test this? I recently received feedback that TrueHD is not working on the FireStick [halibiram#22](https://github.com/halibiram/NuvioTV/pull/22)

Tested #22 on a Fire TV Stick 4K Max 2nd gen, against the branch without it as a control. 

With #22 and tunnelling on it is noise on every HBR format, TrueHD, Atmos TrueHD and DTS-HD
alike. The HAL downgrades the stream to PCM as soon as FLAG_HW_AV_SYNC is set. Every track
open logs the pair about 100ms apart:

  audio_output: format: iec61937, sample_rate: 192000, channels: 8
  audio_output: format: pcm-16,   sample_rate: 192000, channels: 8

and it lands on hdmi_pcm_tunnel instead of hdmi_raw:

  [hdmi_pcm_tunnel] +setParameters_ext(): hw_av_sync=1

The app side is unchanged, audio_input stays format 13 (IEC61937), so the MAT bursts are
clocked out as raw samples. My AVR confirms it, reporting PCM 2 channel. The playback head
agrees. Over one five second window on hdmi_pcm_tunnel the head moved 241k frames while the
write pointer moved 964k, a quarter of the rate. Eight channels packed, two consumed. On
hdmi_raw the two track each other.

Also 27 of these in that session, against 2 with tunnelling off:

  processHeaderBlock() InfoType 1, Header.kAttri = NONE_HW_SYNC while no reference

IecAudioTrack.write falls back to the untimestamped overload when timestampNs < 0, which would
put untimestamped blocks on a hw-sync stream.

The control is the same branch without #22, tunnelling on and off. Seven track opens, every one
iec61937, zero hdmi_pcm_tunnel, no downgrade on either setting. So the PCM downgrade does not
reproduce without this PR.

With #22 and tunnelling off everything is clean as well, so it looks specific to the hw_av_sync
path.

Separately, the buffer raise is working well. With tunnelling off, stalls reached 2 over a 12
minute session, against 1319 in 35 seconds before it.

One thing not specific to this PR: frame drops. 15 over one file. The public build on the same
files does not drop them, so something in this path is costing frame timing.
audio-beforePR22_framedrops
audio-beforePR22

EDIT: and to be clear, with tunneled off or on without PR22, Atmos, Atmos TrueHD all output correctly on my Fire TV. I think that user just needs to reboot.

image

ram130:
Testing again on Saturday.

PR #3128, opened 2026-08-21
Fix trailer preview falling back to low resolution

### LINKED ISSUE #3127: Trailer preview falls back to low resolution when the top adaptive stream can't be probed
### Pre-flight checks

### App version / commit

0.8.7-beta (also reproduces on latest dev, commit 57a7e21)

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

iffalcon android tv

### Android version

android 11

### Player mode

Internal player

### Area (tag)

### Bug description

The in-app YouTube extractor (InAppYouTubeExtractor.kt) probes only the single top-ranked adaptive video stream. If that one CDN URL is unreachable (403/timeout — which happens routinely, e.g. when YouTube changes client attestation requirements as in [#3079](https://github.com/NuvioMedia/NuvioTV/issues/3079)), the extractor falls straight through to the HLS manifest or progressive stream, which is often far below the resolution the trailer actually has on YouTube. Separately, TrailerService.rankTmdbVideoCandidates ranked TMDB's official flag above upload resolution, so an official 480p upload could beat an unofficial 1080p/4K one for the same title.

Verified live: for a recent trending title, YouTube returned 45 adaptive video streams up to 2072p, but the extractor's single-candidate probe failed and the progressive fallback it would have used was only 346p.

### Steps to reproduce

Focus any title on the home screen whose trailer's top-ranked adaptive stream happens to be unreachable from your network (this is intermittent per-CDN-node, not tied to specific titles).
Watch the trailer preview autoplay.
Observe the resolution is visibly low compared to the trailer as viewed directly on YouTube.

### Expected behavior

Trailer preview plays the highest-resolution stream YouTube has for that trailer.

### Actual behavior

Trailer preview plays a low-resolution (often ~360-480p) progressive/HLS fallback stream, even when the same trailer has 1080p/4K adaptive streams available on YouTube.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

I can fix this. this is an additional fix for the resolution fallback

RahulRex0:
## Summary
Two related trailer-quality defects fixed as two small, focused commits:
- The YouTube extractor only probed the single top-ranked adaptive video
  stream before falling back to HLS/progressive (~360-480p).
- TMDB trailer-candidate ranking placed the `official` flag above upload
  resolution, so a low-res official upload could beat a high-res one.

## PR type
- [x] Critical bug fix

## Why
Trailer previews were playing at noticeably lower quality than the same
trailer on YouTube, even when high-resolution streams were available.

## Issue or approval
Fixes #3127

## Reproduction steps
See linked issue.

## UI / behavior impact

## Policy check

## Scope boundaries
Does not touch trailer audio-track or language selection — those are a
separate, follow-up fix.

## Testing
- `./gradlew :app:testFullDebugUnitTest --tests "com.nuvio.tv.data.trailer.*"` — all passing, including new tests for both fixes.
- `./gradlew :app:assembleFullDebug` — builds successfully.
- Manually verified on-device: trailer preview now plays the ranked-highest reachable adaptive stream instead of falling back after a single failed probe.

## Screenshots / Video
Not a UI change.

## Breaking changes
None.

## Linked issues
Fixes #3127

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/trailer/InAppYouTubeExtractor.kt
+++ b/app/src/main/java/com/nuvio/tv/data/trailer/InAppYouTubeExtractor.kt
@@ -26,6 +26,7 @@ import javax.inject.Singleton
 
 private const val TAG = "InAppYouTubeExtractor"
 private const val EXTRACTOR_TIMEOUT_MS = 30_000L
+private const val MAX_ADAPTIVE_VIDEO_PROBES = 4
 private const val DEFAULT_USER_AGENT =
     "Mozilla/5.0 (Linux; Android 12; Android TV) AppleWebKit/537.36 " +
         "(KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36"
@@ -438,10 +439,18 @@ class InAppYouTubeExtractor @Inject constructor() {
         val bestProgressive = sortCandidates(progressive).firstOrNull()
         val bestVideo = pickBestForClient(adaptiveVideo, PREFERRED_SEPARATE_CLIENT)
         val bestAudio = pickBestForClient(adaptiveAudio, PREFERRED_SEPARATE_CLIENT)
+        // Preserve the preferred-client first choice, then try globally ranked fallbacks.
+        val adaptiveVideoUrls = listOfNotNull(bestVideo?.url)
+            .plus(sortCandidates(adaptiveVideo).map { it.url })
+            .distinct()
 
         // Try adaptive video + audio first (best quality, separate streams)
         kotlinx.coroutines.yield()
-        val resolvedVideo = bestVideo?.url?.let { resolveReachableUrl(it) }
+        val resolvedVideo = firstResolvedUrl(
+            candidates = adaptiveVideoUrls,
+            maxAttempts = MAX_ADAPTIVE_VIDEO_PROBES,
+            resolve = ::resolveReachableUrl
+        )
         val resolvedAudio = if (resolvedVideo != null) bestAudio?.url?.let { resolveReachableUrl(it) } else null
 
         if (resolvedVideo != null) {
@@ -829,6 +838,17 @@ class InAppYouTubeExtractor @Inject constructor() {
         }
         return headers.build()
     }
+}
+
+internal suspend fun firstResolvedUrl(
+    candidates: Iterable<String>,
+    maxAttempts: Int,
+    resolve: suspend (String) -> String?
+): String? {
+    for (candidate in candidates.take(maxAttempts)) {
+        resolve(candidate)?.let { return it }
+    }
+    return null
 }
 
 private data class RequestResponse(
--- a/app/src/main/java/com/nuvio/tv/data/trailer/TrailerService.kt
+++ b/app/src/main/java/com/nuvio/tv/data/trailer/TrailerService.kt
@@ -498,8 +498,8 @@ internal fun rankTmdbVideoCandidates(results: List<TmdbVideoResult>): List<TmdbV
         }
         .sortedWith(
             compareBy<TmdbVideoResult> { videoTypePriority(it.type) }
-                .thenBy { if (it.official == true) 0 else 1 }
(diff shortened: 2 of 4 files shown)


skoruppa:
Right now only animated movies (like Toy Story) always fail to low quality and with your change I don't see any improvement.  It only resulted in longer loading before finally failing back to low quality HLS. Do you have any examples where your changes helped?

RahulRex0:
Fair challenge, and I don't have a confirmed example. Two different things are in this PR,
worth separating:

1. Retrying different CDN mirror nodes for the same quality stream -
   that already existed before this PR, not something I touched.
2. Trying a different quality/itag entirely when the top one's mirrors
   all fail - that's what I actually added, and I don't have a verified
   case of that specific mechanism paying off. It's a plausible
   hypothesis, but I can't back it with real evidence right now.

Given that, I think the honest move is either: (a) I add temporary
logging that flags when a non-top candidate succeeds after the top one
failed, so real usage data settles this, or (b) I drop the multi-itag
retry from this PR entirely until there's evidence it does anything,
keeping just the resolution/official ranking fix. which
would you prefer?

kayochiaradia:
Reading this while looking at the same code, there's a coupling worth flagging before you decide between (a) and (b): **the retry is applied to video but not to audio, and those two are load-bearing together.**

On `dev`, adaptive video and audio are collected as separate single-track candidates (`InAppYouTubeExtractor.kt:367` and `:390`, split by `hasVideo` / `hasAudio` at `:355-356`), and the audio track is probed exactly once:

```kotlin
val resolvedAudio = if (resolvedVideo != null) bestAudio?.url?.let { resolveReachableUrl(it) } else null

if (resolvedVideo != null) {
    return TrailerPlaybackSource(videoUrl = resolvedVideo, audioUrl = resolvedAudio)
}
```

So `resolvedVideo != null` with `resolvedAudio == null` returns an adaptive video URL with no audio. `TrailerPlayer` only merges a second source when the audio URL is non-blank (`TrailerPlayer.kt:110-116`):

```kotlin
if (!trailerAudioUrl.isNullOrBlank()) {
    player.setMediaSource(MergingMediaSource(videoSource, audioSource))
} else {
    player.setMediaItem(MediaItem.fromUri(trailerUrl))
}
```

which plays the video-only stream on its own — a silent trailer.

That sharp edge already exists on `dev`; the reason it matters here is that this PR changes how often it is reached. Today `resolvedVideo` is non-null only when the single top candidate resolves. With up to four probes it becomes non-null more often, including in cases that previously fell through to the HLS manifest. Since the audio probe is unchanged, some of the newly-successful adaptive attempts will be video-with-failed-audio, and those return silent playback where `dev` would have returned muxed HLS with sound. Trading 1080p-with-audio for 1440p-silent is a bad trade, and it would show up as "no sound on some trailers" rather than as a resolution regression, so it is easy to misattribute.

If you keep the multi-itag retry, the fix is small: give `adaptiveAudio` the same ranked fallback, and only prefer the adaptive pair when both sides resolved — otherwise fall to HLS/progressive, keeping a silent adaptive stream as a last resort so nothing that plays today stops playing. That is a pure ordering decision and can be unit-tested without touching the network.

On the (a)/(b) question itself I can't settle whether a non-top itag ever succeeds after the top one fails — I don't have that evidence either. But one detail supports @skoruppa's "longer loading" observation: `resolveReachableUrl` already walks the `mn` mirror list for each URL, so the added retry multiplies rather than adds — up to four candidates, each doing its own mirror sweep, before the HLS fallback is reached. If you go with (a), it would be worth logging the failure status per candidate too, not just whether a non-top one succeeded: if the top candidate's mirrors are all returning 403 for a reason that is per-format (rather than per-host), sibling itags will fail the same way and the extra latency buys nothing. That would distinguish "retry helps" from "retry only delays the fallback".

Splitting the ranking change (`rankTmdbVideoCandidates`) out from the retry seems worthwhile regardless — it stands on its own and is much easier to reason about than the probe behaviour.

PR #2106, opened 2026-05-27
feat: DV7 Profile 7 to 8.1 conversion, buffer track, and custom playback buffer engine

ram130:
## Summary
Adds real-time Dolby Vision Profile 7 to 8.1 conversion, buffer track, and a custom playback buffer engine. Replaces the existing "DV7 - HEVC" option with a DV7 Handling setting.
- **DV7 Handling mode**: AUTO, Convert to DV8.1, Strip to HDR10, or Off (Native). Strip to HDR10 is the old "DV7 - HEVC" behavior, now one of the modes. Migration path also added, so no user regression. Device specific pre determined modes also included in auto logic per feedback. 
- Convert DV5 to DV8.1: signals DV5 as 8.1 for HDR10-only displays.
- **Advanced** conversion-mode override: manually pick the libdovi mode.
- DV7 MKV needs a modified extractor — the stock MatroskaExtractor discards the DV7 enhancement-layer/RPU BlockAdditional before it reaches a TrackOutput. Rather than fork all of media3, this switches the DV path to AAR-mode media3 and vendors a single app-level MatroskaExtractor that surfaces the RPU; the media3-source submodule is removed. 
- Rewrote to not need its own fork of media3 or exo like the original PR from another contributor needed.
- **Code optimizations** for playback and buffering during converting
- New defaults: DV7 Handling on (AUTO); custom network and buffer settings off.
- Custom buffer engine: device-aware memory budget (with manual override,) back buffer, optional parallel connections and VOD disk cache.
- Buffered-ahead track on the seek bar, follows the theme color.
- **Stall watchdog**: detects rebuffers/stalls and feeds a playback diagnostics card. For stalls, it assists in recovery of the stream from a stuck buffered state.
- DV diagnostics card to help users report DV-related issues.
- **Subtitle handling preserved** through the DV7 extractor swap (ASS/SSA + embedded fonts still render; pipeline rebuilds when libass is needed).
- libdovi: prebuilt static archives for arm64-v8a / armeabi-v7a / x86 / x86_64 under DV7/libdovi, built from dovi_tool C API. CMake links them directly; stub mode when absent.

## PR type

- [ ] Translation/localization only
- [ ] Critical bug fix

## Why
[...]

PR #3130, opened 2026-08-21
fix: buffer engine #2106 follow up Pt.2: vod cache patch, updated buffer settings and fixes

### LINKED ISSUE #725: Native Player Buffering Issues and Needed Improvements
### Area (tag)

Playback

### Problem statement

The native player often drops the stream and pauses even though the cache is still building. Playback usually doesn’t resume unless you manually skip forward or backward, after which it suddenly catches the buffered data. This is quite annoying.

It also suggests that the buffer size may be too small. Ideally, it should work more like YouTube, where you can continue watching smoothly up to the point already buffered on the progress bar, even if the internet connection temporarily drops. In the current state, the buffering behavior doesn’t seem consistent or stable.

Increasing the buffer size to at least 500 MB could significantly improve playback stability, especially for large files (40 GB+ movies). Large streams require a bigger buffer to maintain smooth playback without interruptions.

Since playback is the most critical part of any streaming app, NuvioTV should focus more on improving these aspects:

Larger and more stable buffer size

No random stream drops or pauses while buffering

Buffer should not reset to 0 after pausing for a longer time

Playback should not restart the streaming process from scratch each time

A solid and reliable buffer progress bar so users can watch smoothly up to the buffered point

It should also display the buffer progress bar color, so users can easily identify how far they can jump ahead for smooth playback similar to how YouTube shows the buffered portion.

Improving the buffering system would greatly enhance the overall streaming experience.

### Proposed solution

I am just an observer 

No, this is only a proposal

Nothing 

Nothing

ram130:
## Summary
**VOD disk cache patches and updates**
- `sharedSimpleCache` never held anything. Playback always streamed direct while settings exposed a cache size slider.
- The cache now clears itself after playback ends, so a stale cache file is not left behind on a normal exit, a crash, or an error.
- The cache window now trims behind the playhead. It previously tried to infer the write head from cache write positions, but a header read at 0, the jump to a resume point and an mp4 index at the tail all look alike as they arrive, so it either trimmed the whole window or never ran at all. The controller now reports the playhead as a byte offset and spans more than 1 GB behind it are dropped.
- Cache writes moved off the read path. They measured at roughly 30 percent of playback wall clock on a high bitrate remux and were starving the forward buffer.
- New VOD logs, plus Disk Cache and Disk Cache Usage rows in the diagnostics card.
- Optimization: Reads reaching the disk cache were 184 bytes on average, 1.4 million of them for a few minutes of playback, because the Matroska extractor pulls EBML fields one at a time. Each one paid the cache's per-call cost, which measured at roughly 7 microseconds and added five seconds before first frame. A 64KB buffer between the extractor and the cache removed it: first frame with the cache on went from 5 seconds slower than direct to 1.5 seconds faster.
**Memory and buffers**
- DV7: conservative path removed and memory ratios updated against every fix since the original release. No more crashes when DV7 is active on low tier devices.
- Low RAM rules and safe guards updated. The memory budget no longer collapses to the 25 MB floor on small heaps because of a flat 210 MB reserve.
- New buffer defaults for both Exo and native Exo, and the defaults are no longer overridden when native Exo is turned on.
- Native Exo now defaults to a 15s back buffer, and resetting to defaults while it is on restores the native values rather than the Exo custom ones.
- Byte budget and back buffer no longer share one flag.
- `BUFFER_NETWORK` logged the raw setting while claiming post gating state.
- `PLAYBACK_CONFIG` reported `native=false` while the native engine was running, because the flag it read is gated behind the parallel network toggle.
- The back buffer is now charged against the target buffer size instead of sitting on top of it. Media3's byte target does not gate the allocator below `minBufferMs`, so a 15s back buffer on a 95 Mbps remux reached 325 MB against a 200 MB target. The byte target now gates everything the allocator holds, and the back buffer is capped at half of `minBufferMs` so the forward buffer keeps the larger share when both cannot fit.
- Some code cleanups.
## PR type
- [x] Critical bug fix
## Why
Two separate problems
1. Playback crashes with `OutOfMemoryError` on low memory devices during DV7 conversion when the memory budget is off. The crash lands around 80 seconds into playback with the heap pinned at 383 of 384 MB and repeated blocking GCs over 300 ms.
2. The VOD disk cache has never worked. The setting and its size slider have shipped for a while, but nothing was ever written to disk, so every seek went back to the network and users had a control that did nothing.
Fixing the cache surfaced three more bugs behind it: a trim window that could not tell playback progress from a container index read, cache writes stalling the read path, and the cache surviving playback with no way for a user to clear it short of clearing app data.
## Issue or approval
Fixes #725
Follow up #2106
## Reproduction steps
For the disk cache:
1. Enable the VOD cache on the public build and set the back buffer to 0.
2. Play a remux or blu ray over 50 GB.
3. Press back to seek backwards, a few times during playback.
4. Every backward seek issues a new network request for data already played instead of reading it from disk. Only the back buffer ms holds anything, and that adds more memory pressure than low RAM devices can take.
For the OOM:
1. On a device with a 384 MB heap, turn the memory budget off.
2. Play a DV7 remux.
3. Playback dies with `OutOfMemoryError` around 80 seconds in.
## UI / behavior impact
## Policy check
## Scope boundaries
- Nothing here changes behaviour when the VOD cache is off, apart from the memory budget work.
- The read buffer sits in front of the cache only. When the cache is off the data source chain is unchanged.
- No new settings keys are cloud synced. The VOD cache keys stay in `localOnlyPlayerProfileSettingsKeys`.
- `sharedConnectionPool` stays a single process wide instance. #2409 made it rebuild in `updateSettings` when the pool size changes, which hands data sources that were already built a pool nothing else uses, so warmed connections are never reused. Sizing it from the connection count is still the right idea and would be better done once at construction; this PR keeps the single instance and leaves that for a follow up. The memory usage calculation added alongside it is kept.
## Testing
Xiaomi MiTV-AFMU0 and Fire TV Stick 4K Max 2nd gen (AFTKRT). DV7 and non DV7 remuxes over debrid, up to 126 GB and 95 Mbps.
Before: `OutOfMemoryError` around 80 seconds in, heap at 383 of 384 MB, repeated blocking GCs over 300 ms.
After, measured across logged sessions:
- Trim window: on a 100 GB remux resumed 27 GB in, 984 MB trimmed against 1694 MB written, holding 710 MB of a 2048 MB cap with back seeks still served from disk. Before the change nothing was trimmed and the cache ran to the cap, forcing 1356 MB of eviction mid playback.
- Cache clears on exit: every following playback starts at `space=0MB`, and the removed bytes match the previous session's written bytes.
- Cache serves seeks: 575 MB read back from disk in one session, and a backward seek resolves without a network round trip.
- Back buffer inside the target: with native on, target 200 MB and a 15s back buffer on a 118 GB remux, usage stays under the target. Before the change the same settings reached 325 MB.

Unit tests added for AUTO cache sizing.
## Screenshots / Video
The diagnostics card with the new Disk Cache and Disk Cache Usage rows. Buffer and Network also gains a storage wear note.
DiagCardinfo
## Breaking changes
None.
## Linked issues
Fixes #725
## Test APK
Try the new defaults on custom network buffer, and turn on VOD after.
Downloader code: 998 6496
Link: https://app.box.com/s/fhakfb6r7v1tag3d0mmict4yorfpcmln


halibiram:
Great work man. VOD cache is working now, and when you seek backward, it continues without rebuffering. Since users won't get the buffer changes directly, it might be a good idea to add a button under advanced settings like a 'reset player settings' button similar to 'clear cw cache'. That way, users can reset all locally saved player settings without having to reinstall the app. @skoruppa what are your thoughts on this?

ram130:
> Great work man. VOD cache is working now, and when you seek backward, it continues without rebuffering. Since users won't get the buffer changes directly, it might be a good idea to add a button under advanced settings like a 'reset player settings' button similar to 'clear cw cache'. That way, users can reset all locally saved player settings without having to reinstall the app. @skoruppa what are your thoughts on this?

Good to know! Forward seek buffered also works fast too since it has less pressure. We do have a reset settings in the buffer and network menu that will do so. I could move that to the top. 

Also considering a potential migration path to the defaults, except for some settings they touched. Let me know your thoughts

JayBog13:
🔥

ram130:
> 🔥

Haha love the emojis and support.

JayBog13:
Yes sir🫡

ram130:
Update: an edge case with fire tv devices. 

Currently fire tv streams smoothly up to 60gb files. After which buffer ahead takes a hit, and vod does not help here due to fire tv slow storage. 

Fire tv GC and background processing is causing more issues than I like. 

But it all may be leading to one bug, some reason back-seek isn't reading the vod cache correctly, only on Amazon devices.  Diagnostics confirm it's active on the OS level. Something else is blocking.  Causing stutter on native exo and regular. 

So far this is an edge case only.  

Cache off returns things to normal levels.

Otherwise other devices are ok with the defaults up to 89GB files. 

Next up:
- Complete research on any limits Amazon may have implemented. Fix bug. 
- Test on an even older device, Google TV with Chromecast. 2gb ram. 8gb storage. 
- Last test 150gb files

-  Possibly resync network buffer settings between exo and native 

- Finalize defaults and migration path

ram130:
Found the issue. Testing today.

skoruppa:
How is testing going? ;)

halibiram:
> Good to know! Forward seek buffered also works fast too since it has less pressure. We do have a reset settings in the buffer and network menu that will do so. I could move that to the top. 
> 
> Also considering a potential migration path to the defaults, except for some settings they touched. Let me know your thoughts

Got it. There are reset buttons there, but what I mean is a more comprehensive function that resets all non synced configurations. Some users try to fix issues caused by dataStore corruption or misconfiguration by uninstalling and reinstalling the app. A button that does the exact same thing without needing that would solve it. When pressed, we can show a modal explaining what will be reset. A better name for this would be "Factory Reset

ram130:
> How is testing going? ;)

Good now that I got some rewriting stuff done. Waiting on one user’s feedback before testing others.  I'll let you know when I'm ready for an official apk.

ram130:
Just a update, added in some fixes discovered during deep testing, some bugs was already in the public build...

added a fix for a mp4 issue with vod  cache trim window .. fix MP4 moov tail write conflict, rewrote the window code..added a line for dv convert summary card to show last time it ran...added a stutter fix caused by a under run  by audio when tunneling is on (viewed as video stutter) on fire tv devices caused by it's Dolby HAL drifts - PTS diff STC goes 11 to  92 to 1176 to  317. All of it in AudioTunnelStreamOut / MS12StreamOut (audio HAL) .fix is a slight buffer increase..underruns back to or near 0...updated to latest commits. 

a user reported on public build when playing mp4 from BTM from sheep detectives, there is frequent buffering after 5s, from what I found, in the logs its flooding frame rate changes and it makes it look like buffering screen, or slight stutter...seeing if this is a simple fix, or more involved, if it is, new pr.... (**update**: _will ignore, fixed with tunneled playback on, fire tv decoder is bugged_)

I'm doing some deep testing still and found the issues long the way, so fixing it within this pr as well.

I  need the hud  info as well, so gonna submit that pr next so I can stop going back and forth.

ram130:
APks now ready for broader testing..link in PR.

halibiram:
Why is it taking so long to reach the first frame? Isn't 14 seconds way too long? Is this issue caused by the test device?

ram130:
> Why is it taking so long to reach the first frame? Isn't 14 seconds way too long? Is this issue caused by the test device?

last part I'm looking at, saved it for last, i think i know the cause just didn't get to it.

ram130:
Update, due to the cache on the read part, DV increased first frame are expected, currently debugging a more optimized option, PR is at 99% ready.

ram130:
**Update**...

I was able to reduce the impact on the read path vod cache uses by 2x. First frame is no longer affected and if  it is, the difference is 1s. Which is fine, due to the speed of the disk and whether or not the OS is also doing something or the CPU got busy. 

Good news, with patches in this PR, the ahead buffer is now even more stabilized under extreme pressure.

Public build 0.8.11

dvhudstat

This PR 

dvhudstat1


This also applies without VOD cache. 

Next steps: 

- Updated APK
- Testing on more devices
- PR no longer draft 


> Why is it taking so long to reach the first frame? Isn't 14 seconds way too long? Is this issue caused by the test device?

Looks like it was the latest commits that came in. I'm seeing a similar issue in the public build now. 

dv

this PR 

dv1

Same settings: native, target 75, defaults 


Something changed in recent commits to affect or increase the first frame.

haaihond:
> **Update**...
> 
> I was able to reduce the impact on the read path vod cache uses by 2x. First frame is no longer affected and if  it is, the difference is 1s. Which is fine, due to the speed of the disk and whether or not the OS is also doing something or the CPU got busy. 
> 
> Good news, with patches in this PR, the ahead buffer is now even more stabilized under extreme pressure.
> 
> Public build 0.8.11
> 
> dvhudstat
> 
> This PR 
> 
> dvhudstat1
> 
> 
> This also applies without VOD cache. 
> 
> Next steps: 
> 
> - Updated APK
> - Testing on more devices
> - PR no longer draft 
> 
> 
> > Why is it taking so long to reach the first frame? Isn't 14 seconds way too long? Is this issue caused by the test device?
> 
> Looks like it was the latest commits that came in. I'm seeing a similar issue in the public build now. 
> 
> dv
> 
> this PR 
> 
> dv1
> 
> Same settings: native, target 75, defaults 
> 
> 
> Something changed in recent commits to affect or increase the first frame.

Results on 0.8.12 and 0.8.9, within margin of error. It doesn't seem to be something in new commits. If your testing has been done on TB that might be your issue, since api response time is unreliable.

ram130:
> > **Update**...
> > 
> > I was able to reduce the impact on the read path vod cache uses by 2x. First frame is no longer affected and if  it is, the difference is 1s. Which is fine, due to the speed of the disk and whether or not the OS is also doing something or the CPU got busy. 
> > 
> > Good news, with patches in this PR, the ahead buffer is now even more stabilized under extreme pressure.
> > 
> > Public build 0.8.11
> > 
> > dvhudstat
> > 
> > This PR 
> > 
> > dvhudstat1
> > 
> > 
> > This also applies without VOD cache. 
> > 
> > Next steps: 
> > 
> > - Updated APK
> > - Testing on more devices
> > - PR no longer draft 
> > 
> > 
> > > Why is it taking so long to reach the first frame? Isn't 14 seconds way too long? Is this issue caused by the test device?
> > 
> > Looks like it was the latest commits that came in. I'm seeing a similar issue in the public build now. 
> > 
> > dv
> > 
> > this PR 
> > 
> > dv1
> > 
> > Same settings: native, target 75, defaults 
> > 
> > 
> > Something changed in recent commits to affect or increase the first frame.
> 
> Results on 0.8.12 and 0.8.9, within margin of error. It doesn't seem to be something in new commits. If your testing has been done on TB that might be your issue, since api response time is unreliable.
> 
> 
> 
> 
> 
> 
> 
> 

Thanks for confirming. Yes on TB. If you want feel free to test the apk as well.

haaihond:
Tested on your build, and @ram130 you have nothing to worry about.

ram130:
> Tested on your build, and @ram130 you have nothing to worry about.
> 
> 

Thanks for confirming. I'll see if I can add TB instant instead and see how that is. 


Update: much better.  
image

ram130:
Testing completed on different devices. Positive results and no issues. 

I'll let migration path not touch those users who already customize the settings. Current already updates defaults if a user never touched it but left it on. Those with issues on the public build can hit reset after updating to get the new defaults, if they enable native or toggle they'll get the new defaults as well. 

PR is ready for merge. @skoruppa

ram130:
New build live,

- new commits merging in from remote
- added back buffer fix; preventing it from starving forward buffer on high remux bit rates
- drop audio related fixes due to new audio engine work I'm apart of in  a different pr.
-  Retested new defaults on fire tv to confirm performance stability and low cpu with vod cache on. 
- confirmed no regression expected with vod off on lower tier devices, only improvement

skoruppa:
I think it is the time ;) Is this ready to be merged? Was the latest build tested?

skoruppa:
@halibiram any remarks?

halibiram:
> @halibiram any remarks?

It looks good overall. I haven’t had a chance to test the latest changes yet. give me a bit of time and I’ll check once I’m home

halibiram:
Tested it. VOD cache and the buffer side are going in the right direction. a few things stood out. 64KB read buffer is wrapping every progressive stream, not only when cache is on. pr says the chain doesnt change when cache is off but the code does. I think it should only wrap when cache is actually active. cache wipe on app start also looks like it can race with the first playback. It starts deleting in the background while the new title is already writing.
And the cache writer close has no timeout. on slow storage like fire stick that might stall the load thread on seek or exit

ram130:
> Tested it. VOD cache and the buffer side are going in the right direction. a few things stood out. 64KB read buffer is wrapping every progressive stream, not only when cache is on. pr says the chain doesnt change when cache is off but the code does. I think it should only wrap when cache is actually active. cache wipe on app start also looks like it can race with the first playback. It starts deleting in the background while the new title is already writing. And the cache writer close has no timeout. on slow storage like fire stick that might stall the load thread on seek or exit

Thanks, confirmed and fixed all 3 and some additional fixes: 

64KB read buffer: it now only wraps the stream when the cache is active. With the cache off the data source chain is unchanged, which matches the description again.

Cache wipe race: the previous run's folder is now renamed aside before the cache opens and deleted in the background with SimpleCache.delete, so the live cache and the cleanup never share files. It was slightly worse than a race with writes, since a resumed title shares its key with the leftovers and could lose its fresh spans.

Writer close: close now waits at most 500 ms. After that it marks the span failed, so the queued writes are skipped, and it only waits for the one write already running. The delegate still commits the part written so far, with no gaps.

Tested on the Fire TV Stick 4K Max 2nd gen: after a force stop with the cache on, the next launch started at space=0MB and the resumed title served back seeks from disk. On a 93 Mbps remux with seeks into uncached ranges, 10 span closes took 21 ms in total with no timeouts and no write errors.

While going through it I also fixed a few things of my own: the player was kept in memory after exit through process-wide references, the cache key could collide on multi-file torrents without a file index, a KDoc had ended up above the wrong function, a stale comment, and hard-coded diagnostics strings. They're in the same commit.

For your the device-wide reset for all non-synced settings you mentioned early on, I'll do a separate PR for it as I want to also update the Reset defaults coverage in network settings as well. 

APKs updated with the latest changes as well. 

> I think it is the time ;) Is this ready to be merged? Was the latest build tested?

@skoruppa yep and now ready to merge. Added in @halibiram feedback. 

Please also add an update note "If you had buffer issues with custom network buffers on or before in general, please reset to the new defaults using the reset button in that area"

BringerOfRainX1:
I tested a few streams and all worked perfectly for me on my Xiaomi 3rd gen box.

skoruppa:
@halibiram if no remarks from you, I will merge it :)

halibiram:
> @halibiram if no remarks from you, I will merge it :)


Looks good to me. go ahead and merge :)

ram130:
> I tested a few streams and all worked perfectly for me on my Xiaomi 3rd gen box.

Thank you for testing and being great. Shout out to @halibiram  and @skoruppa for pushing and being amazing.  Now I can get back to the Native NNTP usenet and Audio system work haha.

PR #3139, opened 2026-08-22
fix: per-episode rating from addon metadata is never parsed

### LINKED ISSUE #3129: [Bug]: Per-episode "rating" from addon metadata is never parsed
### Pre-flight checks

### App version / commit

0.8.7

### Install method

GitHub Release APK

### Platform

Android emulator

### Device model

LDPlayer 9

### Android version

Android 9

### Player mode

Internal player

### Area (tag)

Other

### Bug description

Filing this on Android TV because NuvioMedia/NuvioWeb#690 was closed with:

> Adding this fallback in Web alone would make the two platforms show different results
> when that service is unavailable or differs. For that reason, we can't accept this as a
> Web-only fix. Supporting addon-provided episode ratings would need to be a coordinated
> cross-platform contract and UX change, with Android TV updated first or alongside Web.

That is what this issue is for.

`VideoDto` has no field for `rating`, so the value never survives deserialization:

```kotlin
// data/remote/dto/MetaResponseDto.kt:77-91
data class VideoDto(
    @Json(name = "id") val id: String,
    @Json(name = "name") val name: String? = null,
    @Json(name = "title") val title: String? = null,
    @Json(name = "released") val released: String? = null,
    @Json(name = "thumbnail") val thumbnail: String? = null,
    @Json(name = "streams") val streams: List<StreamDto>? = null,
    @Json(name = "season") val season: Int? = null,
    @Json(name = "episode") val episode: Int? = null,
    @Json(name = "number") val number: Int? = null,
    @Json(name = "overview") val overview: String? = null,
    @Json(name = "description") val description: String? = null,
    @Json(name = "runtime") val runtime: String? = null,
    @Json(name = "available") val available: Boolean? = null
)
```

`rating` is the field the Stremio meta object uses for per-episode ratings, and Cinemeta
populates it (all 67 videos on Breaking Bad, `tt0903747`). Because it is dropped here,
episode ratings can only come from `ImdbEpisodeRatingsRepository`, which needs an IMDb or
TMDB id.

That covers most series, so the gap is invisible until an addon uses its own ids.
`MetaDetailsViewModel` then resolves neither id, sets `episodeImdbRatings = emptyMap()`,
and two things go blank even though the addon sent a rating for every episode:

- every episode card shows no rating
- the **Ratings** tab shows "Episode ratings are unavailable." (`R.string.ratings_unavailable`)

### Steps to reproduce

1. Install Nuvio TV.
2. Add an addon that uses its own ids and provides per-episode ratings.
   I used One Pace Premium: https://stremio-addons.net/addons/one-pace-rd
3. Open any series from that addon.
4. Open a season's episode list -> no rating on any episode card.
5. Open the Ratings tab -> "Episode ratings are unavailable."
6. Compare against the addon's meta response, which has a "rating" value on
   every entry in meta.videos[].

### Expected behavior

Episode cards and the Ratings tab show the rating the addon provided in
meta.videos[].rating when the external episode-ratings service has no value.

### Actual behavior

No rating on any episode card, and the Ratings tab reports them as unavailable. VideoDto
has no rating field, so the value is discarded at deserialization and there is nothing to
fall back to when ImdbEpisodeRatingsRepository returns an empty map.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

**Suggested fix** — four edits, no UX change:

```kotlin
// 1. data/remote/dto/MetaResponseDto.kt - VideoDto
@Json(name = "rating") val rating: String? = null,

// 2. domain/model/Meta.kt - Video
val rating: Double? = null,

// 3. data/mapper/MetaMapper.kt - VideoDto.toDomain()
rating = rating?.trim()?.toDoubleOrNull()?.takeIf { it > 0.0 },

// 4. MetaDetailsViewModel - seed the map the UI already reads
val addonRatings = meta.videos.mapNotNull { v ->
    val s = v.season ?: return@mapNotNull null
    val e = v.episode ?: return@mapNotNull null
    val r = v.rating ?: return@mapNotNull null
    (s to e) to r
}.toMap()

// where neither id resolves: use addonRatings instead of emptyMap(),
// and only set ratings_unavailable when addonRatings is empty as well
// otherwise:                   episodeImdbRatings = addonRatings + ratings
```

`rating` is a `String?` on the DTO to match how `runtime` is already handled, then parsed
in the mapper.

**Seeding the map rather than patching a call site** means both places that show a rating
are covered by one change — the episode cards and the Ratings tab both read
`episodeImdbRatings`, so fixing only the card would leave the tab saying
"Episode ratings are unavailable."

**The ratings repository always wins.** `addonRatings + ratings` lets the right side
override, so any episode the service covers keeps showing exactly what it shows today.
The addon value is only visible where the service returned nothing.

**Nothing extra is needed for unrated series.** Cinemeta sends `"rating": "0"` when it has
no data — all 70 videos on Stranger Things (`tt4574334`). `EpisodeCard` already guards with
`imdbRating?.takeIf { it > 0.0 }` (`EpisodesSection.kt:494`), and the `takeIf` in the mapper
keeps a zero out of the model, matching what `ImdbEpisodeRatingsRepository.toRatingsMap()`
already does with `voteAverage`.

**No UX change.** Both the badge and the Ratings tab already exist and already render
whenever the repository has values. This only fills them when it has none.

**On the cross-platform contract.** Worth noting the platforms already differ here: Web's
`toEpisodeEntry()` accepts an addon-supplied per-episode rating under four keys
(`imdbRating`, `imdb_score`, `ratings.imdb`, `mdbListRatings.imdb`), while Android accepts
none. NuvioMedia/NuvioWeb#690 was not asking for a new contract on Web so much as the one spelling Cinemeta
actually uses. Landing it on Android is what would make the two match.

**Happy to do this as a coordinated set.** I have the equivalent change written for all
four clients and built and verified on Desktop (screenshots in
NuvioMedia/NuvioDesktop#392). If you want it, I can open matching PRs on NuvioTV,
NuvioMobile, NuvioDesktop and NuvioWeb together so no platform is left diverging at any
point — or just this one first, whichever you prefer.

6ip:
## Summary

`VideoDto` has no field for `rating`, so a per-episode rating sent by a metadata addon is dropped at deserialization. This adds it, parses it in `MetaMapper`, and seeds `episodeImdbRatings` from `meta.videos` so both the episode cards and the Ratings tab can use it.

## PR type

- [x] Critical bug fix

## Why

Cinemeta and the TMDB addon both send `rating` on `meta.videos[]`. Because `VideoDto` never declares it, the value is discarded and episode ratings can only come from `ImdbEpisodeRatingsRepository`, which needs an IMDb or TMDB id.

That covers most series, so the gap only appears with addons that use their own ids. `MetaDetailsViewModel` resolves neither id, sets `episodeImdbRatings = emptyMap()`, and two things go blank even though the addon sent a rating for every episode: the episode cards, and the Ratings tab, which shows "Episode ratings are unavailable."

Discussed in #3129, where the change was reviewed favourably.

## Issue or approval

Fixes #3129

Worth flagging honestly: this template offers "Translation/localization only" or "Critical bug fix", and neither is an exact fit — the change adds a nullable field so existing addon data reaches the UI. I have ticked the closer of the two and am happy to change it, split it, or hold it if you would rather categorise it differently.

## Reproduction steps

1. Install NuvioTV 0.8.7 from GitHub Releases.
2. Add an addon that uses its own ids and provides per-episode ratings. I used One Pace Premium: https://stremio-addons.net/addons/one-pace-rd
3. Open any series from that addon and open a season's episode list.
4. No rating on any episode card.
5. Open the Ratings tab — "Episode ratings are unavailable."
6. Compare against the addon's meta response, which has a `rating` on every entry in `meta.videos[]`.

## UI / behavior impact

The rating badge and the Ratings tab already exist and already render whenever the repository returns values. This only fills them when it returns none, so no layout changes.

## Policy check

## Scope boundaries

Four files. Deliberately not included:

- No change to `ImdbEpisodeRatingsRepository` or to how the series graph is called. It keeps priority: `addonRatings + ratings` lets the right side override, so any episode the repository covers is unaffected.
- No new UI, no layout or styling changes.
- No change to how ratings are formatted. `%.1f` in both places already rounds the three-decimal values TMDB sends, so `8.519` shows as `8.5`.
- The same defect is on NuvioDesktop (NuvioMedia/NuvioDesktop#443, open), NuvioWeb (NuvioMedia/NuvioWeb#690) and NuvioMobile. Nothing here depends on those.

## Testing

Built with `:app:assembleFullDebug` and installed on LDPlayer 9 (Android 9), alongside the 0.8.7 release APK for the before/after.

- An addon using its own ids — cards blank and Ratings tab unavailable before; both populated after.
- Breaking Bad — ratings from Cinemeta's `meta.videos[].rating`. Note a debug build has no rating API URLs, so the repository returns nothing and these come from the addon.
- Unrated series are unaffected: `"0"`, `"0.0"`, `"0.000"`, `"-1"`, `""` and a missing field are all dropped by `takeIf { it > 0.0 }` in the mapper and never enter the map.
- `compileFullDebugKotlin` passes with no new warnings.

I checked metadata display only, not playback.

## Screenshots / Video

Before, on the 0.8.7 release APK. One Pace Premium sends a `rating` for every episode; the cards are blank and the Ratings tab reports them unavailable:

Image

After, same addon on a build with the change:

Image

Stranger Things on the patched build. Cinemeta sends `"rating": "0"` for all 70 of its videos and the existing `takeIf { it > 0.0 }` keeps them blank, so series without ratings are unchanged:

Screenshot_20260822-153752

## Breaking changes

None. `rating` is a new nullable field with a default, and the addon value is only read where the repository returned nothing.

## Linked issues

Fixes #3129

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/mapper/MetaMapper.kt
+++ b/app/src/main/java/com/nuvio/tv/data/mapper/MetaMapper.kt
@@ -75,6 +75,7 @@ fun VideoDto.toDomain(episodeLabel: String = "Episode"): Video {
         episode = episode ?: number,
         overview = overview ?: description,
         runtime = parseEpisodeRuntimeMinutes(runtime),
+        rating = rating?.trim()?.toDoubleOrNull()?.takeIf { it > 0.0 },
         available = available
     )
 }
--- a/app/src/main/java/com/nuvio/tv/data/remote/dto/MetaResponseDto.kt
+++ b/app/src/main/java/com/nuvio/tv/data/remote/dto/MetaResponseDto.kt
@@ -87,6 +87,7 @@ data class VideoDto(
     @Json(name = "overview") val overview: String? = null,
     @Json(name = "description") val description: String? = null,
     @Json(name = "runtime") val runtime: String? = null,
+    @Json(name = "rating") val rating: String? = null,
     @Json(name = "available") val available: Boolean? = null
 )
 
--- a/app/src/main/java/com/nuvio/tv/domain/model/Meta.kt
+++ b/app/src/main/java/com/nuvio/tv/domain/model/Meta.kt
@@ -181,6 +181,8 @@ data class Video(
     val episode: Int?,
     val overview: String?,
     val runtime: Int? = null, // episode runtime in minutes
+    /** Per-episode rating supplied by the addon, when it provides one. */
+    val rating: Double? = null,
     val available: Boolean? = null
 )
 
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
@@ -1343,6 +1343,15 @@ class MetaDetailsViewModel @Inject constructor(
                 )
             }
 
+            // Ratings the addon supplied on meta.videos[].rating. The repository below
+            // still wins wherever it has an entry.
+            val addonRatings: Map<Pair<Int, Int>, Double> = meta.videos.mapNotNull { video ->
+                val season = video.season ?: return@mapNotNull null
+                val episode = video.episode ?: return@mapNotNull null
+                val rating = video.rating ?: return@mapNotNull null
+                (season to episode) to rating
+            }.toMap()
+
             try {
                 val tmdbContentType = resolveTmdbContentType(meta)
                 if (tmdbContentType !in listOf(ContentType.SERIES, ContentType.TV)) {
@@ -1368,9 +1377,13 @@ class MetaDetailsViewModel @Inject constructor(
                             state
                         } else {
                             state.copy(
-                                episodeImdbRatings = emptyMap(),
+                                episodeImdbRatings = addonRatings,
                                 isEpisodeRatingsLoading = false,
-                                episodeRatingsError = localizedContext.getString(R.string.ratings_unavailable)
+                                episodeRatingsError = if (addonRatings.isEmpty()) {
+                                    localizedContext.getString(R.string.ratings_unavailable)
+                                } else {
+                                    null
+                                }
                             )
                         }
                     }
@@ -1387,7 +1400,7 @@ class MetaDetailsViewModel @Inject constructor(
                         state
                     } else {
                         state.copy(
-                            episodeImdbRatings = ratings,
+                            episodeImdbRatings = addonRatings + ratings,
                             isEpisodeRatingsLoading = false,
                             episodeRatingsError = null
                         )
--- a/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/detail/MetaDetailsViewModel.kt
@@ -1357,7 +1357,7 @@ class MetaDetailsViewModel @Inject constructor(
                 if (tmdbContentType !in listOf(ContentType.SERIES, ContentType.TV)) {
                     _uiState.update {
                         it.copy(
-                            episodeImdbRatings = emptyMap(),
+                            episodeImdbRatings = addonRatings,
                             isEpisodeRatingsLoading = false,
                             episodeRatingsError = null
                         )
@@ -1415,9 +1415,13 @@ class MetaDetailsViewModel @Inject constructor(
                         state
                     } else {
                         state.copy(
-                            episodeImdbRatings = emptyMap(),
+                            episodeImdbRatings = addonRatings,
                             isEpisodeRatingsLoading = false,
-                            episodeRatingsError = localizedContext.getString(R.string.ratings_load_error)
+                            episodeRatingsError = if (addonRatings.isEmpty()) {
+                                localizedContext.getString(R.string.ratings_load_error)
+                            } else {
+                                null
+                            }
                         )
                     }
                 }


skoruppa:
Thanks for the PR. Will you create similar one for the mobile client?

6ip:
Yes, I will open one for mobile today. The change is ready and tested locally, I just wanted to see how this one went first.

Where the other clients stand:

- **Desktop** — issue NuvioMedia/NuvioDesktop#392 and PR NuvioMedia/NuvioDesktop#443, both opened yesterday. Same fix, three files.
- **Web** — same change, and there it also fills the Ratings tab, which otherwise says "Ratings not available". Ready but not opened yet: NuvioMedia/NuvioWeb#690 was closed asking for Android TV to land first or alongside, so I held it.
- **Mobile** — tested locally, issue and PR coming in a few hours.

I'll tag you on the mobile one. I'm also `ur.e` on Discord if that's easier for anything.
Thanks for considering @skoruppa

skoruppa:
The catch block still sets episodeImdbRatings = emptyMap(). Since we now fall back to addonRatings when TMDB/IMDB IDs are missing, shouldn't we do the same here? A transient network error would wipe out perfectly valid addon ratings that we already have in hand

6ip:
you're right. `addonRatings` is built before the `try` and doesn't touch the network, so losing it because the series graph call failed is just throwing away data we already had.

Fixed: the catch keeps the addon ratings now, and only shows the error when there's genuinely nothing left to display.

While I was in there I found the same thing one branch up — the early return for `tmdbContentType !in listOf(SERIES, TV)` also sits inside the `try` and was discarding them the same way. Changed that one too. Every exit now either falls back to the addon ratings or hasn't parsed them yet.

Desktop and mobile don't have this problem by the way, they read `episodeRatings[it] ?: episode.rating` at the card instead of seeding a map, so an empty map still shows the addon value.

Thanks for the review.

PR #3146, opened 2026-08-22
Feat: Make Preferred Subtitle Language follow the system language

### LINKED ISSUE #3096: [Feature]:     Default Preferred Language to System Default
### Area (tag)

Settings

### Problem statement

The Preferred Subtitle Language setting defaults to English, regardless of the system language.

### Proposed solution

The default Preferred Subtitle Language should be System Default, similar to the app language.

This would allow subtitles to automatically follow the language configured on the device unless the user explicitly selects a different preferred subtitle language.

No, this is only a proposal

_No response_

The same applies to Integration → TMDB → Language. However, the Preferred Audio Language should remain English.

haveAnIssue:
## Summary

Preferred Subtitle Language now follows the device’s system language by default, consistent with the existing App Language behavior.

When no subtitle language has been manually selected, subtitles use the device’s system language. If that language isn’t supported, the default remains English. Any language selected manually is still saved and used as before.

## PR type

- [x] Critical bug fix

## Why

App Language already gives users a seamless first experience by automatically matching their device language. However, when a new user starts watching a movie, subtitles immediately default to English, even if the app itself is using a different language.

This makes the experience feel inconsistent and means users have to manually configure subtitles before they can simply start watching.

This PR makes subtitles work the way users would naturally expect: open the app, press play, and everything is already in the right language. It provides a more welcoming, plug-and-play experience without requiring new users to touch the settings.

## Issue or approval

Fixes #3096 

## Reproduction steps

1. Set the device language to a language other than English.
2. Install the app or use an account that has never changed the Preferred Subtitle Language setting.
3. Open the subtitle language setting.
4. Observe that English is selected by default.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

This change is limited to the default value of the Preferred Subtitle Language setting. Explicitly selected subtitle languages continue to work as before.

No unrelated UI changes, refactoring, or additional subtitle behavior changes are included.

## Testing

* Verified that a fresh install uses the device's system language when no subtitle language has been selected.
* Verified that the Settings screen displays System Default in this state.
* Verified that selecting a specific subtitle language persists the selection.
* Verified that an explicitly selected language continues to be used instead of the system language.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

#3096

### DIFF
--- a/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
+++ b/app/src/main/java/com/nuvio/tv/data/local/PlayerSettingsDataStore.kt
@@ -42,6 +42,10 @@ val SubtitleLanguage.displayName: String
 
 const val SUBTITLE_LANGUAGE_FORCED = "forced"
 
+object SubtitleLanguageOption {
+    const val DEVICE = "device"
+}
+
 val AVAILABLE_SUBTITLE_LANGUAGES = listOf(
     SubtitleLanguage("af", "Afrikaans"),
     SubtitleLanguage("sq", "Albanian"),
@@ -134,6 +138,7 @@ val AVAILABLE_TMDB_LANGUAGES = AVAILABLE_SUBTITLE_LANGUAGES + listOf(
  */
 data class SubtitleStyleSettings(
     val preferredLanguage: String = "en",
+    val isPreferredLanguageSystemDefault: Boolean = true,
     val secondaryPreferredLanguage: String? = null,
     val useForcedSubtitles: Boolean = false,
     val showOnlyPreferredLanguages: Boolean = false,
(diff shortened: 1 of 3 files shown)


halibiram:
Special mapping exists only for two languages. What happens for other languages that also need special mapping?

haveAnIssue:
* Hebrew: Returns iw instead of he
* Yiddish: Returns ji instead of yi
* Indonesian: Returns in instead of id 

and I missed ji→yi (Yiddish).
it’s moot for this specific list since Yiddish isn’t in AVAILABLE_SUBTITLE_LANGUAGES, but worth adding for robustness in case it’s ever added.

haveAnIssue:
Works great (:
@halibiram @skoruppa

halibiram:
I tested it on tv by changing the system language, and the PR does what it aims to do. If we need such a feature, we can merge it :)

haveAnIssue:
@halibiram Hey! Thanks for the review.

Could you remove the needs-info label from the PR when you get a chance?

Also, @skoruppa, what do you think, does the new feature fit well with the existing software from your perspective?

skoruppa:
If merged we will need same change for mobile ;)

haveAnIssue:
Yep, totally! @skoruppa @tapframe

skoruppa:
Already merged but I forget to test/ask

Does it work fine with "forced" setting?

halibiram:
> Already merged but I forget to test/ask
> 
> Does it work fine with "forced" setting?

I tested it with Use Forced Subtitles on.
it is works ;)

PR #3156, opened 2026-08-23
feat(torrent): support custom TorrServer endpoint

JaskaranSM:
## Summary
Adds a configurable TorrServer endpoint under **Settings → Playback → P2P Streaming → Custom TorrServer endpoint**.

When a custom endpoint is set, Nuvio routes all TorrServer traffic (health checks, torrent add/drop, stats, stream URLs) to the remote server instead of starting the bundled binary. Leaving it empty uses the built-in server as before.

## What's included
- **`TorrentSettings`** — new `customTorrServerUrl` preference (DataStore-backed) with a synchronous cache used to resolve the effective base URL.
- **`TorrServerBinary`** — `baseUrl` returns the custom endpoint when configured; `start()`/`stop()` become no-ops so the local binary isn't launched.
- **`TorrServerEndpoint`** — URL normalization/validation helper (blank → built-in, missing scheme → `http://`, invalid rejected).
- **Settings UI** — an action row in the P2P section opens a dialog to enter the endpoint. On save the app pings the server's `/echo` endpoint; the Save button shows a loading spinner while checking, and if the server is unreachable the save is blocked with a red "Server not accessible" message. An empty value clears the setting back to built-in.
- **DI** — `TorrentModule` provides the new DataStore binding.

## Testing
- **Unit tests** — settings persistence, endpoint normalization, binary routing (no-op start/stop), `TorrServerApi.isEndpointReachable` against a real loopback HTTP server, and `PlaybackSettingsViewModel.saveCustomTorrServerUrl` (reachable / unreachable / invalid / empty).
- **Compose instrumentation tests** — dialog prefill, normalized save, invalid URL rejection, unreachable server error, loading indicator, clear.
- **Manual** — verified on an Android TV emulator and a real Android TV: an unreachable endpoint shows the red error and blocks save; a reachable endpoint saves and the row reflects the new URL.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -106,6 +106,7 @@ android {
         targetSdk = 36
         versionCode = 1048
         versionName = "0.8.7-beta"
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
 
         buildConfigField("String", "PARENTAL_GUIDE_API_URL", "\"${localProperties.getProperty("PARENTAL_GUIDE_API_URL", "")}\"")
         buildConfigField("String", "INTRODB_API_URL", "\"${localProperties.getProperty("INTRODB_API_URL", "")}\"")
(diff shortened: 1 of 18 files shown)


JaskaranSM:
@tapframe @skoruppa Please let me know if any changes are required.

PR #3184, opened 2026-08-25
fix(player): restore source picker selection after internal playback

### LINKED ISSUE #3183: [Bug]: Returning from internal player resets source selection to the first result
### Pre-flight checks

### App version / commit

0.8.8 beta - dev@f40c422ee

### Install method

Built from source

### Platform

Android emulator

### Device model

Android TV Emulator (Television_4K / sdk_google_atv64_x86_64)

### Android version

Android 16 (API 36)

### Player mode

Internal player

### Area (tag)

Focus / Remote navigation

### Bug description

When a source other than the first result is selected and opened with Nuvio's internal player, pressing Back to return to the source picker loses that selection.

The source list returns to the top and focus lands on the first result. This is particularly disruptive with long source lists because the source that was actually played may be well outside the visible viewport, forcing the user to find it again.

Most of the time, a user returns from the player to the source list because they want to change sources, for example to try a different quality, provider, audio option, or a source that works better. Returning them to the source they just played makes the nearby alternatives immediately accessible. Resetting to the first result means they have to scroll through the same list again every time, which is frustrating and time-consuming when there are many results.

The selection only needs to be restored for the current source-picker navigation entry. It should not be persisted across different movies, episodes, filters, or app sessions.

### Steps to reproduce

1. Open a movie or episode and let the source results load.
2. Move down the source list and select a result other than the first item. For the clearest reproduction, choose one that requires scrolling.
3. Open that source with the internal player.
4. After playback starts, press Back to return to the source picker.
5. Observe the source list position and focused item.

The same behavior can also be checked through **Ask every time > Internal player**.

### Expected behavior

The source picker should return to the source that was opened:

- The previously selected source card should regain focus.
- The list should scroll far enough to make that source visible.
- The restoration should be transient and limited to returning from that internal-player session.
- External-player restoration, source reloads, filter changes, and navigation to different content should remain unchanged.

### Actual behavior

The source picker returns to the beginning of the list and focuses the first source result. The source that was played is no longer selected or necessarily visible.

### Frequency

Always when returning from a non-first source.

### Did this work before?

Not sure

### Media details (optional)

- Content type: movie or episode
- Player: internal
- Reproducible independently of the selected source or addon
- Most visible with a source located outside the initial viewport

```shell

```

### Anything else? (optional)

Nuvio remembers which source was selected, but it does not restore it after returning from the internal player. Instead, the source list goes back to the top and focuses the first result.

The fix should simply return the user to the source they just played, without saving any source details or remembering selections for other titles.

liongalahad:
## Summary

Restore the source picker to the source the user opened when returning from Nuvio's internal player.

The existing selected index is kept for the current source screen. On return, the list scrolls that source into view before restoring focus. The restoration is one-shot and is cleared after it is handled.

## PR type

- [x] Critical bug fix

## Why

Fixes the source-selection focus bug documented in #3183.

Users usually return from the player to the source list because they want to try another source, such as a different quality, provider, audio option, or a source that works better.

Before this fix, returning from the internal player reset the list to the top and focused the first result. With a long source list, users had to repeatedly scroll back to the source they had just played before they could try a nearby alternative.

Source selection is part of the core playback flow, and this behavior made changing sources unnecessarily frustrating and time-consuming.

## Issue or approval

Fixes #3183

## Reproduction steps

1. Open a movie or episode and let its source results load.
2. Scroll down and select a source other than the first result.
3. Open that source with Nuvio's internal player.
4. After playback starts, press Back to return to the source picker.
5. Before this fix, the picker returns to the top and focuses the first source.
6. After this fix, the source that was opened is visible and focused.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

- Applies only when returning directly from a manually selected source in the internal player.
- Auto-play navigation is unchanged.
- Existing external-player restoration remains unchanged.
- Player exits that return to the detail screen or advance to another episode do not request source restoration.
- Source reloads and addon-filter changes retain their existing behavior.
- The saved index is clamped if the source list changes while playback is open.
- No source URLs, headers, provider results, content IDs, or cross-title source history are stored.
- No UI, settings, dependencies, schemas, or unrelated playback behavior are changed.

## Testing

- Ran the focused policy tests:

  `.\gradlew.bat :app:testFullDebugUnitTest --tests com.nuvio.tv.ui.navigation.SourceSelectionRestorePolicyTest --no-daemon`

  Result: 3 tests passed with no failures or skips.

- Built the full debug APK:

  `.\gradlew.bat :app:assembleFullDebug --no-daemon`

  Result: `BUILD SUCCESSFUL`.

- Installed the x86_64 debug APK side-by-side on an Android TV emulator:
  - AVD: `Television_4K`
  - Device model: `sdk_google_atv64_x86_64`
  - Android 16 / API 36
  - Package: `com.nuviodebug.com`

- Confirmed the app installed, launched, and remained running without a fatal startup exception.
- User-tested the source-return flow on the emulator.
- Ran `git diff --check` successfully.

- The full local unit suite was also attempted:
  - 965 tests executed
  - 951 passed
  - 1 skipped
  - 13 failed in unrelated networking, allocator, Dolby Vision, Matroska, sync, collections, airing-rule, and performance tests
  - The new source-selection tests remained green, and none of the failures referenced the changed files.

## Screenshots / Video

Screenshots added in [comment below](https://github.com/NuvioMedia/NuvioTV/pull/3184#issuecomment-5404097581)

## Breaking changes

None.

## Linked issues

Fixes #3183

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/navigation/NuvioNavHost.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/navigation/NuvioNavHost.kt
@@ -470,6 +470,10 @@ fun NuvioNavHost(
             )
         ) { backStackEntry ->
             val streamArgs = backStackEntry.arguments
+            val streamSavedState = backStackEntry.savedStateHandle
+            val restoreSourceSelection by streamSavedState.getStateFlow(
+                SOURCE_SELECTION_RESTORE_STATE_KEY, false
+            ).collectAsState()
             val returnToDetailOnBack = streamArgs
                 ?.getString("returnToDetailOnBack")
                 ?.toBooleanStrictOrNull() == true
@@ -481,6 +485,10 @@ fun NuvioNavHost(
                 ?.toBooleanStrictOrNull() == true
             StreamScreen(
                 startFromBeginning = startFromBeginning,
+                restoreSourceSelection = restoreSourceSelection,
+                onSourceSelectionRestoreHandled = {
+                    streamSavedState[SOURCE_SELECTION_RESTORE_STATE_KEY] = false
+                },
                 onBackPress = {
                     val streamContentType = streamArgs?.getString("contentType").orEmpty()
                     val streamContentId = streamArgs?.getString("contentId").orEmpty()
@@ -739,6 +747,23 @@ fun NuvioNavHost(
                 }
             )
         ) { backStackEntry ->
+            fun popBackToStream(): Boolean {
+                val autoPlayNavigation = backStackEntry.arguments
+                    ?.getString("autoPlayNav")
+                    ?.toBooleanStrictOrNull() == true
+                val restoreEntry = navController.previousBackStackEntry?.takeIf { previousEntry ->
+                    shouldArmSourceSelectionRestore(
+                        autoPlayNavigation = autoPlayNavigation,
+                        previousRoute = previousEntry.destination.route
+                    )
+                }
+                val returnedToStream = navController.popBackStack(Screen.Stream.route, inclusive = false)
+                if (returnedToStream && restoreEntry != null) {
+                    restoreEntry.savedStateHandle[SOURCE_SELECTION_RESTORE_STATE_KEY] = true
+                }
+                return returnedToStream
+            }
+
             PlayerScreen(
                 onBackPress = { currentVideoId, currentSeason, currentEpisode, autoPlayEnabled, playbackCompleted ->
                     val args = backStackEntry.arguments
@@ -829,7 +854,7 @@ fun NuvioNavHost(
                             if (skipStreamScreen) {
                                 returnToDetail()
                             } else {
-                                val returnedToStream = navController.popBackStack(Screen.Stream.route, inclusive = false)
+                                val returnedToStream = popBackToStream()
                                 if (!returnedToStream) {
                                     if (returnToDetailOnBack && contentType.equals("series", ignoreCase = true) && contentId.isNotBlank()) {
                                         returnToDetail()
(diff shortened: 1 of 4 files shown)


liongalahad:
### CI build note

I built the `fullDebug` APK locally from this branch at commit `56484ff` using `.\gradlew.bat :app:assembleFullDebug`, and the build completed successfully.

The failed GitHub Actions check reached `:app:packageFullDebug` and then exited with `java.lang.OutOfMemoryError: Java heap space` in the Android APK packaging code (`ApkFlinger` / `zipflinger`). Kotlin compilation did not fail.

The same `packageFullDebug` heap failure also occurred on [PR #3171](https://github.com/NuvioMedia/NuvioTV/pull/3171) in [this workflow run](https://github.com/NuvioMedia/NuvioTV/actions/runs/32729591751), which suggests intermittent CI runner or Gradle heap pressure rather than a compilation problem introduced by this patch. The project currently configures Gradle with a 4 GB maximum heap.

**Suggested reviewer action:** rerun the failed job. If the packaging failure repeats, consider increasing the heap available to the PR build workflow or investigating the memory required to package the universal APK. No source change to this PR appears necessary based on the failure.

liongalahad:
### Source-selection behavior (synthetic 30-source fixture)

All source entries shown below are synthetic and use emulator-local dummy URLs. No real add-on or source data is displayed.

**1. Before playback — source 24 is selected**

Source 24 selected before playback

**2. Before the fix — returning from the internal player resets focus to source 1 and scrolls back to the top**

Before fix: selection resets to source 1

**3. After the fix — returning from the internal player restores source 24 in view and focused**

After fix: source 24 is restored

PR #3267, opened 2026-08-29
fix: use Android User-Agent so IPTV DASH streams do not 403

halibiram:
## Summary

Change the internal player's default User-Agent to an Android TV browser string, and replace any incoming desktop Windows User-Agent on playback requests. Some live adaptive streams return HTTP 403 when the player identifies itself as a desktop browser, so playback never starts.

## PR type

- [x] Critical bug fix

## Why

Live streams that otherwise play in other Android players fail in Nuvio with a blocked/403 error. The player was sending a desktop Windows Chrome User-Agent. Some CDNs reject that after the session redirect. Playback itself already supports the stream type; the request is refused before media starts. This is a critical playback failure: the stream never opens.

## Issue or approval

Critical playback bug: live adaptive streams fail with HTTP 403 because the internal player sends a desktop Windows User-Agent. Other Android players play the same streams. No separate GitHub issue is opened; the problem is described in Linked issues.

## Reproduction steps

1. Open a live adaptive stream (HLS/DASH) whose CDN filters client User-Agent.
2. Play it in Nuvio with the internal player and no custom User-Agent override.
3. Playback fails immediately with a blocked/403 error. The same stream plays in other Android players.
4. After this change, Nuvio sends an Android TV User-Agent (and rewrites desktop Windows User-Agent headers), and the stream starts.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Only playback User-Agent handling in `PlayerMediaSourceFactory`: default Android TV User-Agent, and rewriting desktop Windows User-Agent headers on playback requests. No player-engine changes, no DASH/HLS parser changes, no UI, and no addon/scraper User-Agent changes.

## Testing

- Requested the same live adaptive stream with the old desktop Windows Chrome User-Agent: HTTP 403 after the session redirect.
- Requested it with the new Android TV User-Agent: HTTP 200 and a valid manifest.
- Confirmed other Android players already play the stream (they do not send a desktop Windows User-Agent).
- Confirmed addon-supplied desktop Windows User-Agent headers are rewritten so they cannot override the fix.

## Screenshots / Video

Not a UI change.

## Breaking changes

None.

## Linked issues

Live adaptive streams can fail in the internal player with HTTP 403 while the same streams play in other Android players. Nuvio identified itself as desktop Windows Chrome. Some CDNs allow Android/player User-Agents and reject desktop Windows after the session redirect, so Nuvio never receives the manifest. This PR switches the playback User-Agent to Android TV Chrome and rewrites incoming desktop Windows User-Agent headers on playback requests.

### DIFF
--- a/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactory.kt
+++ b/app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactory.kt
@@ -260,7 +260,7 @@ internal class PlayerMediaSourceFactory(private val context: Context) {
         private const val ENABLE_VOD_CACHE = true
         private const val VOD_CACHE_FREE_SPACE_RESERVE_BYTES = 1024L * 1024L * 1024L
         internal const val DEFAULT_USER_AGENT =
-            "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 " +
+            "Mozilla/5.0 (Linux; Android 13; Android TV) AppleWebKit/537.36 " +
                 "(KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
 
         private const val MIME_PROBE_CACHE_SIZE = 64


app/src/main/java/com/nuvio/tv/ui/screens/player/PlayerMediaSourceFactory.kt
```diff
@@ -325,6 +325,12 @@ internal class PlayerMediaSourceFactory(private val context: Context) {
                 val value = (rawValue as? String)?.trim().orEmpty()
                 if (key.isEmpty() || value.isEmpty()) return@forEach
                 if (key.equals("Range", ignoreCase = true)) return@forEach
+                if (key.equals("User-Agent", ignoreCase = true) &&
```
skoruppa: Does this override user agent returned by addons? If so, I don't think we shuld

halibiram: I added this to make sure it always applied, but you're right . we shouldn't override addon user aget. dropping it

PR #3427, opened 2026-09-07
feat: add native nntp engine

### LINKED ISSUE #2197: [Feature]: Add NNTP Support
### Area (tag)

Streams / Sources

### Problem statement

Usenet/NNTP is currently difficult to use for direct streaming without a self-hosted setup, which makes it inaccessible to most non-technical users.

This is a missed opportunity because Usenet can provide content that is often hard to find through debrid services, especially foreign, old, niche, and regional releases. With the recent Real-Debrid issues, users also have fewer reliable alternatives, with TorBox being one of the only strong options while many other debrid services are limited, unreliable, or expensive.

### Proposed solution

Add native NNTP playback support, similar to the latest Stremio beta, so users can stream directly from Usenet providers without needing a separate self-hosted service.

No, this is only a proposal

The main alternative today is using a self-hosted Usenet streaming bridge, such as NZBDav or similar tools. This can work, but it requires users to run and maintain their own server, configure Usenet access, expose services safely, and deal with reliability, updates, and hosting costs.

Another alternative is using Stremio beta, which already supports direct NNTP playback and, from current testing, appears to work well. However, this does not help users who prefer Nuvio or want the same capability inside Nuvio’s own experience.

So the practical alternatives are currently either:

- Use Stremio beta, where NNTP playback already works.
- Self-host a bridge like NZBDav, which is too technical for most mainstream users.

Native NNTP support in Nuvio would remove the need for these workarounds.

_No response_

kernexshadow:
## Summary

- Adds native, on-device playback for nntp streams supplied by addons such as AIOStreams, based on StreamNZB implementation.
- Maps the Stremio nzbUrl, servers, fileIdx, and fileMustInclude fields into Nuvio's stream model.
- Ports the required StreamNZB Go engine packages for NZB parsing, NNTP article retrieval, yEnc decoding, archive handling, media-file selection, and seekable streaming.
- Runs the engine as a token-protected loopback HTTP service and connects the existing player to its HTTP Range endpoint.
- Adds provider connection pooling/failover and a bounded read-ahead segment cache while keeping provider state and credentials session-scoped.
- Integrates engine startup, stream switching, external playback handoff, error reporting, and cleanup with the existing playback lifecycle.
- Has fallback capability of when links are down, tries next. Until 50 fallbacks.
- Bundles one native engine binary per Android ABI together with reproducible build instructions, upstream provenance, GPL-3.0 text, and BSD-2-Clause attribution.

The implementation is ported from Gaisberg/streamnzb at commit 5097e2c1d490dc68d320cd8137334fdd209077ca. Only the dependency closure needed by the Android engine is included, plus Nuvio-specific loopback, configuration, and Stremio file-selection integration.

The ABI-specific binaries are 10.2-11.0 MiB uncompressed. Only the binary matching the device ABI is shipped in each split APK, where it occupies about 4.2 MB compressed. The expected installed-size increase is approximately 15 MB because Android extracts the native executable.

## PR type

- [ ] Critical bug fix

This implements the feature request in #2197.

## Why

Nuvio currently cannot consume the native NZB stream contract used by Stremio addons. Users must instead operate a separate self-hosted bridge. This change lets an addon provide the NZB location and NNTP server definitions once; Nuvio then fetches the NZB, connects directly to the providers, decodes the requested media, and exposes it to the player through a local seekable endpoint.
Stremio already does that altough the implementation doesn't seems so good as this one.

Users also asked this:
#2197
https://discord.com/channels/1379902184207941732/1537151976721227796

## Issue or approval

Fixes #2197

The linked issue is a feature request. No explicit maintainer implementation approval is currently recorded in the issue.

## Reproduction steps

1. Configure an AIOStreams/Stremio addon that returns a stream object containing nzbUrl and servers.
2. Open a title in the current dev build and select that NZB source.
3. Observe that the current app has no native handler for this stream contract and cannot start playback without an external bridge.
4. Install this branch's fullDebug build and select the same source.
5. Confirm that Nuvio starts the local NNTP engine, begins playback, supports seeking, changes sources cleanly, and stops the engine when playback ends.

## UI / behavior impact

- [x] No UI change
- [ ] No behavior change

Playback behavior changes only for stream objects that contain the optional NZB/NNTP fields. Existing HTTP, debrid, torrent, and other addon streams retain their current paths.

## Policy check

- [ ] This PR fits the current PR policy: localization/translation only or a critical bug fix.
- [ ] This PR is small, focused, and limited to one issue.
- [ ] This PR includes a linked issue, reproduction steps, and testing notes if it is a critical bug fix.

## Scope boundaries

- No settings screen or manual NNTP-provider UI was added; server definitions come from the Stremio stream object.
- AIOStreams itself is not modified. It sends the NZB URL and server configuration once, after which Nuvio owns the playback session.
- NNTP credentials, provider health, and usage statistics are not persisted by the Android engine.
- Existing non-NZB playback behavior and public addon contracts are not changed.
- No extra UI polish, visual changes, or unrelated behavior tweaks are included.

## Testing

- ./gradlew :app:compileFullDebugKotlin --console=plain
- ./gradlew :app:testFullDebugUnitTest --tests com.nuvio.tv.core.usenet.NntpServerConfigTest --tests com.nuvio.tv.data.mapper.StreamNzbMapperTest --tests com.nuvio.tv.core.player.StreamAutoPlaySelectorTest --console=plain
- ./gradlew :app:assembleFullDebug --console=plain
- Inspected the generated arm64-v8a split APK and confirmed libnuvionntp.so plus the StreamNZB, rardecode, and provenance license assets are present.
- Manual playback validation on NVIDIA Shield Pro with NZB releases up to 100 GB, including startup and seeking, fast as usenetstreamer.

Go unit tests were not rerun after the final dev synchronization because the Go toolchain is not installed on this machine. The Android build uses the already generated, versioned native binaries.

## Screenshots / Video

Not a UI change.

## Breaking changes

None. The new Stremio fields are optional and existing stream types continue to use their current playback paths.

## Linked issues

Fixes #2197
https://discord.com/channels/1379902184207941732/1537151976721227796


kernexshadow:
### AIOStreams failover compatibility

AIOStreams failover cannot be invoked directly for native Stremio NNTP streams.

For this stream type, AIOStreams sends the client an NZB URL, NNTP provider configuration, and optional file-selection metadata. Playback is then handled entirely by the local Nuvio NNTP engine. Because the media request does not pass through an AIOStreams failover endpoint, AIOStreams cannot observe NNTP errors such as `430 No Such Article` or redirect playback to another release.

This PR therefore handles the immediate failure locally: the NNTP engine performs a lightweight preflight before exposing the loopback playback URL. Releases with definitively unavailable required articles are rejected promptly instead of opening the player and remaining stuck at zero buffered bytes.

Provider fallback and release fallback are separate concerns:

- The NNTP pool can already try another configured provider for articles belonging to the same NZB.
- Trying another NZB/release from the AIOStreams result list would require a client-side fallback mechanism in Nuvio.

A future client-side implementation should use structured NNTP failure reasons, retry only release-specific failures such as missing articles, preserve the result ordering received from AIOStreams, and apply an attempt/time limit. Authentication, configuration, and other provider-wide failures should not trigger retries across multiple releases.

Such a mechanism could only use the results exposed to Nuvio. Internal AIOStreams failover variants, including variants retained through deduplication, are not included in the native NNTP playback contract.

Automatic release fallback is intentionally outside the scope of this PR.

ram130:
> Reviewing.

Got a little busy, but got my notes together. The integration side is clean .. NZB fields optional on the DTO so existing streams are untouched, the fallback policy and server config are unit tested, and reusing the loopback plus range approach the torrent path already uses means the player didn't need special casing. Good work on that. 

**Three things**

1. Build reproducibility. The engine ships as committed binaries in app/src/main/jniLibs, and the only build script is nntp-engine/build-android.ps1, which is PowerShell. That means nobody on Linux or macOS, and no CI, can rebuild libnuvionntp.so from the Go sources in the same repo.

A shell script or a Gradle task that shells out to `go build` would fix it, and would need to confirm the committed .so matches the source. Worth noting #3511  does this and it is
a reasonable bar.

2. _Scope of the port_. nntp-engine/pkg carries 119 files, including media/unpack (39), media/ffprobe (12) and search/parser. 

How much of that is reachable from the Android entry point? Anything that is not is maintenance burden and review surface for the project from here on. If parts are
there for future work, I'd say update the PR as well;  if they are dead, dropping them makes this a lot easier to take on.

3. Failover bound. The description mentions up to 50 attempts. What bounds that in wall clock, and is there backoff between them? On a stick with a flaky provider that is the difference between a slow start and a hang and a possibly seen as regression for the user.

And finally, slight call out, Size: 11 MB per ABI uncompressed, so roughly 44 MB in the repo. Not a blocker given libtorrserver.so is already there at 41 MB, but I'd say to confirm the 4.2 MB per split APK figure holds after the lib*.so extraction rules.

PR #3511, opened 2026-09-13
feat(usenet): add native on-device NNTP streaming engine

### LINKED ISSUE #2197: [Feature]: Add NNTP Support

haaihond:
## Summary

> [!IMPORTANT]
> **Work in Progress**: This implementation is actively being refined and tested across hardware. The detailed diagnostics might be cut down, moved or removed for release.

Adds native, on-device Usenet/NNTP streaming for Stremio sources (`nzbUrl`, `servers`, `fileIdx`, `fileMustInclude`), running an in-process Go daemon on loopback with end-to-end memory-backed streaming and comprehensive tuning options in **Advanced Settings**.

### 1. Native Streaming Engine
- **Direct NNTP Playback**: Direct streaming from NZB sources without requiring external bridges (like NZBDav).
- **In-Memory Payloads**: Slabs and article cache are entirely RAM-backed—no video or RAR payloads are written to disk.
- **Lazy Archive Resolution**: Stored RAR4 and RAR5 extents map arithmetically to NNTP articles without upfront scanning or decompression.
- **Dependency Optimization**: Fetches unchanged upstream C++ RapidYenc SIMD kernels at pinned commit `480bd7b5896f8b3edecc721d23f1384d767ffe2f` (SHA-256 verified) at native build time (-11,221 lines).

### 2. User Toggles & Performance Settings (Advanced Settings)
- **Performance Profiles**: Preset tiers for different hardware constraints:
  - *Low Memory* (32 MiB article budget / 8-segment window)
  - *Balanced* (64 MiB article budget / 16-segment window, default)
  - *High Throughput* (128 MiB article budget / 32-segment window)
- **Read-Ahead Segments**: Configurable segment lookahead window with per-profile byte limits (or Automatic).
- **Max Connections**: Automatic provider allowance balancing or manual global override across servers.
- **Pre-warm Usenet Engine**: Pre-starts the loopback daemon on app launch so it's warm while browsing, without opening provider sockets.
- **Fast MKV Startup**: Pre-fetches and preserves MKV head/tail startup articles across ExoPlayer's Cues index seek to avoid cold tail round trips.
- **Fast NZB Fetch**: Requests compressed NZB metadata to reduce initial stream open latency.
- **Prefetch First Usenet Result**: Pre-resolves and warms startup segments for the top Usenet result on the stream selection screen for instant playback on click.
- **Cache NZB Files**: Document-level disk cache for NZB metadata (capped at 256 MiB LRU, 14-day lifetime) across app restarts.

### 3. Real-Time Startup Diagnostics (Advanced Settings)
- Dedicated **Usenet Diagnostics Card** showing granular metrics for the last stream:
  - *Timing breakdown*: Resolve to first frame, engine boot/reuse, session open, player handoff, decoder setup.
  - *Engine stats*: Head/tail prefetch readiness, NZB fetch & parse duration, RAR continuation discovery time.
  - *Cache efficiency*: Article cache hit rate, NZB disk cache hits/misses, downloads shared vs cancelled, and buffer memory allocated.

### 4. Lifecycle & Security
- Ephemeral session capabilities and random token-authenticated control socket on `127.0.0.1`.
- Zero credential persistence; provider credentials remain in-memory only.
- Strict process lifecycle bound to the Android app via stdin pipe EOF to guarantee clean daemon termination on app exit.

## PR type

- [ ] Critical bug fix

## Why
Currently, playing Usenet streams in Nuvio requires running and maintaining an external self-hosted bridge (such as streamNZB). This PR implements direct, native on-device NNTP playback so users can stream NZB sources directly from Stremio addons without external server infrastructure.

## Issue or approval

Fixes #2197  
*(Feature request #2197; related to review in PR #3427).*

## Reproduction steps
1. Install an addon that returns Stremio stream objects containing `nzbUrl` and `servers` configuration (e.g. AIOStreams).
2. Open a stream results page in Nuvio TV and select the Usenet source.
3. Observe that Nuvio starts the local native daemon, establishes persistent NNTP connections, decodes yEnc segments, and plays back the media stream smoothly with seek and subtitle support.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

- [ ] This PR fits the current PR policy: localization/translation only or a critical bug fix.
- [ ] This PR is small, focused, and limited to one issue.
- [ ] This PR includes a linked issue, reproduction steps, and testing notes if it is a critical bug fix.

## Scope boundaries

- Only streams containing `nzbUrl` and `servers` are handled by this engine; all other playback types are unchanged.
- No drive-by cleanups, unrelated formatting, or architectural modifications outside the Usenet integration.

## Testing
- **Real devices & community**: Tested across multiple physical Android TV devices with Discord users; verified smooth startup (+-7 seconds), seeking, stream switching, subtitles, and clean session cleanup.
- **Build verification**: 
  - Compiled all 4 ABIs from source (`-PbuildUsenetFromSource=true`).
  - Assembled debug APK (`.\gradlew.bat :app:assembleFullDebug`).
- **Automated tests**: Passed `UsenetSidecarTest`, `UsenetStartupBenchmarkTest`, `UsenetStreamProtocolTest`, and Go pool regression tests.

## Breaking changes
None

## Linked issues
Fixes #2197
[Discord feature request](https://discord.com/channels/1379902184207941732/1537151976721227796)


ram130:
Read through the branch. The engineering discipline here stands out: the self-reviews in native/usenet (LATENCY, STARTUP, TUNING, VALIDATION), the instrumented startup benchmark, and the diagnostics card mean this is reviewable in a way native code usually isn't. The Gradle path that builds the engine from source with CMake and cgo rather than trusting a committed binary is great.

Four things I'd raise.

**Prebuilt drift.** native/usenet/prebuilt carries the same four .so files (~40 MB), and the Gradle task copies them unless buildFromSource is set. What stops the prebuilt from drifting from the Go sources? If CI builds with buildFromSource and compares, saying so in the PR would settle it.

**The rapidyenc fetch.** Pinning the upstream commit with a SHA-256 check is better than vendoring, but it makes the build need network at that step. What happens on a build machine that cannot reach it? A vendored fallback or a clear failure message beats a confusing CMake error.

**Prewarm and idle memory.** Bootstrapping the runtime when the app becomes visible buys cold-start time, and LATENCY-REVIEW is honest that the tradeoff is idle runtime memory. On a Fire TV Stick 4K Max the heap is 384 MB and the system starts killing background processes well before that, so a resident Go runtime while the user is browsing is good to measure, not just reasoning about. If you have a PSS figure with prewarm on and off on a low-RAM device, that is the number would be a good indicator for us.

**App surface.** This touches main app, the player initialization, track selection, the MPV surface and three settings screens. Each is justified, but it makes the PR harder to take in one bite. Anything separable into its own PR you think would be good?

And since it is still a draft, what is left before you would call it ready? from your view, I believe this will be the main one we use, then cherry pick from the other PR where needed.

ram130 (commented):
.rar files, are they really needed?

haaihond:
> Read through the branch. The engineering discipline here stands out: the self-reviews in native/usenet (LATENCY, STARTUP, TUNING, VALIDATION), the instrumented startup benchmark, and the diagnostics card mean this is reviewable in a way native code usually isn't. The Gradle path that builds the engine from source with CMake and cgo rather than trusting a committed binary is great.
> 
> Four things I'd raise.
> 
> **Prebuilt drift.** native/usenet/prebuilt carries the same four .so files (~40 MB), and the Gradle task copies them unless buildFromSource is set. What stops the prebuilt from drifting from the Go sources? If CI builds with buildFromSource and compares, saying so in the PR would settle it.
> 
> **The rapidyenc fetch.** Pinning the upstream commit with a SHA-256 check is better than vendoring, but it makes the build need network at that step. What happens on a build machine that cannot reach it? A vendored fallback or a clear failure message beats a confusing CMake error.
> 
> **Prewarm and idle memory.** Bootstrapping the runtime when the app becomes visible buys cold-start time, and LATENCY-REVIEW is honest that the tradeoff is idle runtime memory. On a Fire TV Stick 4K Max the heap is 384 MB and the system starts killing background processes well before that, so a resident Go runtime while the user is browsing is good to measure, not just reasoning about. If you have a PSS figure with prewarm on and off on a low-RAM device, that is the number would be a good indicator for us.
> 
> **App surface.** This touches main app, the player initialization, track selection, the MPV surface and three settings screens. Each is justified, but it makes the PR harder to take in one bite. Anything separable into its own PR you think would be good?
> 
> And since it is still a draft, what is left before you would call it ready? from your view, I believe this will be the main one we use, then cherry pick from the other PR where needed.

Thanks for the thoughtful review.
Prebuilts: I rebuilt all four locally. Before marking this ready, I’ll add a CI check that compares them with source builds.
RapidYenc: I can add the pinned source back to the repo so builds work offline, though that adds about 11k vendored lines. Would you prefer that to the verified fetch?
Prewarm: On a TV emulator, idle combined PSS was 130.2 MiB with prewarm on versus 117.5 MiB off. The sidecar runs separately, so it does not use the app’s Java heap allowance, but it still uses device RAM. I still need a low-RAM device measurement. I plan to ship prewarm off by default; that release tuning is one reason this remains draft.
Scope: I’d like to keep the integration in one PR and provide a clearer review map.
Before marking it ready, I’ll address the binary check, the RapidYenc approach, low-RAM validation, the remaining checks and feature-policy concerns, do some additional testing. I'm currently thinking which place is best for Usenet settings and diagnostics, what do you think @ram130?

haaihond:
> .rar files, are they really needed?

Yes, this type is very common in the Usenet landscape, and removing it will result in a lot of failures. It doesn't take a bit hit even on weaker Android TV devices.

PR #3561, opened 2026-09-16
RTL: Fix Sidebar to Library Card Navigation

### LINKED ISSUE #2950: [Bug]: Focus Moves to the Leftmost Card Instead of the Rightmost Card
### Pre-flight checks

### App version / commit

0.8.2

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

TV

### Android version

Android 14

### Player mode

Internal player

### Area (tag)

UI / Layout

### Bug description

In an RTL layout, when navigating from the Sidebar to the Cards in a Library with multiple items, the focus moves to the leftmost card instead of the rightmost card.

### Steps to reproduce

1. Open a Library with multiple items.
2. Set the layout to RTL.
3. Move the focus from the Sidebar to the Cards by navigating left (or press center)

### Expected behavior

In an RTL layout, the focus should move to the rightmost card, as it is the first card in the RTL direction.

### Actual behavior

The focus moves to the leftmost card, the same way it does in an LTR layout.

### Frequency

Always

### Did this work before?

No, it never worked

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

The behavior I described occurs when you scroll down to rows where you no longer see the filters above.

haveAnIssue:
## Summary

Fixed remote navigation from the sidebar to library cards in RTL layouts.

## PR type

- [x] Critical bug fix

## Why

When navigating from the sidebar to the cards in a Library with multiple items, the focus moved to the leftmost card instead of the rightmost card.

The focus now moves to the rightmost card, which is the first card in the RTL direction.

## Issue or approval

Fixes #2950

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check


skoruppa:
With your changes, focusRestorer always targets the first card, which means the user loses their scroll/focus position when navigating back from the sidebar. It should respect lastFocusedPosterKey, something like:
```
.focusRestorer {
    val lastKey = lastFocusedPosterKey
    if (lastKey != null) {
        posterFocusRequesters[lastKey] ?: firstCardFocusRequester
    } else {
        firstCardFocusRequester
    }
}
```
Also, firstCardFocusRequester replaces the map entry for the first card (if (isFirstCard) firstCardFocusRequester else posterFocusRequesters[focusKey]), so when lastFocusedPosterKey points to the first card, posterFocusRequesters[lastKey] returns a requester that isn't attached to any composable. Consider using a single requester per card to avoid this mismatch

haveAnIssue:
I’ve manually tested the change, and the focus behavior is working as expected. It works correctly in both RTL and LTR layouts, including with a large number of cards and while scrolling.

The only minor issue I noticed is that when moving from a card back to the sidebar, the focus briefly moves to Home before immediately returning to Discover. There is no significant visual impact from this, and overall this is still a substantial improvement over the current behavior, especially in RTL layouts.

I’ll look into how we can address the issue mentioned above, but at this point I believe the PR is ready to be merged. The resulting focus behavior is significantly better than what we currently have in RTL layouts.

I can also attach a video demonstrating the behavior if that would be helpful.

haveAnIssue:
There is currently a focus transition issue that needs to be resolved before this PR can be considered ready.

When the user moves focus to the sidebar from anywhere in the app, including Settings, Library, Search, or Discover, focus briefly moves to Home before landing on the correct Sidebar item.

The same behavior also occurs on the branch associated with PR https://github.com/NuvioMedia/NuvioTV/pull/3600. Since I’m not sure what the best approach is to resolve it, I’m leaving the PR as a draft for now.

PR #3563, opened 2026-09-16
Fix Text Alignment Based on Language Direction

haveAnIssue:
## Summary

Fixed text alignment in steam selection.

## PR type

- [x] Critical bug fix

## Why

Text is now aligned according to its language direction:
* LTR languages are aligned to the left.
* RTL languages are aligned to the right.
This ensures text is positioned correctly within its container regardless of the app's layout direction.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check


skoruppa:
`toAbsoluteAlignment()` is duplicated as a private function in both StreamComponents.kt and StreamScreen.kt. Extract it to TextDirectionUtils.kt next to contentTextDirection() where it logically belongs.

The description Text (line ~199 in StreamComponents.kt) sits outside the CompositionLocalProvider(LocalLayoutDirection provides Ltr) block, but the streamName Row is inside it. This is inconsistent. If an RTL app layout causes the Row to misbehave (which is why you wrapped it), the same problem applies to the description. Either both should be wrapped or neither, otherwise you're fixing it for the title but leaving it broken for the description

haveAnIssue:
Regarding the first point, I extracted `toAbsoluteAlignment()` into `TextDirectionUtils.kt` next to `contentTextDirection()`.

For the second point, I intentionally applied CompositionLocalProvider(LocalLayoutDirection provides Ltr) only to the streamName Row. This is because the “Playing” label is only present in the streamName Row inside StreamComponents.kt and is the part that requires the LTR layout handling. The description does not contain this label, so I left it outside the provider intentionally.

I also verified the result visually, and this approach provides a consistent and well-aligned layout.

Issue #3592, opened 2026-09-17
[Bug]: Returning from the player sometimes restores focus to the Cast tab instead of the episode thumbnail


imler:
### Pre-flight checks

- [x] I searched existing issues and this is not a duplicate.
- [x] I can reproduce this on the latest release or latest main build.
- [x] This issue is limited to a single bug (not multiple unrelated problems).
- [x] This is not a source/stream-specific error (the issue happens regardless of which source is used).

### App version / commit

0.9.4

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

MIBOX3

### Android version

9

### Player mode

Internal player

### Area (tag)

Focus / Remote navigation

### Bug description

In most cases, returning from playback to the movie or TV show details page correctly restores focus to the episode card. However, sometimes focus is restored to a different place.

In the video, I am watching episode 8 of season 2, which is the last episode of that season. When near the end of the episode and pressing Back to return to the details page, focus is restored to the Cast tab below all thumbnails instead of the episode thumbnail.

https://github.com/user-attachments/assets/9bcfe404-29ee-4218-b928-2235634dc112

### Steps to reproduce

1. Open the details page of a TV show.
2. Select the last episode of a season.
3. Start playback.
4. Seek to near the end of the episode.
5. Press the Back button to return to the details page.
6. Observe where focus is restored.

### Expected behavior

Focus should be restored to the episode thumbnail that was selected before playback started.

### Actual behavior

Focus is sometimes restored to the Cast tab below the episode thumbnails instead of the previously selected episode.

### Frequency

Always

### Did this work before?

Not sure

### Logs (required for crash reports)

```shell

```

PR #3611, opened 2026-09-19
Fix details page jump after Studio/Network/Cast/Similar

### LINKED ISSUE #3588: [Bug]: Returning from the Studio/Network view causes the details page layout to jump
### Pre-flight checks

### App version / commit

0.9.4

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

MIBOX3

### Android version

9

### Player mode

Internal player

### Area (tag)

Focus / Remote navigation

### Bug description

When opening the Studio or Network view from a movie or TV show details page and then pressing Back to return to the details page, the entire layout briefly shifts slightly upward and then downward.

This happens every time the Studio or Network view is opened and then closed.

https://github.com/user-attachments/assets/73df39ca-f111-4543-b8b2-dbdf816a8fae

### Steps to reproduce

1. Open the details page of any movie or TV show.
2. Open the Studio or Network view.
3. Press the Back button to return to the details page.
4. Observe the layout.

### Expected behavior

The details page layout should remain in the same position as before when returning from the Studio or Network view.

### Actual behavior

The entire layout shifts slightly upward and then downward when returning to the details page, and remains in the new position.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

### LINKED ISSUE #3589: [Bug]: Returning from a title in the Similar tab restores focus to the wrong place
### Pre-flight checks

### App version / commit

0.9.4

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

MIBOX3

### Android version

9

### Player mode

Internal player

### Area (tag)

Focus / Remote navigation

### Bug description

When opening any title from the Similar tab and then pressing Back to return to the previous details page, focus is not restored to the place where it was before.

Instead, focus is moved to the Crew & Cast tab.

https://github.com/user-attachments/assets/fd7f1924-499c-4729-a4f9-374f8dde39d9

### Steps to reproduce

1. Open the details page of any movie or TV show.
2. Navigate to the Similar tab.
3. Select any title.
4. Open the selected title.
5. Press the Back button to return to the previous details page.
6. Observe where focus is restored.

### Expected behavior

Focus should return to the same place in the Similar tab where the title was originally selected.

### Actual behavior

Focus is moved to the Crew & Cast tab instead of being restored to the previously selected item in the Similar tab.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

### LINKED ISSUE #3590: [Bug]: Returning from an actor page causes unnecessary layout shifts
### Pre-flight checks

### App version / commit

0.9.4

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

MIBOX3

### Android version

9

### Player mode

Internal player

### Area (tag)

Focus / Remote navigation

### Bug description

When opening an actor's details from a movie or TV show details page and then pressing Back to return, the layout is unnecessarily shifted instead of returning to exactly the same position as before.

This happens twice, as shown in the video:

1. Immediately after returning, the layout shifts unnecessarily.
2. Pressing LEFT or RIGHT to move to another actor causes another layout shift.

https://github.com/user-attachments/assets/cc183673-c0af-41f4-b981-46a09a7a5a4a

### Steps to reproduce

1. Open the details page of any movie or TV show.
2. Navigate to the cast section.
3. Open an actor's details.
4. Press the Back button to return to the movie or TV show details page.
5. Observe the layout.
6. Press LEFT or RIGHT to move to another actor.
7. Observe the layout again.

### Expected behavior

The layout should return to exactly the same position as before opening the actor's details and should not shift when navigating between actors.

### Actual behavior

The layout shifts immediately after returning from the actor's details. Pressing LEFT or RIGHT to navigate to another actor causes another unnecessary layout shift.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

### LINKED ISSUE #3591: [Bug]: Returning from an actor page sometimes restores focus to the wrong place
### Pre-flight checks

### App version / commit

0.9.4

### Install method

GitHub Release APK

### Platform

Android TV

### Device model

MIBOX3

### Android version

9

### Player mode

Internal player

### Area (tag)

Focus / Remote navigation

### Bug description

When opening an actor's details from a movie or TV show details page and then pressing Back to return, focus is sometimes not restored to the actor's image as expected. Instead, focus is placed directly on the Cast tab.

It may take several attempts of opening an actor's details and returning before this happens, but the issue can eventually be reproduced. Sometimes it happens even on the first attempt.

Once focus has incorrectly returned to the tab, pressing DOWN focuses a completely different actor instead of the actor that was previously selected.

https://github.com/user-attachments/assets/610b7d96-d20b-453c-947b-8cf947fd3706

### Steps to reproduce

1. Open the details page of any movie or TV show.
2. Navigate to the cast section.
3. Select an actor and open their details.
4. Press the Back button to return to the movie or TV show details page.
5. Repeat the process if necessary until focus is restored incorrectly.
6. Observe where focus is placed.

### Expected behavior

* Focus should be always restored to the actor's image that was selected before opening the actor's details.

### Actual behavior

* Focus sometimes returns directly to the tab instead of the previously selected actor.
* When this happens, pressing DOWN focuses a different actor rather than the one that was previously selected.

### Frequency

Always

### Did this work before?

Not sure

### Media details (optional)

_No response_

```shell

```

### Anything else? (optional)

_No response_

halibiram:
## Summary

Returning from Studio, Network, Creator, Cast, Similar, or Collection no longer recenters the details page or drops focus on the Cast tab. The same logo, person, or title stays focused, the previous vertical offset is kept, the last Cast / Similar / Collection row position survives a hop, Cast-to-Channel eases instead of slamming to the end, and the first fully visible Cast cards no longer nudge a few pixels sideways.

## PR type

- [x] Critical bug fix

## Why

The details `LazyColumn` already remembers its place. The jump was a second scroll after that restore, not a lost `LazyListState`.

Opening Studio, Network, Creator, Cast, Similar, or Collection leaves the details screen (`TmdbEntityBrowse` / person / another title). On Back, `ON_RESUME` bumps `restoreFocusToken` / `companyRestoreToken` and `requestFocus()` puts focus back on the same card. That is correct. What is not correct is the TV `BringIntoViewSpec` that runs on every focus change: `NuvioScrollDefaults.smoothScrollSpec` centers the focused item at **42%** of the viewport (`NuvioFocus.tokens.scrollViewportTarget`). A card that is already fully on screen, and a column that already sits at the saved offset, still get asked to move so the item’s center lands on that 42% line. The user sees a layout jump and then stays at the new offset.

Temporary device traces (not in this PR) made that extra request visible. On return the column was already at `firstVisibleItem` **`2/314`**. Then:

- A **`112px`** company logo sat near the bottom of the **`1080px`** viewport (`offset` ~`956`). 42% of 1080 is **`~454px`**, so `calculateScrollDistance` asked for about **`+558px`** of vertical recenter. After the jump the same page sat at e.g. **`2/93`**.
- In the same moment a smaller **`57px`** child asked for **`-221px`**, so the page twitched up before the large down-scroll.
- The horizontal logo row, already at the start, still calculated about **`-570px`** toward the same 42% target (clamped, but it still fought the vertical move).

Suppressing `bringIntoView` only while that restore `requestFocus` runs stops the jump. After restore finishes, normal D-pad still uses the 42% spec. Play was not stealing focus; the traces ruled that out.

Creator and Cast used the same restore + 42% recenter (#3590). Returning from an actor therefore jumped immediately, and LEFT/RIGHT to the next person asked for another 42% snap. The same restore-only suppress is applied there. `clearPendingRestore` / `mark*` now also zero `restoreFocusToken`, and Cast `onRestoreFocusHandled` runs after `requestFocus` + 2 frames, not on the first focus flicker, so a leftover token cannot land on the Cast tab or a different actor (#3591).

That still left follow-on bugs on the same screen, all from the same 42% / restore interaction.

**Lost page offset and row reset.** After the user moved the details list a little, or hopped Studio → Network → Cast / Similar / Collection, the previous vertical place could not be found. `rememberLazyListState` on the column is not enough: composition leaves, focus restore recenters, and the nested `LazyRow` lived *inside* a `LazyColumn` item. Leaving details disposed that row, so its horizontal index reset to `0`. `focusRestorer` then always fell back to the first card. Network → Cast could also recenter a still-visible **`200px`** card toward 42% (about **`14–65px`** of page motion, e.g. Joely **`-64.6px`**). Cast, Similar, and Collection `LazyListState` and last-focused keys are now hoisted next to the page list. The last column `index/offset` is `rememberSaveable` when leaving, and applied on `ON_RESUME` *before* bumping the restore tokens so `requestFocus` does not fight a later snap.

**Cast → Channel slammed to the end.** Channel / Network logos sit at the bottom of a long details page. A **`112px`** logo still asked to sit at 42% of **`1080px`**. That distance is larger than the remaining scroll, so the column hit the last pixel in one spring. Coming back up then needed the Cast offset, not a second 42% snap. The logo row now keeps horizontal bring-into-view but reports **zero height** to the parent, so it does not yank the column. Any clipped overflow is revealed with `animateScrollBy(..., slowTween())` (EaseOut). The Cast / Similar / Collection offset is pinned on the way down and eased back the same way if the page only nudged (same index, **`1–200px`**).

**First three Cast cards nudge left/right (#3590 LEFT/RIGHT).** The Cast row inherits the same vertical 42% spec for horizontal focus. Home catalog rows override that to start-align; Cast does not, and must keep the existing snap for people who are off-screen. At the start of the row the first cards are already fully visible, but their centers are not exactly on 42% of **`1920px`** (`target` **`806`**):

- Leading / first cards request a large negative distance (Guy **`-610`**, Theo **`-308`**) and the row is already at `0/0`, so Compose clamps and nothing should move.
- The third visible card (Kaya, `offset=714`, `size=200`) is only **`+7.6px`** past 42%. That is a *positive* distance, so the row actually scrolls `0/0` → `0/8` and clips Creator at **`-8`**.

A Cast-only spec or a stay-if-visible rule would change when snap applies (rejected: it killed or delayed real snaps). The only change in `smoothScrollSpec` is: if the item is **already fully visible** and `0 < distance < 12`, return `0`. That covers the ~`8px` Kaya case. Off-screen cards and larger 42% snaps are unchanged.

**Similar / Collection returned to the Cast tab (#3589).** People tabs switch `activePeopleTab` on any `onFocusChanged`. On Back the system often hits the first tab (Cast) before the Similar card is placed, `Crossfade` unloads Similar, and restore `requestFocus` has nothing to attach to. If `moreLikeThis` is empty for a frame, Similar is dropped from the tab list and the saved tab is reset to Cast. Similar restore also inherited the 42% jump (landscape `260×146`). Restore now locks the tab until the pending type finishes, does not reset Similar/Collection to Cast while the list is empty, suppresses BIV for those restores, and the target card grabs focus when it is placed (same as post-play). The last Similar / Collection row and page offset are hoisted like Cast.

## Issue or approval

Fixes #3588
Fixes #3589
Fixes #3590
Fixes #3591

## Reproduction steps

1. Open a series details page that has Network, Studio, Creator, Cast, and Similar.
2. Scroll down until a company logo or a creator/cast card is focused.
3. Open that studio, network, or person, then press Back.
4. Before this fix, the page jumped and recentered. After this fix, the same item stays focused and the list stays at the same scroll offset.
5. Move a little up or down on details, hop Studio/Network/Cast, and come back. The page should return to the previous offset, and Cast should keep the last focused person instead of jumping to the start.
6. Focus a Cast card, then move down to Channel/Network. The page should ease to reveal the logo instead of slamming to the end; moving back up should ease to the pinned Cast offset.
7. On Cast, D-pad across the first three people. The row should stay put instead of sliding a few pixels left or right.
8. From Cast, open an actor, press Back, then LEFT/RIGHT to the next person. Focus stays on that actor (not the Cast tab), and the page does not shift again.
9. Open a title from Similar (or Collection), press Back. Focus returns to that title, not the Cast tab, and the page does not jump.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check

## Scope boundaries

Only details-page restore and the related Cast / Similar / Collection / company scroll on that screen. These four issues are the same restore + 42% jump. Hero and episode restore paths are untouched. Season auto-switch and the 250ms episode-restore timeout are unchanged; #3592 is not claimed here. Normal D-pad scrolling still uses the existing 42% bring-into-view after restore finishes. The `smoothScrollSpec` skip applies only when an item is already fully visible and the requested recenter is under 12px; home catalog rows and larger snaps are unchanged. No layout, copy, or navigation destination changes.

## Testing

Manual on an Android TV device with a series that has Network, Studio, Creator, Cast, and Similar:

- Opened a studio from the details page and pressed Back four times . Before the fix, return started at `2/314` and then moved (for example `2/314` → `2/93`). After the fix the same runs stayed at `2/314`, the `+558` / `-221` recenters were suppressed, and focus returned to the same logo.
- After return, D-pad left/right between logos and up toward Play still scrolled the page as before.
- Confirmed Play did not steal focus while the company row was still on screen.
- Repeat the same return path from a Creator card and a Cast card. The page stays put and focus returns to that person, not the Cast tab. LEFT/RIGHT to the next actor does not shift the page again.
- After a small details scroll, hop Studio/Network/Cast/Similar/Collection and return. The previous page offset is restored, and the row keeps the last focused card instead of resetting to the first item.
- Cast to Channel eases to the logo overflow instead of hitting the end; Channel back to Cast / Similar eases to the pinned offset.
- D-pad across the first three Cast people no longer nudges the row (`0/0` stays `0/0`; the ~`8px` Kaya recenter is skipped). Later cards still snap to 42%.
- Similar (and Collection) title → Back: focus returns to that card, not the Cast tab, and the page does not jump.

## Screenshots / Video

Not a UI change

## Breaking changes

None

## Linked issues

Fixes #3588
Fixes #3589
Fixes #3590
Fixes #3591


imler:
@halibiram I was checking how your fixes work in practice, and in general, they do work, but I found two more things. Should I move these to separate issues, or is it enough that I mention them here as a follow-up for you?

1. When returning from Cast or Similar tab, the Cast tab is still briefly highlighted first (you can see that it enlarges or changes color), and only then does the focus jump to the correct element. You can see that the Cast label gets a brief animation, and then the correct element is selected and gets its border, also with an animation. This is especially noticeable, for example, in the Similar tab where the Cast label is also briefly highlighted at the beginning.

Ideally, the Cast tab shouldn't be highlighted at all, and the focus should immediately be in the desired place without any animations - exactly like it works in the Trailer tab. When returning there, the card is immediately focused and already has its border in place, without any animation or the initial jump from the Cast label to the correct element.

It would be great if the behavior were exactly the same everywhere, to keep things consistent. What we have now works well from a practical point of view and is definitely better than before, but it just feels like the final bit of polish is missing to make the behavior fully consistent :)

https://github.com/user-attachments/assets/6ddb585d-b93e-4dd5-95fa-7a05706aae57

https://github.com/user-attachments/assets/4fbaba10-ee03-43d1-8346-594cc8949ea4

2. This one is a bit more serious. When we switch to any tab other than Cast, then open the Network or Studio view and return to the details view, you can see that unfortunately the selected tab is lost. Every time, it jumps back to the Cast tab and the layout shifts as well.

https://github.com/user-attachments/assets/3356e944-9da0-481a-8e95-db3a6a991234

halibiram:
I'm aware of the issue in the first two videos and I'm still struggling with it. I haven't found a solution yet, but I fixed the issue in the third video

PR #3617, opened 2026-09-20
Fix Initial Stream Focus

haveAnIssue:
## Summary

Fixed stream focus behavior.

## PR type

- [x] Critical bug fix

## Why

After streams are loaded, the focus now consistently remains on the first stream of the first add-on, providing predictable navigation.

## UI / behavior impact

- [ ] No UI change
- [ ] No behavior change

## Policy check


skoruppa:
I'm not sure what this PR is trying to fix

haveAnIssue:
> I'm not sure what this PR is trying to fix

When the streams are loaded, focus should remain on the first stream.
Currently, that’s not the case, and this PR fixes that behavior.

skoruppa:
Never noticed that. Can you show?

haveAnIssue:
The issue occurs in RTL layouts, and this change fixes it. Why would it be considered invalid?

halibiram:
> The issue occurs in RTL layouts, and this change fixes it. Why would it be considered invalid?

If first stream focus is broken in RTL that’s a real bug, but this PR doesn’t touch RTL or the focus code at all. It just starts the screen through OnRefresh instead of the normal load. That’s not a fix for the layout issue, it’s a workaround in the wrong place.
If you have a recording of the RTL behavior before and after, post it. That’s what was asked for earlier. Without that this still isn’t a valid patch for that bug

halibiram:
This also breaks the stream cache. On a normal open we skip the addons if the 15 minute cache is still there. OnRefresh is a force refresh, so it drops that cache and searches again every time. Opening the stream screen would reset it, which isn't worth it for a focus issue

